PR #48, opened 2016-06-29
Break up images utils file in to smaller sub-modules

lucasholucasho:


wtgee:
Great, thanks @lucasholucasho. Let's rename the files so they don't have the `image_` portion. These utils are used from outside the main application, for instance in some of the `scripts/` files, so I would rather the import statements there be able to read `from pocs.utils.images import metadata` etc rather than `from pocs.utils.images import image_metadata`.

wtgee:
We will need to fix some of the scripts and some other areas, but I am going to merge this now so we can work with these new files at the Hack Day.

PR #59, opened 2016-08-18
Refactor

### LINKED ISSUE #35: /var/panoptes path hardcoded
There are several instances where /var/panoptes is hard coded instead of the PANDIR environment variable.

### LINKED ISSUE #36: Determine Environment Variables
Figure out which environment variables we want and create script to ensure variables are set and dirs exist.

See also #35 and #32

### LINKED ISSUE #44: Scheduler Constraints
Developer scheduler `Constraints`. This should be done with consideration to the ongoing development in [astroplan](https://github.com/astropy/astroplan). Minimum first `Constraints` to be developed should maximize observable time so that targets about to go below horizon are not selected.

### LINKED ISSUE #45: Target / Observation / Exposure Refactor
The implementations of the `Target` and `Observation` classes are currently clunky and need some major cleaning up. This should be done with respect to what we want the eventual `Scheduler` to look like (see #44 ). `cameras` should not needlessly be passed into objects.

wtgee:
This covers a lot of material

Closes #45 closes #44 closes #35  closes #36


pocs/observatory.py
```diff
@@ -543,32 +431,43 @@ def _create_mount(self, mount_info=None):
     def _create_cameras(self, **kwargs):
         """Creates a camera object(s)
 
+        Loads the cameras via the configuration.
+
         Creates a camera for each camera item listed in the config. Ensures the
         appropriate camera module is loaded.
 
+        Note: We are currently only operating with one camera and the `take_pic.sh`
```
jamessynge: Is this comment still true (i.e. using only one camera)?

PR #61, opened 2016-08-29
Fits headers

wtgee:
The `Observatory.observe` loops through each camera and creates a subprocess (`scripts/take_pic.sh` is called in the subprocess). Then the camera processes are `wait`ed on in another loop. Finally, in a last loop through the `metadata_info`, the CR2 files are converted to FITS. The `metadata_info` is a collection of the information from the observatory (via `get_standard_headers`), the `current_observation.status()`, and the information from each specific camera. 

The `metadata_info` has descriptive keywords and the `cr2_to_fits` method is responsible for filtering this dictionary and writing out FITS appropriate headers. The `metadata_info` is written as is to mongo after the FITS file has been created.

Also included in this PR are some methods to improve testing, allowing for the option of not having simulators (needs to be tested - simulators are assumed and some tests are skipped).


joshwalawender:
Looks good to me.  My only quibble might be to explicitly specify units on some of the headers.  For example, when filling out the RA, HA, Dec, lat, long, etc. if the values are `Quantities`, then use, for example, `.to(u.degree).value` and then note the units in the FITS header comment.  Otherwise, looks good to merge.

PR #72, opened 2016-12-15
SBIG camera support

AnthonyHorton:
Adds basic support for CCD cameras from SBIG via a Camera class with take_exposure, take_observation & process_exposure() methods.

Also, for SBIG cameras, adds:

- CCD cooling control
- CCD temperature in FITS headers
- Support for dark exposures


wtgee:
@AnthonyHorton, the failing tests are not related to this PR. Sometimes the very last test timeouts out for a weird reason I'm still trying to figure out (it's the one that allows for interrupting for the Grav Wave stuff). 

I'll start a review of this shortly. Thanks!

AnthonyHorton:
Naturally when I re-read my own code the next morning I immediately discover a (minor) error... Fix pushed.

I do have a number of enhancements I want to make, but I figure it's better to get this basic implementation up to standard and merged before making any further significant changes?

I will confess that this code is lacking unit tests (they're a bit tricky when the operation of the code depends on the attached hardware), and I haven't properly tested the `take_observation` and `process_exposure` methods, just copied with minimum changes the methods from the Canon Camera class.

wtgee:
Yeah, I've been trying to figure out a way to test with hardware. We could write a series of tests that are optional and then get skipped on travis-ci, and this is probably a good thing to do for us on the cameras anyway, but it kind of defeats the whole point of the automated testing on each commit if all the tests are skipped.

Currently some of the tests are still failing for unrelated reasons (good old astroplan sun calculations that timeout)

AnthonyHorton:
Yeah, only way to have a reasonably complete set of unit tests for the hardware interface code would be to implement low level hardware simulators so that the tests can be run regardless of what hardware is connected. Actually implementing (and testing!) those simulators would be a bigger project than the hardware interface code, though!

We did that sort of thing when we replaced the AAT's Telescope Control System but would probably be disproportionate here.

Let me know if I need to rebase this branch to get the automated tests working reliably.

wtgee:
Pulling in @joshwalawender for his interest in the SBIGs and the general hardware testing issues.

Yeah, for the most part I don't think we want to write low-level simulators as that will be more work than writing the actual drivers. I think if we keep the method stubs largely the same, especially in terms of what they return, then we can get by with the basic camera simulator testing the `take_observation`, `take_exposure`, and `process_exposure` methods. For PANOPTES we might want to some time in the future have an actual testing unit that responds to these commits and does a real test (we would have to switch to jenkins instead of travis-ci). But for the most part I don't think it's worth our effort.

However, I could make an argument for making some automated tests that interact with the rest of the system well so that once we have all the hardware attached you could run through a whole series of tests in order to ensure your particular unit (which in this case is Huntsman) was working properly. So @AnthonyHorton it might be worth your time to just move the items from the notebook into a test within `$POCS/tests/`

joshwalawender:
Glad to see this happening!  I think  @wtgee 's idea of having tests to deploy once you've connected to hardware is a good one.  That is a reasonable compromise in my opinion.

AnthonyHorton:
Started to add some tests. They're all skipped if the SBIG Universal Driver/Library is not installed on the machine running the tests, and if there are no actual SBIG cameras connected and powered on there will be some xfails.

AnthonyHorton:
Added more tests, now covering a reasonable fraction of the Camera code. Unfortunately most will incorrectly xfail due to a failure to fully unload the SBIG Universal Driver/Library C shared library when the Python wrapper object (`pocs.camera.sbigdrv.SBIGDriver`) is deleted.

wtgee:
@AnthonyHorton, for now I am setting the coverage to skip most of the actual tests that require hardware so that even if the tests fail codecov will be ignoring it. If you want to do the same you can add entries here:
https://github.com/panoptes/POCS/.codecov.yml

AnthonyHorton:
I reorganised the tests to work around the inability to properly unload the `ctypes.CDLL`, all tests now run successfully provided that the driver is installed and at least 2 SBIG USB cameras are connected. If only 1 SBIG camera is available there will be 1 xfail but the rest of the tests will succeed.

The tests now include a test of the `take_observation()` and `process_exposure()` methods, though I wasn't sure of the best way to confirm success with this.

I've also added `sbig.py` and `sbigudrv.py` to the ignore list in `codecov.yml`.

wtgee:
Awesome. Let me know if it is at a point you want me to merge in or if you want to just keep contributing to this branch.

AnthonyHorton:
Very nearly ready to merge, as far as I can tell (test suite appears to get stuck in `pocs` tests at present). I just want to add a safeguard for attempts to start an exposure on a camera that already has one in progress, at present doing that results causes a hang. Once that's in, and if you're happy with the code, then I'd like to get it merged and move on to the Birger adapter/focusing stuff.

wtgee:
Yeah, sorry, the hanging tests should be fixed with the latest commits. (right after you rebased of course!)

AnthonyHorton:
OK, I _think_ this is ready to merge.

AnthonyHorton:
One final change, added some code to disable a camera option (on by default) that may have adversely effected bias consistency.

Oddly the test suite is now completing on Travis but *not* on the Huntsman Mac Pro. It's still getting stuck in `pocs` tests despite me rebasing to the latest `panoptes/develop` commit. Something wrong with my POCS environment?

PR #75, opened 2017-01-17
SBIG & Birger integration

AnthonyHorton:
Changes integrate SBIG `Camera` and Birger Adaptor `Focuser` with each other & with `Observatory`.

- `Camera` can optionally create/accept & associate with a `Focuser`
- `Observatory` can now create SBIG `Camera` and `Focuser` instances from configs
- Additional tests

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -19,6 +19,7 @@ class AbstractCamera(PanBase):
                  model='simulator',
                  port=None,
                  primary=False,
+                 set_point=None,
                  focuser=None,
                  focus_port=None,
                  *args, **kwargs):
@@ -28,7 +29,7 @@ class AbstractCamera(PanBase):
             self._image_dir = self.config['directories']['images']
         except KeyError:
             self.logger.error("No images directory. Set image_dir in config")
- 
+
         self.model = model
         self.port = port
         self.name = name
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -55,7 +55,7 @@ class AbstractCamera(PanBase):
             else:
                 try:
                     module = load_module('pocs.focuser.{}'.format(focuser))
-                except ImportError as err:
+                except AttributeError as err:
                     self.logger.critical("Couldn't import Focuser module {}!".format(module))
                     raise err
                 else:
@@ -102,7 +102,7 @@ class AbstractCamera(PanBase):
         raise NotImplementedError
 
     def __str__(self):
-        return "{}({}) on {}".format(self.name, self.uid, self.port)
+        return "{} ({}) on {}".format(self.name, self.uid, self.port)
 
 
 class AbstractGPhotoCamera(AbstractCamera):  # pragma: no cover
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -1,6 +1,8 @@
 from .. import PanBase
 from ..utils import error
 from ..utils import listify
+from ..focuser.focuser import AbstractFocuser
+from ..utils import load_module
 
 import re
 import shutil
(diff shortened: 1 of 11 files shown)


AnthonyHorton:
Oops, decreasing the coverage again. I think I can fix that, though.

pocs/focuser/focuser.py
```diff
@@ -46,6 +49,21 @@ def position(self, position):
...
+        """
+        Reference to the Camera object that the Focuser is assigned to, if any. A Focuser
+        should only ever be assigned to one or zero Cameras!
+        """
+        return self._camera
+
+    @camera.setter
+    def camera(self, camera):
+        if self._camera:
+            self.logger.error("{} already assigned to camera {}!".format(self, self.camera))
```
wtgee: This would probably be better as a `warning` that also says something like ", skipping assignment".

pocs/camera/camera.py
```diff
@@ -17,6 +19,9 @@ def __init__(self,
                  model='simulator',
                  port=None,
                  primary=False,
+                 set_point=None,
```
wtgee: It looks like nothing happens with this `set_point` here. The abstract class already accepts the `kwargs` so I think it might be slightly confusing to specify it here and then do nothing with it.

pocs/camera/camera.py
```diff
@@ -80,7 +102,7 @@ def process_exposure(self, *args, **kwargs):
         raise NotImplementedError
 
     def __str__(self):
-        return "{}({}) on {}".format(self.name, self.uid, self.port)
+        return "{} ({}) on {}".format(self.name, self.uid, self.port)
```
wtgee: Does it make sense to include the `focuser` here if we have one?

pocs/focuser/focuser.py
wtgee: This would probably be better as a `warning` that also says something like ", skipping assignment".

pocs/camera/camera.py
AnthonyHorton: OK, fair point, I'll take that out.

The SBIG `Camera` class has `CCD_set_point` as a property and implements getter and setter methods. Would it make sense to include stub versions of these in the `AbstractCamera` base class, to encourage a consistent interface for cameras with cooling controls?

pocs/camera/camera.py
AnthonyHorton: It does, though the resulting string representation could get pretty long. The `Focuser` classes have their own `name`, `uid` and `port` attributes which are included in their string representations.  Perhaps it should be just `"{} ({}) on {} with {} focuser".format(self.name, self.uid, self.port, self.focuser.name)` for a `Camera` with a `Focuser`?

pocs/focuser/focuser.py
AnthonyHorton: OK. I made it an error as attempting to assign a `Focuser` instance to more than one `Camera` instance seemed like a pretty clear indication that something has gone wrong. I can't think of any situation where this would be what you intended to happen.

pocs/camera/camera.py
wtgee: Yeah, I think that makes sense. Make sure it has a simple descriptive docstring though as I think people dealing primarily with DSLRs and PANOPTES won't necessarily know what a set point is and might think they need to implement it.

pocs/camera/camera.py
wtgee: Just the name of the focuser is fine, I just think that when printing the representation of a camera it should at least be obvious that it does have a focuser attached.

pocs/focuser/focuser.py
wtgee: Yeah, I don't think it should ever happen either it doesn't halt the code or change operation in any way if it gets there, so I think a warning if fine. To some extent we just use `error` like a `critical`. (I think `critical` is only used in a few places and even there they should probably just be `error`s)

PR #79, opened 2017-02-07
Add autofocus capability for cameras with focusers

AnthonyHorton:
Implements an `autofocus()` method in the `AbstractCamera` base class to provide autofocusing capability for cameras with focusers. Also adds a `autofocus_cameras()` method to `Observatory` for focusing all connected cameras.

Note: I know that more work on the test suite will be needed before this is ready to merge, just wanted to get the process started.


pocs/camera/camera.py
```diff
@@ -1,13 +1,30 @@
...
+
+from ..focuser.focuser import AbstractFocuser
+
+from astropy.io import fits
+from astropy.stats import sigma_clipped_stats
+from astropy.modeling import models, fitting
+
+import numpy as np
+
+import matplotlib
```
wtgee: As discussed, let's move `autofocus` and all plot-related dependencies into the `Focuser` itself. Can make plotting smart or not within that class but might not need it.

pocs/camera/camera.py
```diff
@@ -153,6 +174,249 @@ def take_exposure(self, *args, **kwargs):
     def process_exposure(self, *args, **kwargs):
         raise NotImplementedError
 
+    def autofocus(self,
```
wtgee: Let's leave an `autofocus` thin-wrapper within the `Camera`

pocs/camera/camera.py
```diff
@@ -153,6 +174,249 @@ def take_exposure(self, *args, **kwargs):
...
+                self.logger.info('Fine focus plot for camera {} written to {}'.format(self, plot_path))
+
+        self.logger.debug('Autofocus of {} complete - final focus position: {}'.format(self, final_focus))
+
+        if finished_event:
+            finished_event.set()
+
+        return initial_focus, final_focus
+
+    def _get_thumbnail(self, seconds, file_path, thumbnail_size):
```
wtgee: I think this is fine to leave within the camera itself. Public?

pocs/camera/simulator.py
```diff
@@ -83,26 +89,31 @@ def take_observation(self, observation, headers=None, **kwargs):
...
+                                function=self._fake_exposure,
+                                args=[seconds, start_time, filename, exposure_event, dark])
+        exposure_thread.start()
 
-        # Send command to camera
-        try:
-            proc = subprocess.Popen(run_cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True)
-        except error.InvalidCommand as e:
-            self.logger.warning(e)
+        if blocking:
```
wtgee: Does the blocking need to happen within the `take_exposure`? For other exposures the "blocking" happens within the state machine itself, while `POCS` is in the `observing` state (or `pointing`). It's also not a true block, but instead checks for interrupt messages to the system as a whole. 

See https://github.com/panoptes/POCS/blob/develop/pocs/state/states/default/observing.py#L18

Can we have the user of `take_exposure` control the blocking or is there a reason to have it here?

pocs/focuser/birger.py
```diff
@@ -112,7 +130,6 @@ def move_to(self, position):
         Does not do any checking of the requested position but will warn if the lens reports hitting a stop.
         Returns the actual position moved to in lens encoder units.
         """
-        self.logger.debug('Moving focus to {} encoder units'.format(position))
```
wtgee: We've long talked about making `logger` level that's lower-level than `DEBUG`. This is the kind of thing that would be nice there :(

pocs/camera/camera.py
```diff
@@ -153,6 +174,249 @@ def take_exposure(self, *args, **kwargs):
...
+
+        for i, position in enumerate(focus_positions):
+            # Move focus, updating focus_positions with actual encoder position after move.
+            focus_positions[i] = self.focuser.move_to(position)
+
+            # Take exposure
+            thumbnail = self._get_thumbnail(seconds, file_path, thumbnail_size)
+
+            # Calculate Vollath F4 focus metric
+            f4[i] = images.vollath_F4(thumbnail)
```
wtgee: Let's call `images.get_focus_metric(thumbnail, algorithm='vollath_F4')` here. `vollath_F4` can be the default but gives us the ability to add more in the future without too much trouble (although then algorithm should probably be read from the config). `pocs.utils.images` can then just have a some kind of thin-wrapper lookup.

pocs/camera/simulator.py
AnthonyHorton: The blocking exposure option is only in the Simulated Camera for consistency with the SBIG Camera, and it's only in the SBIG because I found it convenient for interactive operation of a standalone camera when I was developing the SBIG code. It's easy enough to catch the returned Event and wait() on it instead but having the option makes standalone use marginally easier.

pocs/focuser/birger.py
AnthonyHorton: Would switching the logging level to INFO & above for regular use go too far the other way, i.e. not enough stuff being logged?

wtgee: I try to keep `INFO` for basic information you see when running a session. So if you load a `POCS` instance and initialize it, you get:

```
In [2]: pocs = POCS(simulator=['camera', 'weather', 'night'])
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO ******************************** Starting POCS *********************************
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO Initializing PANOPTES unit
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO Loading state table: simple_state_table
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO 	Initializing observatory
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO 		 Setting up location
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO 		 Setting up mount
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO Loading mount commands file: /storage/panoptes/POCS/resources/mounts/bisque.yaml
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO 		 Setting up cameras
> 2017-02-10 21:29:31 ganymede panoptes[11144] INFO 		 Setting up scheduler
> 2017-02-10 21:29:32 ganymede panoptes[11144] INFO 	 Observatory initialized
> 2017-02-10 21:29:34 ganymede panoptes[11144] INFO Welcome GeeBot!

In [3]: pocs.initialize()
> 2017-02-10 21:29:34 ganymede panoptes[11144] INFO Connecting to mount
> 2017-02-10 21:29:34 ganymede panoptes[11144] INFO Mount connected
> 2017-02-10 21:29:34 ganymede panoptes[11144] INFO Initializing pocs.mount.bisque mount
> 2017-02-10 21:29:34 ganymede panoptes[11144] WARNING TheSkyX requires location to be set in the application
> 2017-02-10 21:29:34 ganymede panoptes[11144] INFO Mount initialized: True
```

From experience of watching the logs while running, you don't really want much more than that. You get some basic info but it's pretty high level and is assuming that someone is actually reading (most) of it while in use.

wtgee: So, to actually answer the question, I don't think there would be a lot logged from the focuser at the `INFO` level but that's fine. Basically a message for creation. Maybe before and after focusing if that whole procedure takes time.

pocs/camera/camera.py
AnthonyHorton: Have moved the body of `autofocus()` and `_autofocus()` from `AbstractCamera` to `AbstractFocuser`. It could be argued that I should move the bulk of the autofocus tests from `test_camera.py` to `test_focuser.py` too, but then again you can't autofocus without both a `Camera` and a `Focuser`.

pocs/focuser/birger.py
AnthonyHorton: What I was trying (and failing) to ask was whether there was really a need for a logging level below `DEBUG` to be available, i.e. does reading the logs at `INFO` level & above give you enough information for normal use?

AnthonyHorton: I've certainly kept use of `INFO` to a minimum in the `Camera` and `Focuser` code. Basically a message or two on initialisation, plus one at the end of an autofocus if plotting saving where it was written to.

pocs/camera/camera.py
AnthonyHorton: Yes, no particular reason for it to be private.

pocs/camera/camera.py
AnthonyHorton: Done. Means I don't have to re-write all those tests.

pocs/camera/camera.py
AnthonyHorton: OK, done.

AnthonyHorton:
OK, with that latest commit I think this is ready to merge.

PR #81, opened 2017-02-10
Bisque software

wtgee:
Not quite ready to be merged but almost there. All bisque-related tests are disabled without a running version of TheSkyX.


wtgee:
Ignoring the codecov for now and merging. Need to make up for that later.

PR #85, opened 2017-04-02
Horizon features

brendan-o:
#Adding the Horizon class


requirements.txt
```diff
@@ -14,8 +14,8 @@ PyYAML >= 3.11
 pyzmq >= 15.3.0
 pycodestyle
 wcsaxes
-readline
-gcloud
+#readline
+#gcloud
```
wtgee: Actually you should leave these in here and just fix locally.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +163,123 @@ def get_score(self, time, observer, observation, **kwargs):
...
+    # Image Segmentation with Watershed Algorithm
+    # def process_image():
+
+    # Get the user to input az, el coordinates
+    # After a horizon instant has been instantiated this method can be called
+    # to populate the obstruction_points from user input
+
+    def enter_coords():
+
+        import ast
```
wtgee: All `import` statements should be at the top of the file.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+
+class Horizon(BaseConstraint):
+
+    """ Implements horizon and obstruction limits"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.obstruction_points = []
+
+        print("Horizon.__init__")
```
wtgee: Need to remove all `print` statements and use the `self.logger.debug` (or `.info`, `.warning`, etc) system.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+        to allow for incomplete horizon images
+
+        Note: 
+            Incomplete method, further work to be done to automate the horizon limits
+
+        Args:
+            image_filename (file): The horizon panorama image to be processed
+        """
+
+        from skimage.io import imread
```
wtgee: `import` statements should always be at the top of the file

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+        np.set_printoptions(threshold=np.nan)
+        print(edges1.astype(np.float))
+
+    def get_config_coords(self):
+        """
+        Retrieves the coordinate list from the config file and validates it
+        If valid sets up a value for obstruction_points, otherwise leaves it empty
+        """
+
+        from pocs.tests.test_horizon_limits import obstruction_points_valid
```
wtgee: Move to top.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+        print("Enter a list of azimuth elevation tuples with increasing azimuths.")
+        print("For example (10,10), (20,20), (340,70), (350,80)")
+
+        self.obstruction_points = input()
+        if not obstruction_points_valid(self.obstruction_points):
+            self.obstruction_points = []
+
+    def interpolate(self, A, B, az):
```
wtgee: Variable names should be lowercase and `a` and `b` are a little ambiguous since this is a tuple of points. Let's name them `point_a` and `point_b` or something more descriptive.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+
+        Args:
+            az (float or int): the target azimuth
+        """
+
+        el = 0
+        prior_point = self.obstruction_points[0]
+        i = 1
+        found = False
+        while(i < len(self.obstruction_points) and found is False):
```
wtgee: `while` loops are generally used when you have continuous looping until some breaking condition. Here you have a set number of points you want to loop over so might as well loop them directly, something like:
```python
for i, point in enumerate(self.obstruction_points):
    if found:
        break
    ....
```
Then just use `point` instead of `next_point` assignment below.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+        veto = False
+        score = self._score
+
+        az = observer.altaz(time, target=target).az
+        alt = observer.altaz(time, target=target).alt
+
+        el = self.determine_el(az)
+
+        # Determine if the target altitude is above or below the determined minimum elevation for that azimuth
+        if alt - 7.5 > el:
```
wtgee: What is this magic number?

pocs/tests/test_constraints.py
```diff
@@ -197,3 +197,30 @@ def test_moon_avoidance(observer):
...
+    time = Time('2016-08-13 10:00:00')
+
+    observation1 = Observation(Field('HD189733', '20h00m43.7135s +22d42m39.0645s'))  # HD189733
+    observation2 = Observation(Field('Hat-P-16', '00h38m17.59s +42d27m47.2s'))  # Hat-P-16
+
+    veto1, score1 = h.get_score(time, observer, observation1)
+    veto2, score2 = h.get_score(time, observer, observation2)
+
+    assert veto1 is False and veto2 is False
+    assert score2 > score1 #Are these scores still relevant? How can I test scores for my class as its either 100 or 0?
```
wtgee: I would just test one observation at at time and see if `score1 > 0` or something. Depends on what you are actually trying to test. Let me know if that doesn't seem to make sense.

pocs/tests/test_horizon_limits.py
```diff
@@ -0,0 +1,263 @@
...
+        validation_7(obstruction_points)
+    except AssertionError:
+        print("obstruction_points should not consecutively have the same azimuth thrice or more")
+        v = False
+    assert v
+    return v
+
+
+# Unit tests for each of the evaluations
+def test_validations():
```
wtgee: I'm a bit confused as to what is going on here. The `validation_x` subs should be given names that make sense. If `validation_1` is testing that there is only one point it should be called `test_single_point` and a `test_no_points_fail` or something like that. Maybe we can sit down and go over a little bit of this in person.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +165,160 @@ def get_score(self, time, observer, observation, **kwargs):
...
+        binary = image > thresh
+
+        # Compute the Canny filter
+        edges1 = feature.canny(binary, low_threshold=0.1, high_threshold=0.5)
+
+        # Turn into array
+        np.set_printoptions(threshold=np.nan)
+        print(edges1.astype(np.float))
+
+    def get_config_coords(self):
```
wtgee: This is reading the config file but not necessarily "getting" them since it doesn't return anything. I would rename to `lookup_config_coords` or something.

brendan-o:
Hi Wilfred,

Sorry I only just saw this!

I am currently working on a summer vacation scholarship at CSIRO so I am
too busy right now to try fix these, hopefully I will be able to find some
time early next year for it!

Hope you are well.

Cheers,
Brendan

wtgee:
> I am currently working on a summer vacation scholarship at CSIRO so I am
too busy right now to try fix these, hopefully I will be able to find some
time early next year for it!

No problem. We have a number of changes to get integrated that will take some time. We got through most of our prep work for the 1.0.0 stream of things os early next year might be a good time for it.

In the mean time we will try to test out what is there for Huntsman when we do the commissioning run. Enjoy the holidays!

jamessynge:
FYI, we should move forward with this. I need this for PAN006, which has a big obstruction in the north-west quadrant.

wtgee:
> FYI, we should move forward with this. I need this for PAN006, which has a big obstruction in the north-west quadrant.

Brendan changed a few other things and then emailed the changes to me instead of submitting to PR. I'll try to get those integrated soon but it will need additional work to get it going.

wtgee:
Closing in favor of #368

PR #99, opened 2017-11-04
Move PEAS and PACE into POCS

### LINKED ISSUE #98: Consolidate POCS, PEAS, and PACE
We want to merge the three separate repos so there is less confusion for new users as to what is going on.

wtgee:
Consolidate repos, closes #98 

This adds the files from the different repos but doesn't otherwise do any cleanup.


wtgee:
Closing this in favor of suggestion from @jamessynge above.

See panoptes/PACE#5
panoptes/PEAS#14
#104

PR #106, opened 2017-11-08
Consolidate repos

### LINKED ISSUE #98: Consolidate POCS, PEAS, and PACE

wtgee:
Closes #98 

This pulls in the histories for both PACE and PEAS


jamessynge:
Can we hold off on merging this into develop until we've got PAN006 running automatically?

wtgee:
> Can we hold off on merging this into develop until we've got PAN006 running automatically?

I'm thinking about merging it into a separate branch but I guess not totally necessary. My other thought is that we should get `develop` to a decent state soon and merge it into `master` and switch PAN006 to that. Then we can do "hotfixes" on `master` and move those back to develop. Thoughts?

jamessynge:
I think that starting a new branch for the merge makes sense. It will probably take a while to get things in order, so a separate branch would be best.

And I think think the idea of updating master periodically, along with using tags, also makes sense. That way we can tell builders to use a specific tag, or ask them to use master to test a hotfix.

wtgee:
Replaced by #124 which merges into new branch.

Issue #108, opened 2017-11-10
POCS Shell Init


wtgee:
The startup of `pocs_shell` can be very slow, which can present issues when testing or doing polar alignment as it constantly needs to be restarted. Determine if there is a sane way to mitigate this.

wtgee:
As per @jamessynge "read all gphoto2 settings, only change those that are wrong"

PR #109, opened 2017-11-11
gphoto2 cameras

wtgee:
* Add a method to set a number of properties (by index) all at once
* Do not set the `artist` and `copyright` every time we start (will probably
need to work out how best to incorporate `first_init` into overall unit setup)

Part of #108 

:warning: This has been minimally tested with hardware but did work without problems during limited testing.

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -311,6 +311,22 @@ class AbstractGPhotoCamera(AbstractCamera):  # pragma: no cover
         # Forces the command to wait
         self.get_command_result()
 
+    def set_property_values(self, prop_vals):
+        """ Sets a number of properties all at once
+
+        Args:
+            prop_vals (dict): A dict with keys corresponding to the property to
+            be set and values corresponding to the index option
+        """
+        set_cmd = list()
+        for prop, val in prop_vals.items():
+            set_cmd.append('--set-config-index {}={}'.format(prop, val))
+
+        self.command(set_cmd)
+
+        # Forces the command to wait
+        self.get_command_result()
+
     def get_property(self, prop):
         """ Gets a property from the camera """
         set_cmd = ['--get-config', '{}'.format(prop)]
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -311,16 +311,32 @@ class AbstractGPhotoCamera(AbstractCamera):  # pragma: no cover
         # Forces the command to wait
         self.get_command_result()
 
-    def set_property_values(self, prop_vals):
-        """ Sets a number of properties all at once
+    def set_properties_by_index(self, prop_indices):
+        """ Sets a number of properties all at once by index
 
         Args:
-            prop_vals (dict): A dict with keys corresponding to the property to
+            prop_indices (dict): A dict with keys corresponding to the property to
             be set and values corresponding to the index option
         """
         set_cmd = list()
+        for prop, val in prop_indices.items():
+            set_cmd.extend(["--set-config-index", "{}={}".format(prop, val)])
+
+        self.command(set_cmd)
+
+        # Forces the command to wait
+        self.get_command_result()
+
+    def set_properties_by_value(self, prop_vals):
+        """ Sets a number of properties all at once by value
+
+        Args:
+            prop_vals (dict): A dict with keys corresponding to the property to
+            be set and values corresponding to the value
+        """
+        set_cmd = list()
         for prop, val in prop_vals.items():
-            set_cmd.append('--set-config-index {}={}'.format(prop, val))
+            set_cmd.extend(["--set-config-value", "{}='{}'".format(prop, val)])
 
         self.command(set_cmd)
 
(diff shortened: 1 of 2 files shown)


pocs/camera/camera.py
```diff
@@ -311,6 +311,22 @@ def set_property(self, prop, val):
...
+    def set_property_values(self, prop_vals):
+        """ Sets a number of properties all at once
+
+        Args:
+            prop_vals (dict): A dict with keys corresponding to the property to
+            be set and values corresponding to the index option
+        """
+        set_cmd = list()
+        for prop, val in prop_vals.items():
+            set_cmd.append('--set-config-index {}={}'.format(prop, val))
```
jamessynge: Is there a need to escape either the prop or val, or to otherwise enclose in quotes?

jamessynge: I note that the set_property function has the flag (--set-config) in a separate list entry from the prop=val entry.

pocs/camera/canon_gphoto2.py
```diff
@@ -33,24 +33,29 @@ def connect(self):
...
+            '/main/capturesettings/drivemode': 0,
+            '/main/capturesettings/shutterspeed': 0,
+        }
+
+        self.set_property_values(properties)
+
+        if first_init:
+            self.set_property('/main/settings/artist', 'Project PANOPTES')
+            self.set_property('/main/settings/ownername', 'Project PANOPTES')
+            self.set_property('/main/settings/copyright', 'Project PANOPTES 2016')
```
jamessynge: Perhaps we should use the current year? E.g. use datetime.today().year to provide this value.

pocs/camera/canon_gphoto2.py
```diff
@@ -33,24 +33,29 @@ def connect(self):
...
+            '/main/capturesettings/focusmode': 0,
+            '/main/capturesettings/continuousaf': 0,
+            '/main/capturesettings/autoexposuremode': 3,
+            '/main/capturesettings/drivemode': 0,
+            '/main/capturesettings/shutterspeed': 0,
+        }
+
+        self.set_property_values(properties)
+
+        if first_init:
```
jamessynge: Why not include these 3 properties with the ones above?

pocs/camera/canon_gphoto2.py
```diff
@@ -33,24 +33,29 @@ def connect(self):
...
 
-        self.set_property('/main/actions/viewfinder', 1)       # Screen off
-        self.set_property('/main/settings/autopoweroff', 0)     # Don't power off
-        self.set_property('/main/settings/reviewtime', 0)       # Screen off
-        self.set_property('/main/settings/capturetarget', 0)    # Internal RAM (for download)
-        self.set_property('/main/settings/artist', 'Project PANOPTES')
-        self.set_property('/main/settings/ownername', 'Project PANOPTES')
-        self.set_property('/main/settings/copyright', 'Project PANOPTES 2016')
-        self.set_property('/main/imgsettings/imageformat', 9)       # RAW
-        self.set_property('/main/imgsettings/imageformatsd', 9)     # RAW
```
jamessynge: Please keep the comments that explain each value.

pocs/camera/camera.py
wtgee: These all get passed to a `subprocess` which accepts a list of options for the command.  It could probably be set up either way, it would just require to `append` statements here. The original `set_property` only has two options so it just constructs it as a list to begin with and avoids the `append`. Would be worth investigating if there is any practical difference in the end.

pocs/camera/canon_gphoto2.py
wtgee: Oops, was lazy and doing it quickly. Will get this fixed.

pocs/camera/canon_gphoto2.py
wtgee: Because they need to be set with `--set-config-value` (rather than `-index`) so I was going to separate them and then build another bulk set by value method but it seemed a bit overkill. Ideally I think we want some kind of one-time setup of hardware script that would run this as it is a little redundant to run each time. Not sure yet what the best way to do that is.

pocs/camera/canon_gphoto2.py
wtgee: :+1: I am also probably going to switch the `artist` to the unit name.

pocs/camera/camera.py
wtgee: This method is only setting the index and worked with minimal testing, but the property here can be a `/` delimited string so might make sense. I'll test out more when I get back to lab later today.

pocs/camera/canon_gphoto2.py
jamessynge: Sounds good.

pocs/camera/canon_gphoto2.py
jamessynge: I think just setting them each time is fine, esp. if we have a way to do it "relatively" fast. Perhaps a method that takes two maps, one with property to index, and the other with property to value? Then they can all be done in a single call.

pocs/camera/camera.py
jamessynge: I realize that I haven't checked whether shell=True is being passed into the subprocess, which determines whether that array is concatenated into a single string, or is treated as a raw argv.

wtgee:
@jamessynge I kept the by-value and by-index separate, which adds another 1-2 seconds to the init but I think is cleaner.

wtgee:
Closing, superseded by #114

PR #113, opened 2017-11-12
Use panoptes installation of solve-field (astrometry).

jamessynge:


jamessynge:
Added code to issue an error if missing solve-field.

wtgee:
Actually, with this in place we should probably capture and handle as well.

https://github.com/panoptes/POCS/blob/develop/pocs/camera/canon_gphoto2.py#L178

jamessynge:
Sorry, I don't understand the comment about "capture and handle".

wtgee:
> Sorry, I don't understand the comment about "capture and handle".

Probably because I linked to wrong place. See #116

PR #114, opened 2017-11-12
Set gphoto2 faster

jamessynge:
Set multiple gphoto2 properties in a single command, and support setting by both index or value.
Fix regular expression literals that should have been raw-string-literals (i.e. needed r prefix).

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -331,22 +331,6 @@ class AbstractGPhotoCamera(AbstractCamera):  # pragma: no cover
         # Forces the command to wait
         self.get_command_result()
 
-    def set_property_values(self, prop_vals):
-        """ Sets a number of properties all at once
-
-        Args:
-            prop_vals (dict): A dict with keys corresponding to the property to
-            be set and values corresponding to the index option
-        """
-        set_cmd = list()
-        for prop, val in prop_vals.items():
-            set_cmd.append('--set-config-index {}={}'.format(prop, val))
-
-        self.command(set_cmd)
-
-        # Forces the command to wait
-        self.get_command_result()
-
     def get_property(self, prop):
         """ Gets a property from the camera """
         set_cmd = ['--get-config', '{}'.format(prop)]
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -263,7 +263,7 @@ class AbstractGPhotoCamera(AbstractCamera):  # pragma: no cover
 
             try:
                 self._proc = subprocess.Popen(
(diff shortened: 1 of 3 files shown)


pocs/camera/camera.py
```diff
@@ -311,6 +311,42 @@ def set_property(self, prop, val):
...
+            set_cmd.extend(['--set-config-index', '{}={}'.format(prop, val)])
+        for prop, val in prop2value.items():
+            set_cmd.extend(['--set-config-value', '{}={}'.format(prop, val)])
+
+        self.command(set_cmd)
+
+        # Forces the command to wait
+        self.get_command_result()
+
+    def set_property_values(self, prop_vals):
```
wtgee: I think we can just drop this method if we have the above and the individual `set_property`

pocs/camera/canon_gphoto2.py
```diff
@@ -1,3 +1,4 @@
+import datetime
```
wtgee: Not terribly important here, but we try to make all date/time items come from `current_time` (which allows us to override time in a simulator mode). Could be argued that here you don't ever want to do that and we always want the current.

pocs/camera/canon_gphoto2.py
```diff
@@ -33,25 +34,29 @@ def connect(self):
...
+            '/main/capturesettings/focusmode': 0,         # Manual (don't try to focus)
+            '/main/capturesettings/continuousaf': 0,      # No auto-focus
+            '/main/capturesettings/autoexposuremode': 3,  # 3 - Manual; 4 - Bulb
+            '/main/capturesettings/drivemode': 0,         # Single exposure
+            '/main/capturesettings/shutterspeed': 0,      # Bulb
+        }
+        prop2value = {
+            '/main/settings/artist': 'Project PANOPTES',
+            '/main/settings/ownername': 'Project PANOPTES',
+            '/main/settings/copyright': 'Project PANOPTES {}'.format(datetime.datetime.today().year),
```
wtgee: As per comment above, this can be:
```
current_time().datetime.year
```

pocs/camera/camera.py
jamessynge: Done.

pocs/camera/canon_gphoto2.py
jamessynge: Done.

pocs/camera/canon_gphoto2.py
jamessynge: Done.

jamessynge:
Fixed trailing whitespace.

PR #116, opened 2017-11-13
Capture missing solve-field error

wtgee:
* Exit with code 127 if solve-field not found
* Catch error from python and show appropriate message

Related to #113

### DIFF
--- a/scripts/solve_field.sh
+++ b/scripts/solve_field.sh
@@ -6,7 +6,7 @@ then
   echo 2>&1 ""
   echo 2>&1 "Please install astrometry into ${PANDIR}/astrometry to enable $0"
   echo 2>&1 ""
-  exit 127 # Command not found
+  exit 3 # Command not found
 fi
 
 if [[ $# == 1 ]]; then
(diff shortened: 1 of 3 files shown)


scripts/solve_field.sh
```diff
@@ -6,7 +6,7 @@ then
   echo 2>&1 ""
   echo 2>&1 "Please install astrometry into ${PANDIR}/astrometry to enable $0"
   echo 2>&1 ""
-  exit 1
+  exit 127 # Command not found
```
wtgee: I don't actually know if this is the appropriate use of the error code 127 so don't mind switching back to `1` as long as we still catch above.

jamessynge: You might be better to go with 3. Codes 1, 2, 126 and 127 have meanings assigned by bash, so avoiding those codes may avoid confusion down the road. See: http://tldp.org/LDP/abs/html/exitcodes.html

wtgee:
I'll get this changed back. I also am not sure why I mixed the `requirements.txt` update into this but will probably remove.

PR #127, opened 2017-11-22
Add AbstractDome, base class for Dome controller classes.

jamessynge:
Converted the existing bisque dome to use that as its base class.
Converted existing bisque/test_dome.py to use slightly altered API.
Added a simplistic dome simulator and test.

Long term we may want an API that supports asynchronous activity (moving the dome takes time) and safety info (e.g. whether the dome is hooked up to a rain sensor, or whether the dome requires that mount to be parked before it can open or close, if that is necessary to avoid collision).

### DIFF
--- a/pocs/dome/__init__.py
+++ b/pocs/dome/__init__.py
@@ -22,7 +22,7 @@ class PanFixedDome(PanBase):
 
     def __init__(self, *args, **kwargs):
         """Initialize a PanFixedDome, no connected to the underlying device.
-        
+
         Customization generally comes from the config file, so that the
         caller doesn't need to know the params needed by a specific type of
         dome interface class."""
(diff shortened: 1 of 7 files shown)


jamessynge:
This is in preparation for #134.

pocs/dome/__init__.py
```diff
@@ -0,0 +1,134 @@
+from abc import ABCMeta, abstractmethod, abstractproperty
+
+from .. import PanBase
+from ..utils import load_module
+from ..utils.logger import get_root_logger
+
+# TODO(jamessynge): Discuss with Wilfred how to handle long running operations
```
wtgee: Can we remove this and place somewhere else?

pocs/dome/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+        return None
+    dome_config = config['dome']
+    if 'dome' in config.get('simulator', []):
+        brand = 'simulator'
+        driver = 'simulator'
+        dome_config['simulator'] = True
+    else:
+        brand = dome_config.get('brand')
+        driver = dome_config['driver']
+    logger = get_root_logger()
```
wtgee: In here twice.

Issue #115, opened 2017-11-13
Make install of software as simple as possible.


wtgee:
Related:

- [x] #94
- [x] ~~#27~~
- [x] #77 
- [ ] #87 
- [x] #26
- [x] #97

jamessynge:
See https://github.com/jamessynge/POCS/tree/install where I'm working on an install script.

wtgee:
#77 and #87 will still apply, but this will be closed by #951.

wtgee:
Mostly replaced by the guide at projectpanoptes.org

Issue #121, opened 2017-11-19
Don't download IRES files every time tests are run


wtgee:
In general we need a better mechanism for checking if these are outdated. The [astroplan documentation](https://astroplan.readthedocs.io/en/latest/) might provide clues as to the best way to achieve this.

jamessynge:
Let's consider changing utils/data.py to download the file at most every week (for example), storing a copy in /var/panoptes/data whose date it can check.

wtgee:
We can do this but I think that we also want to just install a cron job that performs this action once per week regardless. Could actually be a part of #129 when installed for normal units.

Issue #97, opened 2017-11-03
Startup scripts


wtgee:
Create some startup scripts, based off what @jamessynge has put together.

- Ensure astroplan data is updated 
- Mongo started
- Messenger started (zeromq forwarder)
- Website started (PAWS)
- Environment sensing started (PEAS) 

Supersedes #26

jamessynge:
I've written scripts that run at boot of PAN006 which launches tmux as panoptes user with necessary programs running. Will work with you to get those submitted and updated to be general.

wtgee:
I think most of this has been handled by @jamessynge. Closing this issue. Further items should have their own Issue.

PR #129, opened 2017-11-24
Install script for panoptes dependencies

jamessynge:
This installs:
- Linux packages
- Mongodb (including starting the service, but not setting up the panoptes db)
- Anaconda (miniconda)
- Astrometry (builds software and fetches indices)
- Python packages (pip)

Has code for building and installing latest version of cfitsio, but that is currently disabled in favor of using the latest linux package for cfitsio. Avoiding the build because I was finding some problems with using astrometry.

Partially addresses #115. Would still like to include clone of git repos, and properly configuring linux for ssh, mongodb for security, etc.


jamessynge:
Another TODO: it would be nice to not install astrometry if it hasn't changed, but they don't seem to expose a convenient way to determine the installed version. Sigh. May work around by writing a version number file from the install script.

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
...
+  sudo apt-get install -y mongodb-org
+  echo "[Unit]
+Description=High-performance, schema-free document-oriented database
+After=network.target
+Documentation=https://docs.mongodb.org/manual
+
+[Service]
+User=mongodb
+Group=mongodb
+ExecStart=/usr/bin/mongod --quiet --config /etc/mongod.conf
```
wtgee: We should also specify a `--dbpath` (or alter the `mongod.conf`). I've put this in `$PANDIR/data` previously. Our current setup writes a lot of entries to the database and it grows very large (usually needs > 4GB to run). On PAN001 all of $PANDIR is on a separate mount from the rest of `/var` (and everything else)

wtgee: Note also that mongo does not require you to actually create a collection (read: database) before using it. You can just start writing to something and it will store properly

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
...
+  fi
+}
+
+# Install miniconda (conda with no additional packages); we can then select
+# the set to install.
+function install_conda() {
+  local -r the_script="${PANDIR}/tmp/miniconda.sh"
+  echo_bar
+  echo
+  echo "Installing miniconda"
```
wtgee: Just to be good sports we should mention here that there is a default acceptance of the license agreement.

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
...
+  rm -rf "${SCRATCH_DIR}"
+}
+
+function install_astrometry_indices() {
+  mkdir -p "${PANDIR}/astrometry/data"
+  cd "${PANDIR}/astrometry/data"
+  # See http://astrometry.net/doc/readme.html#getting-index-files
+  # for info on the index files you might need.
+  if [[ "${DO_ASTROMETRY_FULL_INDICES}" -eq 1 ]] ; then
+    if [[ ! -f index-4107.fits ]] ; then
```
wtgee: Just a note that if download/install gets interrupted this will not work properly second time around.

We also have `$POCS/pocs/utils/data.py` that does this same checking and is run before tests.

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
...
+  (install_astrometry)
+fi
+
+
+if [[ "${DO_ASTROMETRY_INDICES}" -eq 1 ]] ; then
+  (install_astrometry_indices)
+fi
+
+
+if [[ "${DO_PIP_REQUIREMENTS}" -eq 1 ]] ; then
```
wtgee: Let's actually create a separate panoptes environment in conda and use that. I've just started calling mine `panoptes-env`. So we will want to create it, install pip items there, and optionally set it as the default environment for the system (via `~/.profile`).  Something like:

```bash
$ conda create -y -n panoptes-env python=3
# Wait
$ source activate panoptes-env
(panoptes-env) $ pip install -U pip
(panoptes-env) $ pip install -r requirements.txt
(panoptes-env) $ echo "source activate panoptes-env" >> ~/.profile # Assume we always want
```
Note also we can create a conda environment file that can do all this and have the right collection of files. That is what I was doing in the consolidated branch but was having hassle. We should get this install script working first and then later worry about the environment file. But I do think we want to create/use a specific panoptes environment as outlined above.

wtgee:
> Another TODO: it would be nice to not install astrometry if it hasn't changed, but they don't seem to expose a convenient way to determine the installed version. Sigh. May work around by writing a version number file from the install script.

We should submit a patch to astrometry.net to do this. It annoys me when a `--version` flag outputs the `--help`.

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+libjpeg-dev
+libnetpbm10-dev
+libpng12-dev
+netpbm
+python-dev
+python-numpy
+python-pyfits
+swig
+zlib1g-dev
+
```
wtgee: Looks like `byobu` comes with Ubuntu server but not desktop. Let's put it in here to make sure we have it. It will pull in `tmux` also (I think we should default to byobu but will make comments in #97 )

.travis.yml
```diff
@@ -59,7 +59,7 @@ install:
   - source activate test-environment
 
   - cd $TRAVIS_BUILD_DIR
-  - pip install -r requirements.txt
+  - pip install -r scripts/install/requirements.txt
```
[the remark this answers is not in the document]
jamessynge: Done.

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
+# This file contains a list of apt packages to install.
+# Lines starting with # are comments as are blank lines,
+# and # starts trailing comments.
+
+# dcraw decodes Canon raw images, and gphoto2 takes pictures with
+# Canon DSLRs, among others.
+dcraw
+gphoto2
```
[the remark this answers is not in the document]
jamessynge: I can add a todo, but if we're happy enough with the package version, I think we should stick with it for simplicity.

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+# We use cfitsio command line tools (e.g. fpack) from POCS, in addition to the
+# uses of the cfitsio libraries by astrometry above.
+libcfitsio-bin
+
+# Linux monitoring sensors provides access to hardward sensors in the computer,
+# such as CPU or GPU temperature. May require bios or other changes to make use
+# of these.
+# TODO(jamessynge): Discuss these with Wilfred as an additional way to get
+# temperature data.
+# lm-sensors
```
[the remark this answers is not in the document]
jamessynge: Added an issue.

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+
+# These are used if you want to setup SSH access into the computer.
+# openssh-client
+# openssh-server
```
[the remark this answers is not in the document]
jamessynge: Done

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+
+# Graphviz is used for rendering the state machine of POCS.
+# TODO(wtgee): Are there any other uses?
+graphviz
+libgraphviz-dev
```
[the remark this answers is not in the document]
jamessynge: Removed TODO.

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+
+# Linux tool used during building and installing software (e.g. astrometry).
+pkg-config
+
+# Cross platform python library for getting system info.
+# TODO(wtgee): Is this used by any tools we use? It isn't directly used by
+# POCS et al.
+# python-psutil
```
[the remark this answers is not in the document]
jamessynge: Removed.

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+
+# These additional packages may or may not be needed, I simply installed them on
+# the NUC for PAN006 and need to determine if they are needed. Commenting them
+# out for now until a reason for their presence is established.
+
+# iw  # Wireless networking config & info.
+
+# libgeos-3.*
+# libgeos-c1v5
+# libgeos-dev
```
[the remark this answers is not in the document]
jamessynge: Should we have a separate installer for PIAA?

scripts/install/apt-packages-list.txt
```diff
@@ -0,0 +1,71 @@
...
+# libncurses5-dev
```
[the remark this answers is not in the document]
jamessynge: How is curses accessed? Via readline?

scripts/install/apt-packages-list.txt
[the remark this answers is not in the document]
wtgee: > Should we have a separate installer for PIAA?

Yes, I think so. I imagine there will be people who want to work on the processing without necessarily caring about POCS (although some things, like `pocs/utils/images.py` are also used by PIAA so this needs some cleanup).

scripts/install/apt-packages-list.txt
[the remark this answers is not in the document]
wtgee: Yes, I believe so although not entirely sure. I feel like I almost always end up needing ncurses for some reason so almost always have it on my system. We could try leaving it out until we discover where it is actually needed.

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
+#!/bin/bash -e
+
+THIS_DIR="$(dirname $(readlink -f "${0}"))"
+THIS_PROGRAM="$(basename "${0}")"
+
+if [[ -z "${PANDIR}" || -z "${POCS}" || -z "${PAWS}" || -z "${PANLOG}" ||
+      -z "${PANUSER}" || ! -d "${PANDIR}" ]] ; then
```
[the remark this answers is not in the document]
jamessynge: Fixed.

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
...
+
+function echo_bar() {
+  eval $(resize|grep COLUMNS=)
+  printf "%${COLUMNS:-80}s\n" | tr ' ' '#'
+}
+
+# Append $1 to PATH and write command to do the same to .bashrc.
+function add_to_PATH() {
+  local -r the_dir="$(readlink -f "${1}")"
+  if [[ -z "$(egrep -- "PATH=.*${the_dir}" ~/.bashrc)" ]] ; then
```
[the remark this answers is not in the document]
jamessynge: Good idea.

scripts/install/install-dependencies.sh
```diff
@@ -0,0 +1,394 @@
...
+  # names are left.
+  APT_PKGS="$(cut '-d#' -f1 "${THIS_DIR}/apt-packages-list.txt" | sort | uniq)"
+  echo
+  echo "Running sudo apt-get install, you may be prompted for your password."
+  echo
+  (set -x ; sudo apt-get update)
+  (set -x ; sudo apt-get install --yes ${APT_PKGS})
+}
+
+function install_mongodb() {
```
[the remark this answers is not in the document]
jamessynge: Agreed.

scripts/install/install-dependencies.sh
jamessynge: Not surprisingly, I didn't know that info about --dbpath.
Does mongodb allow multiple dbpath directories (i.e. can we arrange for our collections to be in /var/panoptes)?

No change made yet.

scripts/install/install-dependencies.sh
jamessynge: Added link to license.

scripts/install/install-dependencies.sh
wtgee: There is some info (for instance, [here](https://docs.mongodb.com/manual/administration/configuration/#run-multiple-database-instances-on-the-same-system)) but I think the better answer would be to run what we want in a container.

scripts/install/install-dependencies.sh
jamessynge: I notice that it installs more index files.

Do you have a suggestion regarding the interruption? Explicitly do the download of each?
Or just run $POCS/pocs/utils/data.py near the end of the install.

FWIW, I'd like to eliminate all downloading from tests.

scripts/install/apt-packages-list.txt
jamessynge: Added byobu & tmux (explicitly).

scripts/install/install-dependencies.sh
wtgee: The tests shouldn't download but they rely on having the files previously downloaded. As of right now they download [here](https://github.com/panoptes/POCS/blob/develop/pocs/tests/conftest.py#L9).

The astroplan IRES stuff still needs to be figured out better as per #121

scripts/install/install-dependencies.sh
jamessynge: Why do you say "# Wait" after conda create?

wtgee: I just meant that it takes a minute and there is output between those two steps.

jamessynge: Done (well, no wait included).

jamessynge:
Closing and re-opening PR to trigger another travis build. I reviewed the failure, and it smells like a problem interacting with temp files and subprocesses, rather than with the commits, so I want to see if it is reproducible as is.

jamessynge:
Well, that was a false hope. The failure was essentially identical:

> =================================== FAILURES ===================================
> __________________________ test_solve_field_unsolved ___________________________
> unsolved_fits_file = '/home/travis/build/panoptes/POCS/pocs/tests/data/unsolved.fits'
>     def test_solve_field_unsolved(unsolved_fits_file):
>         im0 = Image(unsolved_fits_file)
>     
>         assert isinstance(im0, Image)
>         assert im0.wcs is None
>         assert im0.pointing is None
>     
> >       im0.solve_field(verbose=True, replace=False, radius=4)
> pocs/tests/test_images.py:79: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> pocs/images.py:174: in solve_field
>     **kwargs)
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> fname = '/home/travis/build/panoptes/POCS/pocs/tests/data/unsolved.fits'
> replace = False, remove_extras = True
> kwargs = {'dec': 44.456875, 'ra': 301.56439, 'radius': 4, 'verbose': True}
> verbose = True, out_dict = {}
> output = 'Reading input file 1 of 1: "/home/travis/build/panoptes/POCS/pocs/tests/data/unsolved.fits"...\nExtracting sources......er from file "/tmp/tmp.wcs.Wwe3cD" extension 0\n qfits:errfunc error: Failed to read FITS file "/tmp/tmp.wcs.Wwe3cD"\n'
> errs = None, proc = <subprocess.Popen object at 0x7ffb3c771320>

pocs/utils/data.py
```diff
@@ -12,18 +12,22 @@ def download_all_files(data_folder=None):
...
         if not os.path.exists(dest):
-            url = "http://data.astrometry.net/4200/{}".format(fn)
+            url = "http://data.astrometry.net/{}".format(fn)
             df = data.download_file(url)
             try:
                 shutil.move(df, dest)
             except OSError as e:
                 print("Problem saving. (Maybe permissions?): {}".format(e))
 
+    for i in range(4112, 4119):
```
wtgee: Just reproduced the test failure on my local and it's because we are missing 4110 and 4111 index files. The fits files we use for testing have a small field-of-view, which is what those index files will search. 10 and 11 are fairly large (24MB and 10MB) so this is a bit annoying as we only really need them for the tests.

I also removed the 42xx from my local and all tests still pass. The 41xx are wide-field so should be all that we need although I suppose it doesn't hurt to have all of them.

jamessynge: OK. Can we replace the file used for testing to a wide field also?

jamessynge:
I've discovered that writing to .profile didn't leave me with a working profile (activate was apparently in the wrong place). Will investigate further.

PR #133, opened 2017-11-26
Fix call to subprocess.Popen

jamessynge:
cmd is supposed to be a list of args, so shouldn't join options into a single string.
Add some verbose output to images.py.

### DIFF
--- a/pocs/utils/images.py
+++ b/pocs/utils/images.py
@@ -75,7 +75,7 @@ def solve_field(fname, timeout=15, solve_opts=[], **kwargs):
 
     cmd = [solve_field_script] + options + [fname]
     if verbose:
-        print("Cmd: ", cmd)
+        print("Cmd:", cmd)
 
     try:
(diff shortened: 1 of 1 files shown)


pocs/utils/images.py
```diff
@@ -137,13 +140,14 @@ def get_solve_field(fname, replace=True, remove_extras=True, **kwargs):
         proc.kill()
         raise error.Timeout("Timeout while solving")
     else:
-        if proc.returncode == 3:
-            raise error.SolveError('solve-field not found: {}'.format(output))
-
         if verbose:
+            print("Returncode: {}", proc.returncode)
```
wtgee: Looks like I meant these print statements to have a `format` to go with the `{}`. Let's either add format or remove curly braces. I realize you were probably just copying what was there but since in this part of the code right now let's get changed.

jamessynge: Will fix in the AM. Have a good Sunday.

jamessynge:
Fixed print statements.

PR #144, opened 2017-11-28
With 100 char limit 

wtgee:
Another lazy way to compare.

### DIFF
--- a/pocs/camera/sbigudrv.py
+++ b/pocs/camera/sbigudrv.py
@@ -103,7 +103,9 @@ class SBIGDriver(PanBase):
         # Reopen driver ready for next command
         self._send_command('CC_OPEN_DRIVER')
 
-        self.logger.info('\t\t\t SBIGDriver initialised: found {} cameras'.format(self._camera_info.camerasFound))
+        self.logger.info(
+            '\t\t\t SBIGDriver initialised: found {} cameras'.format(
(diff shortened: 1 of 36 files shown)


pocs/camera/sbigudrv.py
```diff
@@ -162,7 +165,8 @@ def assign_handle(self, serial=None):
 
         # Serial number, name and type should match with those from Query USB Info obtained earlier
         camera_serial = str(self._camera_info.usbInfo[index].serialNumber, encoding='ascii')
-        assert camera_serial == ccd_info['serial_number'], self.logger.error('Serial number mismatch!')
+        assert camera_serial == ccd_info['serial_number'], self.logger.error(
```
jamessynge: FWIW, this seems like an odd way to test things. assert isn't always executed, IIUC.

wtgee:
I started getting rid of some of the `format` calls in the logger statements but only got half way through before decided probably wasn't totally neccesarry.

I'd like to get this or the 80 char PR merged as soon as possible. My preference is still for 100 (this PR) so unless there are major objections will want to merge within 1 day.

jamessynge:
Let's go with 100, and we can see if that works well enough for us.


On Thu, Nov 30, 2017 at 6:19 AM, Wilfred Tyler Gee <notifications@github.com

PR #145, opened 2017-11-28
Log file updates

### LINKED ISSUE #96: Log to unique filename by program and time
Running peas_shell logs to panoptes.log, as does pocs_shell. They should each log to a unique file name.
And it would help debugging if each run created a new file, e.g. with a timestamp in the name.
To make it easier to find the latest log file, create a symlink from <progname>.log to <progname>.<timestamp>.log.

wtgee:
* Support use of `{}` in logger statements so we can avoid superfluous `format` calls
* Update CONTRIBUTING doc for logging
* Update log file name (#96)

Will probably need to do something in the future to clean up old log files.

Long-running (i.e. multi-day) programs will still be rotated daily at 11:30 am local. These will carry an additional day stamp.

Generates log files with current PID and a utc timestamp:
```
$ ls /var/panoptes/logs
panoptes.log -> /var/panoptes/logs/panoptes-26947-20171128T035245Z.log
panoptes-26947-20171128T035245Z.log
```

Note that this is being merged against `1.0-beta`.

Closes #96

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,5 +46,15 @@ playground rules and follow them during all your contributions.
       + WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but operation of unit can continue.
       + ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot continue.
    + The logger support variable information without the use of the `format` method. Examples:
-      * `self.logger.info("Welcome {}", self.config['name'])`
-      * `self.logger.debug("Connection to camera {} on {}", cam_num, cam_port)`
+      
+```
+self.logger.info("Starting Exposure {} of {}", i, num_exposures)
+self.logger.debug("Taking {} second exposure", num_seconds)
+self.logger.debug("Observation: {}: {} exposures in blocks of {}, minimum {}, priority {}",
+   obs_name,
+   exp_time,
+   block_size,
+   block_min_size,
+   priority   
+)
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,17 +38,36 @@ playground rules and follow them during all your contributions.
 - Do not leave in commented-out code or unnecessary whitespace.
 - Variable/function/class and file names should be meaningful and descriptive.
 - File names should be underscored, not contain spaces ex. my_file.py.
-- Define any project specific terminology or abbreviations you use in the file you use them.
+- Define any project specific terminology or abbreviations you use in the file where you use them.
 
 # Log Messages
 
 Use appropriate logging:
 - Log level:
-   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a person actively watching a running unit.
    - DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
+   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a person actively watching a running unit.
    - WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but operation of unit can continue.
    - ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot continue.
-- The logger supports variable information without the use of the `format` method. See examples in screenshot below.
+- The logger supports variable information without the use of the `format` method.
+- There is a `say` method that is meant to be used in friendly manner to convey information to a user. This should be used only for personable output and is typically displayed in the "chat box" of the PAWS website. These messages are also sent to the INFO level logger
+
+Logging examples:
+__Note__ These are meant to illustrate the logging calls and are not necessarily indicative of real operation
+
+```
+self.logger.info("PANOPTES unit initialized: {}", self.config['name'])
+self.say("I'm all ready to go, first checking the weather")
+self.logger.debug("Setting up weather station")
+self.logger.warning('Problem getting wind safety: {}'.format(e))
+self.logger.debug("Rain: {} Clouds: {} Dark: {}",
+   is_raining,
+   is_cloudy,
+   is_dark
+)
+self.logger.error('Unable to connect to AAG Cloud Sensor, cannot continue')
+```
+
+
 - The [`grc`](https://github.com/garabik/grc) (generic colouriser) can be used with `tail` to get pretty log files. The following screenshot shows commands entered into a `jupyter-console` in the top panel and the log file in the bottom panel.
 
 <p align="center">
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,22 +39,18 @@ playground rules and follow them during all your contributions.
 - Variable/function/class and file names should be meaningful and descriptive.
 - File names should be underscored, not contain spaces ex. my_file.py.
 - Define any project specific terminology or abbreviations you use in the file you use them.
-- Use appropriate logging:
-   + Log level:
-      + INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a person actively watching a running unit.
-      + DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
-      + WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but operation of unit can continue.
-      + ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot continue.
-   + The logger support variable information without the use of the `format` method. Examples:
-      
-```
-self.logger.info("Starting Exposure {} of {}", i, num_exposures)
-self.logger.debug("Taking {} second exposure", num_seconds)
-self.logger.debug("Observation: {}: {} exposures in blocks of {}, minimum {}, priority {}",
-   obs_name,
-   exp_time,
-   block_size,
-   block_min_size,
-   priority   
-)
-```
+
+# Log Messages
+
+Use appropriate logging:
+- Log level:
+   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a person actively watching a running unit.
+   - DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
+   - WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but operation of unit can continue.
+   - ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot continue.
+- The logger supports variable information without the use of the `format` method. See examples in screenshot below.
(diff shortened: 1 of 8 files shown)


pocs/utils/logger.py
```diff
@@ -8,6 +8,33 @@
...
+
+        return log_str
+
+    def debug(self, fmt, *args, **kwargs):
+        self.logger.debug(self._process_str(fmt, *args, **kwargs))
+
+    def info(self, fmt, *args, **kwargs):
+        self.logger.info(self._process_str(fmt, *args, **kwargs))
+
+    def warning(self, fmt, *args, **kwargs):
```
jamessynge: warn rather than warning? Or change the one use of warning into warn?

pocs/utils/logger.py
```diff
@@ -18,14 +45,18 @@ def get_root_logger(profile='panoptes', log_config=None):
     # Get log info from config
     log_config = log_config if log_config else load_config('log').get('logger', {})
 
+    log_dir = '{}/logs'.format(os.getenv('PANDIR', '/var/panoptes/'))
+    log_fname = 'panoptes-{}-{}.log'.format(os.getpid(), datetime.datetime.utcnow().strftime('%Y%m%dT%H%M%SZ'))
```
jamessynge: Instead of panoptes, please use the name of the script. We don't need panoptes in the name, as the file is already in /var/panoptes.

import sys    
import os    
the_invoked_script =  os.path.basename(sys.argv[0])

pocs/utils/logger.py
```diff
@@ -18,14 +45,18 @@ def get_root_logger(profile='panoptes', log_config=None):
...
+    log_fname_generic = 'panoptes.log'
```
jamessynge: Ditto here.

CONTRIBUTING.md
```diff
@@ -34,7 +34,23 @@ playground rules and follow them during all your contributions.
...
+
+# Log Messages
+
+Use appropriate logging:
+- Log level:
+   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a person actively watching a running unit.
+   - DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
+   - WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but operation of unit can continue.
+   - ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot continue.
+- The logger supports variable information without the use of the `format` method. See examples in screenshot below.
```
jamessynge: Could you add some inline code examples, and not solely depend on the screen shot? The screen shot isn't as useful for someone reading the file in an editor.

CONTRIBUTING.md
```diff
@@ -34,7 +34,23 @@ playground rules and follow them during all your contributions.
...
 - All Python should use [PEP 8 Standards](https://www.python.org/dev/peps/pep-0008/)
    - Line length is set at 120 characters instead of 80
+   - It is recommended to have your editor auto-format code whenever you save a file rather than attempt to go back and change an entire file all at once. 
 - Do not leave in commented-out code or unnecessary whitespace.
-- Variable/function/class and file names should be meaningful and descriptive
-- File names should be underscored, not contain spaces ex. my_file.py
-- Define any project specific terminology or abbreviations you use in the file you use them
+- Variable/function/class and file names should be meaningful and descriptive.
+- File names should be underscored, not contain spaces ex. my_file.py.
+- Define any project specific terminology or abbreviations you use in the file you use them.
```
jamessynge: in the file *where* you use them.

CONTRIBUTING.md
```diff
@@ -34,7 +34,23 @@ playground rules and follow them during all your contributions.
...
+
+# Log Messages
+
+Use appropriate logging:
+- Log level:
+   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a person actively watching a running unit.
+   - DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
+   - WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but operation of unit can continue.
+   - ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot continue.
```
jamessynge: Perhaps some examples. From these descriptions, I'm guessing that the rain sensor being trigger would be at INFO level (normal event), and WARNING level might be used if we can't solve a field (e.g. due to clouds).

pocs/utils/logger.py
```diff
@@ -18,14 +45,18 @@ def get_root_logger(profile='panoptes', log_config=None):
...
     # Alter the log_config to use UTC times
     if log_config.get('use_utc', True):
         for name, formatter in log_config['formatters'].items():
             log_config['formatters'][name].setdefault('()', _UTCFormatter)
 
     log_file_lookup = {
-        'all': "{}/logs/panoptes.log".format(os.getenv('PANDIR', '/var/panoptes')),
-        'warn': "{}/logs/warnings.log".format(os.getenv('PANDIR', '/var/panoptes')),
+        'all': "{}/{}".format(log_dir, log_fname),
+        'warn': "{}/warnings.log".format(log_dir),
```
jamessynge: I'm unclear on how this lookup is used.

I'm used to a model where we do actually have 3 sets of log files for each execution, one for each level INFO, WARN and ERROR (actually, there is also a FATAL file too, created when a program is crashing). Each file contains all messages of its level, and all the more severe levels. And each file has the same name except for the text INFO, WARN (or maybe WARNING) or ERROR.

wtgee: I'm not opposed to this or to naming the file after the invoked script but we now have to have some kind of log file maintenance in place beyond the simple rotation we had before. I already have 20+ log files in the directory that are generated during development.

pocs/utils/logger.py
wtgee: I think you mean the other way around? I changed the one case of `warn` to `warning`

pocs/utils/logger.py
jamessynge: Sorry, but I laughed at that. I periodically clean up the thousands of log files in my ~/tmp directory at work.

pocs/utils/logger.py
jamessynge: OK, which ever way it is supposed to be.

pocs/utils/logger.py
wtgee: Yeah, it will only really be a problem for developers. A running unit shouldn't be switching all that much.

Regardless, I created #146

pocs/utils/logger.py
```diff
@@ -45,25 +46,34 @@ def get_root_logger(profile='panoptes', log_config=None):
     # Get log info from config
     log_config = log_config if log_config else load_config('log').get('logger', {})
 
+    invoked_script = os.path.basename(sys.argv[0])
     log_dir = '{}/logs'.format(os.getenv('PANDIR', '/var/panoptes/'))
```
jamessynge: Should we assume /var/panoptes exists? Or use the python determined temporary directory, i.e. tempfile.gettempdir()?

pocs/utils/logger.py
```diff
@@ -45,25 +46,34 @@ def get_root_logger(profile='panoptes', log_config=None):
...
-    log_fname = 'panoptes-{}-{}.log'.format(os.getpid(), datetime.datetime.utcnow().strftime('%Y%m%dT%H%M%SZ'))
-    log_fname_generic = 'panoptes.log'
+    log_fname = '{}-{}-{}'.format(invoked_script, os.getpid(),
+                                  datetime.datetime.utcnow().strftime('%Y%m%dT%H%M%SZ'))
+    log_fname_generic = '{}'.format(invoked_script)
```
jamessynge: Why is this not just log_fname_generic = invoked_script

pocs/utils/logger.py
wtgee: We assume `$PANDIR` exists. It is checked [here](https://github.com/panoptes/POCS/blob/develop/pocs/__init__.py#L71) before any logging can be loaded. So really we don't even need the `getenv` call but it was already there. Might not be a bad idea to switch it to a temp file but shouldn't really every get there.

pocs/utils/logger.py
```diff
@@ -45,25 +46,34 @@ def get_root_logger(profile='panoptes', log_config=None):
...
-        'warn': "{}/warnings.log".format(log_dir),
-    }
-
-    # Setup the TimeedRotatingFileHandler to backup in middle of day intead of middle of night
+    # Set log filename and rotation
     for handler in log_config.get('handlers', []):
-        log_config['handlers'][handler].setdefault('filename', log_file_lookup[handler])
-        if handler in ['all', 'warn']:
-            log_config['handlers'][handler].setdefault('atTime', datetime.time(hour=11, minute=30))
+        # Set the filename
```
jamessynge: Is it necessary with the Python logger system to create all of the files up front, along with those symlinks, or can we defer until the first message is written to the handler?

pocs/utils/logger.py
```diff
@@ -45,25 +46,34 @@ def get_root_logger(profile='panoptes', log_config=None):
     # Get log info from config
     log_config = log_config if log_config else load_config('log').get('logger', {})
 
+    invoked_script = os.path.basename(sys.argv[0])
     log_dir = '{}/logs'.format(os.getenv('PANDIR', '/var/panoptes/'))
-    log_fname = 'panoptes-{}-{}.log'.format(os.getpid(), datetime.datetime.utcnow().strftime('%Y%m%dT%H%M%SZ'))
-    log_fname_generic = 'panoptes.log'
+    log_fname = '{}-{}-{}'.format(invoked_script, os.getpid(),
+                                  datetime.datetime.utcnow().strftime('%Y%m%dT%H%M%SZ'))
```
jamessynge: Let me just say (whine) how much I hate timezones.

wtgee: We try to do everything UTC but I was tempted to change this to local time. All the timestamps in file are in UTC but you might want to know when the log file was generated locally. Probably not a big deal though.

pocs/utils/logger.py
wtgee: Ha, good point. This is why code reviews are good and speed-coding is bad. :)

pocs/utils/logger.py
wtgee: Note that I'm only creating the symlink for the `all` file (which I suppose should just be called `debug`). But I'm actually not sure about the actual question. We immediately start writing some messages so the `all` file is created right away.  I'll check to see order.

PR #147, opened 2017-11-29
Add testing of rs232.SerialData

jamessynge:
The latest test, test_unthreaded_io, isn't working yet, probably because of the module being loaded twice (i.e. two separate paths), and thus the global variables in the module aren't shared between the two distinct loads of the module.


pocs/utils/rs232.py
```diff
@@ -142,27 +156,26 @@ def write(self, value):
...
         return response
 
-    def read(self):
-        """
-        Reads value using readline
-        If no response is given, delay and then try to read again. Fail after 10 attempts
+    def read(self, retry_limit=5, retry_delay=0.5):
+        """Reads next line of input using readline.
+
+        If no response is given, delay for retry_delay and then try to read
```
wtgee: Better than previous docstring, but we have explicit Args section. Need to start enforcing docstring standards. I have a plugin for Sublime that fills out all the boilerplate.

http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google

pocs/utils/rs232.py
```diff
@@ -142,27 +156,26 @@ def write(self, value):
 
         return response
 
-    def read(self):
-        """
-        Reads value using readline
-        If no response is given, delay and then try to read again. Fail after 10 attempts
+    def read(self, retry_limit=5, retry_delay=0.5):
```
wtgee: Do we want these params set at the class level?

pocs/utils/rs232.py
```diff
@@ -189,13 +202,16 @@ def get_reading(self):
...
         """ Clear Response Buffer """
-        count = 0
-        while self.ser.inWaiting() > 0:
-            count += 1
-            self.ser.read(1)
-
+        # Not worrying here about bytes arriving between reading in_waiting
+        # and calling reset_input_buffer().
+        count = self.ser.in_waiting
+        self.ser.reset_input_buffer()
```
wtgee: Should this be `reset_output_buffer`? Or perhaps both.

pocs/utils/rs232.py
```diff
@@ -189,13 +202,16 @@ def get_reading(self):
...
         # self.logger.debug('Cleared {} bytes from buffer'.format(count))
 
     def __del__(self):
-        if self.ser:
+        try:
```
wtgee: I don't understand what's going on here.

pocs/tests/serial_handlers/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+
+        If you need to know which property has been changed, override the
+        setter for the appropriate properties.
+        """
+        pass
+
+    def _update_rts_state(self):
+        """Handle rts being set to some value.
+        
+        "self.rts = value" has been executed, for some value. This may not
```
wtgee: Backticks here for the day we finally get some kind of readthedocs setup going properly.

pocs/tests/serial_handlers/protocol_test.py
```diff
@@ -0,0 +1,243 @@
...
+        self._dtr = False
+        # Read-only property values
+        self._name = False
+        self._rts = False
+        self._ctx = False
+        self._dsr = False
+        self._ri = False
+        self._ri = False
+        self._ri = False
+        self._ri = False
```
wtgee: Probably don't need 4 of these.

pocs/tests/serial_handlers/protocol_test.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @break_condition.setter
+    def break_condition(self, value):
+        """Set break_condition: Controls TXD. When active, to transmitting is
+        possible."""
+        self.__break_condition = value
+        if self.logger:
+            self.logger.info('ignored setBreak({!r})'.format(level))
+
+    def setRTS(self, level=True):
```
wtgee: All these method names are deprecated in favor of properties. Should we still be using?

pocs/tests/test_rs232.py
```diff
@@ -0,0 +1,81 @@
+import pytest
+import serial  # PySerial, from https://github.com/pyserial/pyserial
```
wtgee: Why the github comment?

pocs/tests/test_rs232.py
```diff
@@ -0,0 +1,81 @@
...
+def test_basic(handler):
+    # Confirm we can create the SerialData object.
+    ser = rs232.SerialData(port='no_op://', open_delay=0)
+    # Peek inside, it should have a PySerial instance as member ser.
+    assert ser.ser
+    assert ser.ser.__class__.__name__ == 'NoOpSerial'
+    print(str(ser.ser))
+    # Open is automatically called by SerialData.
+    assert ser.is_connected
+    # Not using threading.
```
wtgee: `threaded=True` is the default on `SerialData`. Does this get unset in one of the serial_handler files above?

pocs/tests/test_rs232.py
```diff
@@ -0,0 +1,81 @@
...
+    # Peek inside, it should have a PySerial instance as member ser.
+    assert ser.ser
+    assert ser.ser.__class__.__name__ == 'BuffersSerial'
+    print(str(ser.ser))
+    # Open is automatically called by SerialData.
+    assert ser.is_connected
+    # Not using threading.
+    assert not ser.is_listening
+
+    # no_op handler doesn't do any reading or writing.
```
wtgee: Is this comment valid here?

pocs/utils/rs232.py
```diff
@@ -12,56 +12,66 @@
...
-
-            self.name = name
-            self.queue = deque([], 1)
-            self._is_listening = False
-            self.loop_delay = 2.
-
-            if self.is_threaded:
-                self._serial_io = TextIOWrapper(BufferedRWPair(self.ser, self.ser),
-                                                newline='\r\n', encoding='ascii', line_buffering=True)
+        self.ser = serial.serial_for_url(port, do_not_open=True)
```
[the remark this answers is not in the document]
jamessynge: Not sure what "this" refers to. If you mean do_not_open, the advantage is that you set all of the properties (e.g. baud rate and byte size) before opening the port. Depending on the device, it may require effectively closing and reopening to change properties (_reconfigure_port is called for this).

pocs/utils/rs232.py
```diff
@@ -12,56 +12,66 @@
...
+        self.process = None
+        self._serial_io = None
+        self.loop_delay = 2.
 
-                self.logger.debug("Using threads (multiprocessing)")
-                self.process = Thread(target=self.receiving_function, args=(self.queue,))
-                self.process.daemon = True
-                self.process.name = "PANOPTES_{}".format(name)
+        # Properties have been set to reasonable values, ready to open the port.
+        self.ser.open()
```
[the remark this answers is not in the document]
jamessynge: Since we don't have any "open on write" behavior, this means that we have a completely useless SerialData instance if we ignore the failure. And since we use exceptions elsewhere for failures, I didn't see the point in hiding this failure. All the tests passed after making that change.

pocs/utils/rs232.py
```diff
@@ -12,56 +12,66 @@
...
 
-            self.logger.debug('Serial connection set up to {}, sleeping for two seconds'.format(self.name))
-            time.sleep(2)
-            self.logger.debug('SerialData created')
-        except Exception as err:
-            self.ser = None
-            self.logger.critical('Could not set up serial port {} {}'.format(port, err))
+        if self.is_threaded:
```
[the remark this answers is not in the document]
jamessynge: I too haven't looked at that. While probably worth looking into, I don't want to entangle these issues.

pocs/utils/rs232.py
```diff
@@ -12,56 +12,66 @@
...
+                encoding='ascii',
+                line_buffering=True)
+
+            self.logger.debug("Using threads (multiprocessing)")
+            self.process = Thread(
+                target=self.receiving_function, args=(self.queue, ))
+            self.process.daemon = True
+            self.process.name = "PANOPTES_{}".format(name)
+
+        # TODO(jamessynge): Consider eliminating this sleep period, or making
```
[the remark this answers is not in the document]
jamessynge: Not if you're comfortable with the configurable behavior.

And can you tell me what problem you were working around with the 2 second delay?

pocs/utils/rs232.py
jamessynge: Sure, probably a good idea.

pocs/utils/rs232.py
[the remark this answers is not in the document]
wtgee: Sorry, bad comment on my part. This was in reference to using `serial_for_url`. I reviewed this out of order, starting with this file, so after looking at how the tests are set up this makes more sense.

pocs/tests/test_rs232.py
jamessynge: My mistake, I misunderstood.

pocs/tests/test_rs232.py
jamessynge: No need, removing.

wtgee:
In meetings for a bit now, so quick comment: I don't think we are using the threaded anywhere so maybe look at scrapping it. More in a few.

pocs/tests/test_rs232.py
```diff
@@ -1,79 +1,171 @@
...
+        ser.start()
+        assert ser.is_listening
+        assert ser.is_threaded
+
+    # no_op handler doesn't do any reading, analogous to /dev/null, which
+    # never produces any output.
     assert '' == ser.read(retry_delay=0.01, retry_limit=2)
-    assert 0 == ser.write('')
 
+    # Assert how much is written, which unfortunately isn't consistent.
```
wtgee: I guess that's a bit concerning. Do you know why this is?

jamessynge:
Still need to review your comments.

pocs/utils/rs232.py
jamessynge: The docstring for clear_buffer says "Clear Response Buffer", which I interpret to mean the bytes coming in from the device (i.e. the response to the commands sent by POCS).

There is only a single call to clear_buffer, and it is commented out. Shall I simply remove the clear_buffer function?

wtgee: Actually it is one of the things I was going to test whether we need incorporate more, for input and output. Seems to me that the output from the arduino is lagging behind real-time, in some cases by 10+ seconds. 

In my head the serial line would just output constantly such that whenever you did a read you were getting a current reading but I guess that is not the case (hence, serial, I suppose). Experimenting it seems to work fine to clear the output buffer and then take a reading in order to get as close to "real-time" as possible.

This seems to be the case on input too, where a specified command won't take effect for a number of seconds. By first clearing the input buffer and then sending the command it seems to always be immediate. 

Clearly needs to be explored more and I'm not expert at it so let me know if I need to change how read or if I am misunderstanding something.

jamessynge: My experience is that true serial lines need to be read promptly because the internal UART (hardware) buffers are really short. We're dealing with USB to serial here, so those short UART buffers shouldn't be an issue. I don't know what options there are with Linux to get the OS to buffer potentially large amounts of input that the client hasn't read.

I think the thing to do is what I'd suggested in chat at some point: run a separate thread for each serial device "owner" (e.g. the PEAS code for reading from an arduino), so that each can read as fast as possible. For the arduinos, there isn't much need to coordinate between reading and writing, so there could be a thread for reading, and the main thread for the process can do the writing, given that writing is rare. For the mount and dome, we essentially have a synchronous protocol, where we mostly send a command and read a response.

pocs/utils/rs232.py
jamessynge: Adding comments.

jamessynge:
Superseded by a later PR

PR #148, opened 2017-12-01
Remove support for threaded IO from rs232.SerialData

jamessynge:
It wasn't in use, complicates testing, and we may be able
to use newer features in PySerial toachieve the same result.

### DIFF
--- a/pocs/utils/rs232.py
+++ b/pocs/utils/rs232.py
@@ -12,19 +12,31 @@ from .error import BadSerialConnection
 
 
 class SerialData(PanBase):
-
     """
     Main serial class
     """
 
-    def __init__(self, port=None, baudrate=115200, threaded=True, name="serial_data"):
+    def __init__(self,
+                 port=None,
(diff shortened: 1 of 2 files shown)


pocs/utils/rs232.py
```diff
@@ -12,19 +12,31 @@
...
-
     """
     Main serial class
     """
 
-    def __init__(self, port=None, baudrate=115200, threaded=True, name="serial_data"):
+    def __init__(self,
+                 port=None,
+                 baudrate=115200,
+                 threaded=None,
```
wtgee: I'm temped to say we should just add a `*args, **kwargs` and remove the `threaded=None`.

jamessynge: I'd prefer to rip out the param as soon as the PACE code is merged in here, as that way we don't accidentally accept bogus params.

jamessynge:
Merged in develop. Please review once travis is done, and merge into develop if OK.

PR #150, opened 2017-12-01
Trucker board support - WIP

### LINKED ISSUE #149: Support Infineon board (i.e. Trucker board)

wtgee:
* Pulls in `CharBuffer` and `SerialInputHandler` from panoptes/PACE#3
* Gathers data then prints so output is more seamless
* Add newline to `rs232` `write` method

Note that I am using Luc's code for the current readings and have only minimally tested but they appear to be correct.

Also, I will be adding a jupyter notebook setup that will help POCS/PEAS understand which relay is which. This would be in lieu of writing that into the arduino file itself.

Closes #149

### DIFF
--- a/pocs/utils/rs232.py
+++ b/pocs/utils/rs232.py
@@ -141,7 +141,7 @@ class SerialData(PanBase):
         if self.is_threaded:
             response = self._serial_io.write(value)
         else:
-            response = self.ser.write("{}\n".format(value).encode())
+            response = self.ser.write(value.encode())
 
         return response
 
--- a/pocs/utils/rs232.py
+++ b/pocs/utils/rs232.py
@@ -141,7 +141,7 @@ class SerialData(PanBase):
         if self.is_threaded:
             response = self._serial_io.write(value)
         else:
-            response = self.ser.write(value.encode())
+            response = self.ser.write("{}\n".format(value).encode())
 
         return response
 
--- a/resources/arduino_files/infineon_board/infineon_board.ino
+++ b/resources/arduino_files/infineon_board/infineon_board.ino
@@ -1,0 +1,362 @@
+#include <stdlib.h>
+
+#include <OneWire.h>
+#include <DallasTemperature.h>
+#include <DHT.h>
+
+#define DHTTYPE DHT22   // DHT 22  (AM2302)
+
+/* DECLARE PINS */
+
+// Digital Pins
+const int DS18_PIN = 11; // DS18B20 Temperature (OneWire)
+const int DHT_PIN = 10;  // DHT Temp & Humidity Pin
+
+// Relays
+const int RELAY_0 = A3; // 0_0 PROFET-0 Channel 0 (A3 = 17)
+const int RELAY_1 = 3;  // 1_0 PROFET-0 Channel 1
+const int RELAY_2 = 4;  // 0_1 PROFET-1 Channel 0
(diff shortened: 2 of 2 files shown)


pocs/utils/rs232.py
```diff
@@ -141,7 +141,7 @@ def write(self, value):
         if self.is_threaded:
             response = self._serial_io.write(value)
         else:
-            response = self.ser.write(value.encode())
+            response = self.ser.write("{}\n".format(value).encode())
```
jamessynge: I found it confusing that SerialData was dealing with anything except bytes. How would you feel about changing the API for write (in a future PR) so that we have writebytes and writeline? The former takes bytes (no encode call), and the latter takes a text string (ASCII only?), and must call encode (and maybe append a line terminator).

Or, perhaps replace (in time) SerialData with SerialIO, conforming to the io package's API for streams... which is really what PySerial already does.

resources/arduino_files/infineon_board/infineon_board.ino
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+  private:
+    char buf_[kBufferSize];
+    uint8_t write_cursor_;
+    uint8_t read_cursor_;
+};
+
+// Accumulates a line, parses it and takes the requested action if it is valid.
+class SerialInputHandler {
```
jamessynge: FWIW, I have a thought about making this more resilient to leading garbage: record the time when the first byte arrives for a line of input. If too much time passes before a newline is received (5 seconds?), the input is discarded.

Alternately, we switch to a mode where each line has start of record and end of record marks (byte sequences used only for that purpose), and a checksum. Then we can verify input, and report back whether we accepted the input (fields in the json data for num currently buffered bytes, num correctly processed records and num discarded bytes).

To go along with that we'd need a simple command-line tool for interacting with that protocol.

resources/arduino_files/infineon_board/infineon_board.ino
```diff
@@ -0,0 +1,362 @@
...
+#define DHTTYPE DHT22   // DHT 22  (AM2302)
+
+/* DECLARE PINS */
+
+// Digital Pins
+const int DS18_PIN = 11; // DS18B20 Temperature (OneWire)
+const int DHT_PIN = 10;  // DHT Temp & Humidity Pin
+
+// Relays
+const int RELAY_0 = A3; // 0_0 PROFET-0 Channel 0 (A3 = 17)
```
jamessynge: What does "A3 = 17" mean?

pocs/utils/rs232.py
jamessynge: Another comment on this same line: appending a \n is going to change the behavior enough that my tests will fail. Oh, and how is the mount going to respond to getting those \n values? And the dome?

I think you should drop the \n.

wtgee: Agreed, not the appropriate place for it and will be removing.

We should probably move toward the SerialIO concept as well as the start/end bytes as commented on below. I'm going to leave that to a separate PR and only make changes to the actual infineon board code here.

resources/arduino_files/infineon_board/infineon_board.ino
wtgee: I'll probably remove comment altogether. I send relay index numbers right now (0-4) but the previous version had used actual pin numbers where 17 was required instead of A3.

wtgee:
I am closing this and will re-submit against develop after other features (consolidation) are merged.

PR #159, opened 2017-12-02
100 char limit

wtgee:
This is the 100 char limit applied directly to develop.

The rebase with `1.0-beta` was getting ugly so I am essentially re-doing the "features" directly on develop now that we are all agreed to use develop.

This will be follow by the fixes to the log file which will be followed by the merging of PACE and PEAS (as separate features)


pocs/utils/theskyx.py
```diff
@@ -41,7 +41,7 @@ def connect(self):
                 self.logger.info('Connected to TheSkyX via {}:{}'.format(self._host, self._port))
 
     def write(self, value):
-        assert type(value) is str
+        assert isinstance(value, str)
```
jamessynge: Ah ha, found the one part of this reformat that wasn't a reformat! ;-)

PR #160, opened 2017-12-02
Log file updates against develop

wtgee:
This is the #145 and #152 but applied to the develop branch.

This should follow #159 and will need to be rebased.

PR #163, opened 2017-12-02
Log file updates

wtgee:
Updates to log files.

Replaces #160

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,17 +1,21 @@
 # Getting Started
-If you are unsure about a possible contribution to the project contact the project owners about your idea.
+If you are unsure about a possible contribution to the project contact the project owners about 
+your idea.
 
-Please see the [code of conduct](https://github.com/panoptes/POCS/blob/develop/CODE_OF_CONDUCT.md) for our
+Please see the [code of conduct](https://github.com/panoptes/POCS/blob/develop/CODE_OF_CONDUCT.md) 
+for our
 playground rules and follow them during all your contributions.
 
 
 # Pull Request Process
 * Pre-requisites
    - Ensure you have [github account](https://github.com/join)
-   - If the change you wish to make is not already an Issue in the project please create one specifying the need.
+   - If the change you wish to make is not already an Issue in the project please create one 
+   specifying the need.
 * Process
    - Create a fork of the repository and use a topic branch within your fork to make changes.
-      - All of our repositories have a default branch of `develop` when you first clone them, but your work should be in a separate branch.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,22 +26,80 @@ playground rules and follow them during all your contributions.
...
+ [Coding in PANOPTES](https://github.com/panoptes/POCS/wiki/Coding-in-PANOPTES) page.
 
 
 # Code Formatting
 
 - All Python should use [PEP 8 Standards](https://www.python.org/dev/peps/pep-0008/)
-   - Use a tool such as [yapf](https://github.com/google/yapf) to format your
-     files; we'd rather spend time developing PANOPTES and not arguing about
-     style.
+   - Line length is set at 120 characters instead of 80
```
jamessynge: 100, not 120, right?

CONTRIBUTING.md
```diff
@@ -22,22 +26,80 @@ playground rules and follow them during all your contributions.
...
+   - It is recommended to have your editor auto-format code whenever you save a file rather than 
+   attempt to go back and change an entire file all at once. 
 - Do not leave in commented-out code or unnecessary whitespace.
-- Variable/function/class and file names should be meaningful and descriptive
-- File names should be lower case and underscored, not contain spaces. For
-  example, `my_file.py` instead of `My File.py`
-- Define any project specific terminology or abbreviations you use in the file you use them
+- Variable/function/class and file names should be meaningful and descriptive.
+- File names should be underscored, not contain spaces ex. my_file.py.
```
jamessynge: Please keep the old file names description (lower case, etc.).

pocs/utils/logger.py
```diff
@@ -23,23 +50,41 @@ def get_root_logger(profile='panoptes', log_config=None):
...
+            finally:
+                os.symlink(full_log_fname, log_symlink)
 
     # Configure the logger
     logging.config.dictConfig(log_config)
 
     # Get the logger and set as attribute to class
     logger = logging.getLogger(profile)
 
+    # Don't want log messages from state machine library
```
jamessynge: I appreciate the comment as I had no idea what this was about. I wonder if this is the kind of thing that could be in conf_files/log.yaml. What do you think?

CONTRIBUTING.md
wtgee: Arg...this and the comment below are the same as what you saw happening with the serial testing PR where it is pulling in the old items.

wtgee:
Closing in favor of #165, which is the same thing (with corrections from above) but without the corrupted history.

PR #164, opened 2017-12-03
Add testing of rs232.SerialData

jamessynge:
This is a re-do of PR #147 after merging in the big changes from develop.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,11 +1,7 @@
 ## [Unreleased]
 
-- PR#164 (no sha yet)
-  - rs232.SerialData doesn't hide exceptions during opening of the port.
-  - Support added for testing of serial devices.
-
-- PR#148 (a436f2a127af43b6655410321f820d7660a0fe51)
-  - Remove threading support from rs232.SerialData, it wasn't used.
+- Support added for testing of serial devices.
+- Remove threading support from rs232.SerialData.
 
 ## [0.5.1] - 2017-12-02
 ### Added
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,12 @@
 ## [Unreleased]
 
+- PR#164 (no sha yet)
+  - rs232.SerialData doesn't hide exceptions during opening of the port.
+  - Support added for testing of serial devices.
+
+- PR#148 (a436f2a127af43b6655410321f820d7660a0fe51)
+  - Remove threading support from rs232.SerialData, it wasn't used.
+
 ## [0.5.1] - 2017-12-02
 ### Added
 - First real release!
--- a/pocs/utils/rs232.py
+++ b/pocs/utils/rs232.py
@@ -62,8 +62,6 @@ class SerialData(PanBase):
         # Properties have been set to reasonable values, ready to open the port.
         self.ser.open()
 
-        # TODO(jamessynge): Consider eliminating this sleep period, or making
-        # it an option in the configure file. For one thing, it slows down tests!
         open_delay = max(0.0, float(open_delay))
         self.logger.debug(
             'Serial connection set up to %r, sleeping for %r seconds', self.name, open_delay)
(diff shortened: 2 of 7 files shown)


pocs/utils/rs232.py
```diff
@@ -33,38 +36,41 @@ def __init__(self,
...
+        self.ser.timeout = 1.0
+        self.ser.xonxoff = False
+        self.ser.rtscts = False
+        self.ser.dsrdtr = False
+        self.ser.write_timeout = False
+
+        # Properties have been set to reasonable values, ready to open the port.
+        self.ser.open()
+
+        # TODO(jamessynge): Consider eliminating this sleep period, or making
```
wtgee: I still need to dig back through my notes about why the delay  is there in the first place but would prefer to capture these TODOs in an Issue unless it will be resolved as part of this PR>

pocs/utils/rs232.py
```diff
@@ -109,16 +115,20 @@ def write(self, value):
...
-        """
-        Reads value using readline
-        If no response is given, delay and then try to read again. Fail after 10 attempts
+    def read(self, retry_limit=None, retry_delay=None):
+        """Reads next line of input using readline.
+
+        If no response is given, delay for retry_delay and then try to read
+        again. Fail after retry_limit attempts.
         """
+        assert not self._start_needed
```
wtgee: I don't see where this comes from or get what it is doing.

pocs/utils/rs232.py
```diff
@@ -148,13 +158,27 @@ def get_reading(self):
...
-
+        # Not worrying here about bytes arriving between reading in_waiting
+        # and calling reset_input_buffer().
+        # Note that Wilfred reports that the Arduino input can seriously lag behind
+        # realtime (e.g. 10 seconds), and that clear_buffer may exist for that reason
+        # (i.e. toss out any buffered input, and then read the next full line...
+        # which likely requires tossing out a fragment of a line).
+        count = self.ser.in_waiting
+        self.ser.reset_input_buffer()
         # self.logger.debug('Cleared {} bytes from buffer'.format(count))
```
wtgee: IIRC this was generating too many log entries and so was commented out. Could consider eliminating along with `count` variable above.

pocs/utils/rs232.py
```diff
@@ -148,13 +158,27 @@ def get_reading(self):
...
     def clear_buffer(self):
         """ Clear Response Buffer """
-        count = 0
-        while self.ser.inWaiting() > 0:
-            count += 1
-            self.ser.read(1)
-
+        # Not worrying here about bytes arriving between reading in_waiting
+        # and calling reset_input_buffer().
+        # Note that Wilfred reports that the Arduino input can seriously lag behind
```
wtgee: The lag of 10 seconds would probably apply more to the output buffer. Once it is lagging the only way to catch-up would be to clear output buffer.

pocs/utils/rs232.py
jamessynge: Ah. Sorry, I'm in the habit of capturing these things in the code where their context is readily apparent.

pocs/utils/rs232.py
jamessynge: Removed.

pocs/utils/rs232.py
jamessynge: output meaning writing from POCS to the Arduino? That sounds like a flush call needs to be made on the device to ensure it is making it out of buffers and into the OS.

pocs/utils/rs232.py
jamessynge: I can do so now if you like.

pocs/utils/rs232.py
wtgee: Output meaning from the Arduino to POCS. Unless we connect to and start reading from the arduino immediately after it is reset the sensor readings lag behind.

pocs/utils/rs232.py
wtgee: Yeah, let's get rid of it.

wtgee:
With a big addition to this I would like to get into the habit of adding a quick user-friendly entry to the Changelog.md file. This should be added to the `Unreleased` section. Can you add something for the PR? Thanks!

pocs/utils/rs232.py
jamessynge: Let's try to be specific when we use the words input and output (I wasn't sufficiently in the comment). So, "input from Arduino" rather than "Arduino input", which was ambiguous; or "POCS output to the mount".

jamessynge: I've updated the comment accordingly.

pocs/utils/rs232.py
jamessynge: Issue added, TODO removed.

pocs/utils/rs232.py
jamessynge: Done.

jamessynge:
Added Changelog.md. Will test and push.

jamessynge:
PTAL

Changelog.md
```diff
@@ -1,5 +1,12 @@
 ## [Unreleased]
 
+- PR#164 (no sha yet)
```
wtgee: I guess we should first clarify what the purpose of the changelog is. I'm mostly following http://keepachangelog.com/en/0.3.0/

I don't think we care about the PR or the SHA. Things like "...doesn't hide exceptions..." also seem like too much detail for how I think about this.

jamessynge: Fixed.

PR #165, opened 2017-12-03
Log files updates

wtgee:
* Add docs
* Log file per level
* Log files name consistency
* Clean up a few instances where it just had `warn` instead of `warning`

Replaces #163 which replaces #160. The difference with #163 here is just that it has a cleaner history without extraneous commits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,17 +1,21 @@
 # Getting Started
-If you are unsure about a possible contribution to the project contact the project owners about your idea.
+If you are unsure about a possible contribution to the project contact the project owners about 
+your idea.
 
-Please see the [code of conduct](https://github.com/panoptes/POCS/blob/develop/CODE_OF_CONDUCT.md) for our
+Please see the [code of conduct](https://github.com/panoptes/POCS/blob/develop/CODE_OF_CONDUCT.md) 
+for our
 playground rules and follow them during all your contributions.
 
 
 # Pull Request Process
 * Pre-requisites
    - Ensure you have [github account](https://github.com/join)
-   - If the change you wish to make is not already an Issue in the project please create one specifying the need.
+   - If the change you wish to make is not already an Issue in the project please create one 
+   specifying the need.
 * Process
    - Create a fork of the repository and use a topic branch within your fork to make changes.
-      - All of our repositories have a default branch of `develop` when you first clone them, but your work should be in a separate branch.
+      - All of our repositories have a default branch of `develop` when you first clone them, but 
+      your work should be in a separate branch.
       - Create a branch with a descriptive name, e.g.:
          - `git checkout -b new-camera-simulator`
          - `git checkout -b issue-28`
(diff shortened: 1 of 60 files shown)


CONTRIBUTING.md
```diff
@@ -22,22 +26,84 @@ playground rules and follow them during all your contributions.
...
+- Log level:
+   - DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
+   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a
+    person actively watching a running unit.
+   - WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but
+    operation of unit can continue.
+   - ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot 
+   continue.
+- The logger supports variable information without the use of the `format` method.
+- There is a `say` method that is meant to be used in friendly manner to convey information to a
```
jamessynge: There is a `say` method available on sub-classes of PanBase that is meant...

pocs/core.py
```diff
@@ -43,6 +43,11 @@ def __init__(self, state_machine_file='simple_state_table', messaging=False, **k
         PanBase.__init__(self, **kwargs)
 
         self.logger.info('Initializing PANOPTES unit')
```
jamessynge: Remove old, non-parameterized message.

pocs/images.py
```diff
@@ -115,7 +115,8 @@ def pointing_error(self):
             namedtuple: Pointing error information
         """
         if self._pointing_error is None:
-            assert self.pointing is not None, self.logger.warn("No WCS, can't get pointing_error")
+            assert self.pointing is not None, self.logger.warning(
+                "No WCS, can't get pointing_error")
```
jamessynge: WCS?

pocs/utils/logger.py
```diff
@@ -23,23 +50,41 @@ def get_root_logger(profile='panoptes', log_config=None):
...
+            finally:
+                os.symlink(full_log_fname, log_symlink)
 
     # Configure the logger
     logging.config.dictConfig(log_config)
 
     # Get the logger and set as attribute to class
     logger = logging.getLogger(profile)
 
+    # Don't want log messages from state machine library
```
jamessynge: I appreciate the comment as I had no idea what this was about. I wonder if this is the kind of thing that could be in conf_files/log.yaml. What do you think?

pocs/utils/logger.py
```diff
@@ -1,16 +1,43 @@
...
 
 
+class PanLogger(object):
+    """ Logger for PANOPTES with format style strings """
+
+    def __init__(self, logger):
+        super(PanLogger, self).__init__()
+        self.logger = logger
+
+    def _process_str(self, fmt, *args, **kwargs):
```
jamessynge: I don't understand why this method is necessary. According to the [python documentation](https://docs.python.org/3/library/logging.html#logging.Logger.debug), the loggers do exactly this formatting.

CONTRIBUTING.md
```diff
@@ -1,17 +1,21 @@
 # Getting Started
-If you are unsure about a possible contribution to the project contact the project owners about your idea.
+If you are unsure about a possible contribution to the project contact the project owners about 
+your idea.
 
-Please see the [code of conduct](https://github.com/panoptes/POCS/blob/develop/CODE_OF_CONDUCT.md) for our
+Please see the [code of conduct](https://github.com/panoptes/POCS/blob/develop/CODE_OF_CONDUCT.md)
```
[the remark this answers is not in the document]
wtgee: Unfortunately those kind of links don't work in github. The `blob/develop` is the only way to link to an active item in to the repo AFAIK.

CONTRIBUTING.md
```diff
@@ -22,22 +26,84 @@ playground rules and follow them during all your contributions.
...
-- Variable/function/class and file names should be meaningful and descriptive
-- File names should be lower case and underscored, not contain spaces. For
-  example, `my_file.py` instead of `My File.py`
-- Define any project specific terminology or abbreviations you use in the file you use them
+- Variable/function/class and file names should be meaningful and descriptive.
+- File names should be lower case and underscored, not contain spaces. For example, `my_file.py` 
+instead of `My File.py`.
+- Define any project specific terminology or abbreviations you use in the file you use them.
+- Variable/function/class and file names should be meaningful and descriptive.
+- File names should be underscored, not contain spaces ex. my_file.py.
```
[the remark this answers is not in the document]
wtgee: Wow, I think I have merged this incorrectly every single time. ;)

CONTRIBUTING.md
```diff
@@ -22,22 +26,84 @@ playground rules and follow them during all your contributions.
...
+   - DEBUG (i.e. `self.logger.debug()`) should attempt to capture all run-time information.
+   - INFO (i.e. `self.logger.info()`) should be used sparingly and meant to convey information to a
+    person actively watching a running unit.
+   - WARNING (i.e. `self.logger.warning()`) should alert when something does not go as expected but
+    operation of unit can continue.
+   - ERROR (i.e. `self.logger.error()`) should be used at critical levels when operation cannot 
+   continue.
+- The logger supports variable information without the use of the `format` method.
+- There is a `say` method that is meant to be used in friendly manner to convey information to a 
+user. This should be used only for personable output and is typically displayed in the "chat box"
```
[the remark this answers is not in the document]
wtgee: Sort of yes. I made a change to description above. The `say` method is only available to the `POCS` class (not in `PanBase` and therefore not in all classes) and should be very friendly output. As in something that you could feasibly send to twitter if you wanted your unit to do such a thing.

The only place so far that there is really operator output was for the drift alignment stuff and that was simply to support Nem the one time he was doing that. During normal operations the `say` output doesn't include anything technical.

The method is also a place that could easily be internationalized, which I have an Issue for, which would mostly be for fun. But that way someone can make their unit "speak" in whatever language they want.

pocs/utils/logger.py
wtgee: I don't think we ever want to show these. There is a lot of output that is generated (something like: prepare, before enter, enter, after enter, before exit, exit, etc) and we already have our own logging for entering/exiting each state.

pocs/utils/logger.py
wtgee: This supports the actual format style, e.g. `{:0.2f}`, `{:40s}`, rather than just supporting `{}` as placeholders. The loggers support this via the `%f` syntax, which is simply for backwards-compatibility.  This allows us to just get rid of actual `format` calls in the logger but still keep the same syntax.

I could definitely be wrong about this but the format specifiers didn't seem to work for me otherwise.

CONTRIBUTING.md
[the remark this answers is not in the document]
jamessynge: Got it.

pocs/utils/logger.py
jamessynge: OK. Perhaps add a comment explaining that?

pocs/utils/logger.py
jamessynge: Got it. In that case, please say so succinctly.

# Don't want log messages from state machine library, it is very noisy
# and we have our ow logging of state transitions.

pocs/utils/logger.py
jamessynge: Update now that I'm trying to use the logger: it is confusing to have PanLogger use a different formatting convention from python's loggers, even though the API is otherwise identical.

wtgee: Not sure I follow. You should be able to use either convention.

PR #166, opened 2017-12-03
Version number updates

### LINKED ISSUE #154: Change hard-coded version number in test_observatory
https://github.com/panoptes/POCS/blob/1.0-beta/pocs/tests/test_observatory.py#L171

wtgee:
* Change terminology to match semantic versioning
* Don't hard-code version number in test (closes #154)

### DIFF
--- a/pocs/version.py
+++ b/pocs/version.py
@@ -1,4 +1,3 @@
-# this file was automatically generated
 major = 0
 minor = 5
 patch = 1
--- a/pocs/version.py
+++ b/pocs/version.py
@@ -1,6 +1,6 @@
 # this file was automatically generated
 major = 0
 minor = 5
-release = 1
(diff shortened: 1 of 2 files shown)


pocs/version.py
```diff
@@ -1,6 +1,6 @@
 # this file was automatically generated
```
jamessynge: Is this claim true? If so, how is it generated?

jamessynge: Having now read [Creating a New Release](https://github.com/panoptes/POCS/wiki/Creating-a-New-Release), I can see that this file manually updated, so the comment should be removed.

wtgee: Amazing how I can miss that in a file that has 5 lines total. Thanks.

PR #167, opened 2017-12-03
Merge PACE

wtgee:
Merges the PACE history into POCS.

Arduino files are placed in `$POCS/resources/arduino_files`

**Note: Should not be squashed.**


Changelog.md
```diff
@@ -1,4 +1,6 @@
 ## [Unreleased]
+### Added
```
jamessynge: FYI, I'm a bit skeptical about using a file like this for tracking major changes that should be included release notes. One issue is that we'll have frequent conflicts on this file as there will be many changes at the same point in the file.

I wonder if using the github PR "labels" feature would be useful here (e.g. mark new features with FEATURE and mark notable bug fixes with BUGFIX). Thoughts?

wtgee: Essentially these are the release notes.

I couldn't quite figure out what to do about the frequent conflicts either. It's an issue when there are multiple outstanding Feature PRs that you want to merge at once.

The idea is to alleviate the work on whomever is doing the release so they don't have to dig back through everything that has been changed since last release.

Definitely open to other suggestions.

wtgee:
There are files removed but nothing is changed.

PR #171, opened 2017-12-03
Testing Updates

wtgee:
* Update README to include details about testing
* Cleanup up how hardware tests are marked

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,8 @@ to test the code related to the cameras one can run:
 (panoptes-env) $ pytest -xv pocs/tests/test_camera.py
 ```
 
-Here the `-x` option will stop the tests upon the first failure and the `-v` makes the testing verbose.
+Here the `-x` option will stop the tests upon the first failure and the `-v` makes 
+the testing verbose.
 
 Any new code should also include proper tests. See below for details.
 
@@ -202,8 +203,29 @@ For more details see the [Writing Tests](https://github.com/panoptes/POCS/wiki/W
 
 ### Hardware Testing
 
+Hardware testing uses the same testing suite as the software testing but with 
+additional options passed on the command line to signify what hardware should be
+tested.
+
+The options to pass to `pytest` is `--with-hardware`, which accepts a list of
+possible hardware items that are connected. This list includes `camera`, `mount`, 
+and `weather`. Optionally you can use `all` to test a fully connected unit.
+
 > :warning: The hardware tests do not perform safety checking of the weather or
-> dark sky. It is assumed that hardware testing is always done with direct supervision.
+> dark sky. The `weather` test mentioned above tests if a weather station is 
+> connected but does not test the safety conditions. It is assumed that hardware
+> testing is always done with direct supervision.
+
+```bash
+# Test an attached camera
+pytest --with-hardware=camera
+
+# Test an attached camera and mount
+pytest --with-hardware=camera,mount
+
+# Test a fully connected unit
+pytest --with-hardware=all
+```
 
 **In Progress**
 
--- a/README.md
+++ b/README.md
@@ -172,8 +172,8 @@ Here you can see that certain tests were skipped (`s`) for various reasons while
 the others passed. Skipped tests are skipped on purpose and thus are not considered 
 failures. Usually tests are skipped because there is no attached hardware 
 (see below for running tests with hardware attached). All passing tests are represented
-by a single period (`.`) and any errors would show as a `e`. If there are any errors
-while running the tests the output from those errors will be displayed.
+by a single period (`.`) and any failures would show as a `F`. If there are any failures
+while running the tests the output from those failures will be displayed.
 
 #### Testing your code changes
 
--- a/README.md
+++ b/README.md
@@ -80,15 +80,98 @@ See below for more details.
 
 ## Test POCS
 
-Once you have setup your computer (above), the next step is to test your setup.
-This is easy to do using our built-in test suite. In a terminal, simply type:
+POCS comes with a testing suite that allows it to test that all of the software works and is installed corretly. Running the test suite by default will use simulators for all of the hardware and is meant to test that the software works correctly. Additionally, the testing suite can be run with various flags to test that attached hardware is working properly.
+
+All of the test files live in `$POCS/pocs/tests`.
+
+### Software Testing
+
+There are two scenarios where you want to run the test suite:
+
+1. You are getting your unit ready and want to test software is installed correctly
+2. You are helping develop code for POCS and want test your code doesn't break something
+
+> :bulb: NOTE: The test suite can take a while to run and often appears to be stalled. Check the log files to ensure activity is happening. The tests can be cancelled by pressing `Ctrl-c` (sometimes entering this command multiple times is required).
+
+It is often helpful to view the log output in another terminal window while the test suite is running:
 
 ```bash
-cd ${POCS}
-pytest
+# Follow the log file
+> tail -f $PANDIR/logs/panoptes.log
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -80,15 +80,98 @@ See below for more details.
 
 ## Test POCS
 
-Once you have setup your computer (above), the next step is to test your setup.
-This is easy to do using our built-in test suite. In a terminal, simply type:
+POCS comes with a testing suite that allows it to test that all of the software works and is installed corretly. Running the test suite by default will use simulators for all of the hardware and is meant to test that the software works correctly. Additionally, the testing suite can be run with various flags to test that attached hardware is working properly.
```
jamessynge: c/corretly/correctly/.

Please add line breaks.

Please mention that testing with hardware enabled should be done with caution, as there will be no safety checks (i.e. unparking of a mount while it is raining).

README.md
```diff
@@ -80,15 +80,98 @@ See below for more details.
...
+
+All of the test files live in `$POCS/pocs/tests`.
+
+### Software Testing
+
+There are two scenarios where you want to run the test suite:
+
+1. You are getting your unit ready and want to test software is installed correctly
```
jamessynge: ... including after upgrading to a new release of software (POCS, its dependencies or the operating system).

README.md
```diff
@@ -80,15 +80,98 @@ See below for more details.
...
+2. You are helping develop code for POCS and want test your code doesn't break something
```
jamessynge: Please end sentences with periods (generally, please use complete sentences and normal punctuation).

README.md
```diff
@@ -80,15 +80,98 @@ See below for more details.
...
+
+> :bulb: NOTE: The test suite can take a while to run and often appears to be stalled. Check the log files to ensure activity is happening. The tests can be cancelled by pressing `Ctrl-c` (sometimes entering this command multiple times is required).
```
jamessynge: Perhaps move this note and the tail command below the first example of how to run pytest.

README.md
```diff
@@ -80,15 +80,98 @@ See below for more details.
...
+pocs/tests/bisque/test_dome.py ssss
+pocs/tests/bisque/test_mount.py sssssssssss
+pocs/tests/bisque/test_run.py s
+
+=========================== 203 passed, 57 skipped, 6 warnings in 435.76 seconds ===================================
+
 ```
 
-This may take 5 to 10 minutes as there are a lot of tests to run! If you experience any errors, ask for check the [Issues](https://github.com/panoptes/POCS/issues) listed above or ask one of our friendly team members!
+Here you can see that certain tests were skipped (`s`) for various reasons while the others passed. Skipped tests are skipped on purpose and thus are not considered failures. Usually tests are skipped because there is no attached hardware (see below for running tests with hardware attached).
```
jamessynge: Perhaps some more detail on how to read the output: a . for each passed test, an f for each failed test, etc?

pocs/tests/conftest.py
```diff
@@ -2,39 +2,70 @@
...
-        help="If a real weather station attached")
+    parser.addoption("--hardware-test", action="store_true", default=False,
+                     help="Test with hardware attached")
+    parser.addoption("--camera-test", action="store_true", default=False,
+                     help="If a real camera attached")
+    parser.addoption("--mount-test", action="store_true", default=False,
+                     help="If a real mount attached")
+    parser.addoption("--weather-test", action="store_true", default=False,
+                     help="If a real weather station attached")
     parser.addoption("--solve", action="store_true", default=False,
```
jamessynge: Why is --solve disabled by default?

And should it be renamed --solve-test or --disable-solver (the latter because it sounds like it is not a test of the solver, but a focus on reducing test duration).

pocs/tests/conftest.py
```diff
@@ -2,39 +2,70 @@
...
-        action="store_true",
-        default=False,
-        help="If a real camera attached")
-    parser.addoption("--mount", action="store_true", default=False, help="If a real mount attached")
-    parser.addoption(
-        "--weather",
-        action="store_true",
-        default=False,
-        help="If a real weather station attached")
+    parser.addoption("--hardware-test", action="store_true", default=False,
```
jamessynge: "Test with all hardware attached"

But, this strikes me as problematic for multiple reasons: few will have domes, many will have some hardware available but not all, and testing with the cameras doesn't risk exposing them to water, but testing with the mount does.

Perhaps a variant would be the opposite of the simulator list: --with-hardware=camera,weather

README.md
wtgee: Added line breaks here but don't necessarily want to change whole document in this PR.

Added warning about hardware in hardware section below.

pocs/tests/conftest.py
wtgee: Good point plus that gives the same sort of calling structure as the simulator. I'll get this added in to this PR, which will require some more work so might not get to today.

pocs/tests/conftest.py
wtgee: It was disabled by default because of the problems getting astrometry.net installed on travis-ci. I think that is resolved now however so we should look at changing this behavior as part of another PR. Created #175

pocs/tests/conftest.py
wtgee: I've got this change in and will be pushing soon, but another question might be whether or not we should skip all non-hardware related tests when testing hardware. If I am interested in testing the camera I don't necessarily want to wait for the `test_pocs.py` software tests to run.

Probably can just leave it for now but might require some thinking about.

wtgee:
I changed this so there is only one hardware option that accepts a list of items. @jamessynge please take a look when you get a chance. Thanks!

README.md
```diff
@@ -147,30 +168,65 @@ pocs/tests/bisque/test_run.py s
 
 ```
 
-Here you can see that certain tests were skipped (`s`) for various reasons while the others passed. Skipped tests are skipped on purpose and thus are not considered failures. Usually tests are skipped because there is no attached hardware (see below for running tests with hardware attached).
+Here you can see that certain tests were skipped (`s`) for various reasons while 
+the others passed. Skipped tests are skipped on purpose and thus are not considered 
+failures. Usually tests are skipped because there is no attached hardware 
+(see below for running tests with hardware attached). All passing tests are represented
+by a single period (`.`) and any errors would show as a `e`. If there are any errors
```
jamessynge: Don't test failures show as "`f`"?

pocs/tests/conftest.py
```diff
@@ -24,48 +19,39 @@ def pytest_collection_modifyitems(config, items):
...
     `@pytest.mark.with_camera`: Run tests with camera attached
     `@pytest.mark.with_mount`: Run tests with camera attached
     `@pytest.mark.with_weather`: Run tests with camera attached
 
     Note:
-        The `--hardware-test` supersedes the other options, only use to test a
-        fully attached unit.
-
+        We are marking which tests to skip rather than which tests to include
+        so the logic is opposite of the options
```
jamessynge: Append period.

pocs/tests/conftest.py
```diff
@@ -24,48 +19,39 @@ def pytest_collection_modifyitems(config, items):
     Certain tests should only be run when the appropriate hardware is attached.
     These tests should be marked as follows:
 
-    `@pytest.mark.hardware`: Run all hardware tests
     `@pytest.mark.with_camera`: Run tests with camera attached
```
jamessynge: Append period to each.

wtgee: The periods annoy you so much you didn't notice they all say "camera" :)

jamessynge:
Apparently so! :-D

PR #176, opened 2017-12-06
Build improvements

wtgee:
* Cleaning up PanBase for easier setup and build purposes
        * Removing `distutils` fallback (probably not necessary but makes it a little cleaner - see [footnote 5](https://packaging.python.org/guides/tool-recommendations/#id13))
	* Moving `PanBase` to its own file
	* Fix tests
* Proper version (i.e. `pocs.__version` works)

This is a step toward easily having an easy anaconda package.


pocs/__init__.py
```diff
@@ -6,127 +6,11 @@
...
-        if 'logger' in d:
-            del d['logger']
-
-        if 'db' in d:
-            del d['db']
-
-        return d
-
-
+from .version import __version__
```
jamessynge: Please use root relative paths.

pocs/core.py
```diff
@@ -1,7 +1,10 @@
+import os
+import sys
 import queue
 import time
 import zmq
 
+from warnings import warn
```
jamessynge: Please move towards importing packages and modules, not the members of modules.

pocs/tests/test_observatory.py
```diff
@@ -8,7 +8,7 @@
 from pocs.scheduler.dispatch import Scheduler
 from pocs.scheduler.observation import Observation
 from pocs.utils import error
-from pocs.version import version
+from pocs.version import __version__ as version
```
jamessynge: Ditto.

setup.cfg
```diff
@@ -16,7 +16,7 @@ author_email = info@projectpanoptes.org
 description = Finding exoplanets with small digital cameras
 edit_on_github = True
 github_project = panoptes/POCS
-keywords = Citizen-science open-source exoplanet digital dslr camera
+keywords = Citizen-science open-source exoplanet digital DSLR camera
```
jamessynge: + astronomy & STEM.

pocs/core.py
wtgee: Hm, I must have understood the conversation backwards. The `from` syntax should be the preferred. There is no pep8 specifically about this but the recommendation is to import the minimal needed from a package as opposed to the entre package or module.

pocs/__init__.py
wtgee: This might be an odd situation where that doesn't work as it is needed for setup.py, which means it's needed before module is installed. I'll double check and change if not

wtgee:
@jamessynge Can you look at this again.

pocs/base.py
```diff
@@ -1,20 +1,19 @@
...
-_logger = None
 
 
 class PanBase(object):
 
-    """ Base class for other classes within the Pan ecosystem
+    """ Base class for other classes within the PANOPTES ecosystem
 
-    Defines common properties for each class (e.g. logger, config)self.
+    Defines common properties for each class (e.g. logger, config).
```
jamessynge: We should soon elaborate on those common properties... Or maybe put them in another object shared by PanBase instances.

pocs/core.py
```diff
@@ -427,7 +425,7 @@ def _check_environment(self):
             POCS      Base directory for POCS
         """
         if sys.version_info[:2] < (3, 0):  # pragma: no cover
-            warn("POCS requires Python 3.x to run")
+            warnings.warn("POCS requires Python 3.x to run")
```
jamessynge: Why warnings instead of self.logger?

pocs/core.py
jamessynge: I'm not referencing pep8, but instead the Google Python Style Guide, which recommends importing packages or modules, but not the members of modules (e.g. not the functions and classes in a module). This is intended to make it easier for a reader to tell where a symbol comes from (i.e. if I see foo.Bar(), I know that the symbol Bar is defined in module foo, assuming that is a constructor call).

pocs/core.py
wtgee: It checks the environment before it adds the logger.

PR #179, opened 2017-12-09
Adding a .style.yapf file, to guide yapf (Yet-Another-Python-Formatter).

jamessynge:
The only changes from the baseline settings are the placement of binary
operators when wrapped (before the break, not after) and the lenght of
lines (100, not 79).


wtgee:
Could we maybe add something in the README or Coding wiki about how to use this?

PR #180, opened 2017-12-09
Update, document and test SerialData

jamessynge:
Make the API of rs232.SerialData a bit more regular, and document
it much more thoroughly. Updated the calls accordingly.
Increased test coverage of SerialData.

### DIFF
--- a/pocs/mount/serial.py
+++ b/pocs/mount/serial.py
@@ -24,15 +24,7 @@ class AbstractSerialMount(AbstractMount):
                 'No mount port specified, cannot create mount\n {}'.format(
                     self.config['mount']))
 
-        try:
-            self.serial = rs232.SerialData(port=self._port, baudrate=9600)
-        except Exception as err:
-            self.serial = None
-            # This won't be triggered because SerialData is created even if the mount port
-            # can't be opened. We can't tell the difference between not currently present
-            # and totally invalid (e.g. com1: on a Linux box, or /dev/tty7 on Windows),
-            # TODO(wtgee): What would you like to do about this?
-            raise error.MountNotFound(err)
+        self.serial = rs232.SerialData(port=self._port, baudrate=9600)
 
 
 ##################################################################################################
--- a/pocs/mount/serial.py
+++ b/pocs/mount/serial.py
@@ -10,7 +10,9 @@ from .mount import AbstractMount
 class AbstractSerialMount(AbstractMount):
 
     def __init__(self, *args, **kwargs):
-        """
+        """Initialize an AbstractSerialMount for the port defined in the config.
+
+        Opens a connection to the serial device, if it is valid.
         """
         super(AbstractSerialMount, self).__init__(*args, **kwargs)
 
@@ -26,6 +28,10 @@ class AbstractSerialMount(AbstractMount):
             self.serial = rs232.SerialData(port=self._port, baudrate=9600)
         except Exception as err:
             self.serial = None
+            # This won't be triggered because SerialData is created even if the mount port
+            # can't be opened. We can't tell the difference between not currently present
+            # and totally invalid (e.g. com1: on a Linux box, or /dev/tty7 on Windows),
+            # TODO(wtgee): What would you like to do about this?
             raise error.MountNotFound(err)
 
 
@@ -34,14 +40,15 @@ class AbstractSerialMount(AbstractMount):
 ##################################################################################################
 
     def connect(self):
-        """ Connects to the mount via the serial port (`self._port`)
+        """Connects to the mount via the serial port (`self._port`)
 
         Returns:
-            bool:   Returns the self.is_connected property which checks the actual serial connection.
+            Returns the self.is_connected property (bool) which checks
+            the actual serial connection.
         """
         self.logger.debug('Connecting to mount')
 
-        if self.serial.ser and self.serial.ser.isOpen() is False:
+        if self.serial and not self.serial.is_connected:
             try:
                 self._connect()
             except OSError as err:
@@ -57,7 +64,9 @@ class AbstractSerialMount(AbstractMount):
 
     def disconnect(self):
         self.logger.debug("Closing serial port for mount")
-        self._is_connected = self.serial.disconnect()
+        if self.serial:
+            self.serial.disconnect()
+        self._is_connected = self.serial.is_connected
 
     def set_tracking_rate(self, direction='ra', delta=0.0):
         """Set the tracking rate for the mount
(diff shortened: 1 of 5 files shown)


jamessynge:
I'm perplexed about the lack of coverage of pocs/utils/rs232.py. Running coverage.py locally, I see coverage of it, but codecov doesn't show it here. The .coveragerc file does not mark it to be omitted.

pocs/mount/serial.py
```diff
@@ -26,6 +28,10 @@ def __init__(self, *args, **kwargs):
             self.serial = rs232.SerialData(port=self._port, baudrate=9600)
         except Exception as err:
             self.serial = None
+            # This won't be triggered because SerialData is created even if the mount port
+            # can't be opened. We can't tell the difference between not currently present
+            # and totally invalid (e.g. com1: on a Linux box, or /dev/tty7 on Windows),
+            # TODO(wtgee): What would you like to do about this?
```
wtgee: This is why I was a little skeptical about creating the `SerialData` object regardless. The `error.BadSerialConnection` is a good replacement, but it is a big annoying when in use via `pocs_shell`, because the error right here will trigger before anything else is set up and tell the user to use a simulator. 

However, I think that could also be handled in a better manner. I think we could simple remove the `try/except` here.

pocs/tests/test_rs232.py
```diff
@@ -32,9 +50,16 @@ def test_detect_bogus_scheme(handler):
         rs232.SerialData(port='bogus://')
 
 
+def test_detect_bogus_scheme(handler):
+    """When our handlers are installed, will still detect unknown scheme."""
```
wtgee: What are valid schemes? Have we documented these handlers and schemes as they apply to us?

pocs/tests/test_rs232.py
```diff
@@ -43,34 +68,51 @@ def test_basic_no_op(handler):
...
-        ser.read(retry_delay=0.01, retry_limit=1)
-    with pytest.raises(AssertionError):
-        ser.write('a')
+    # Several passes of reading, writing, disconnecting and connecting.
+    for _ in range(3):
+        # no_op handler doesn't do any reading, analogous to /dev/null, which
+        # never produces any output.
+        assert '' == ser.read(retry_delay=0.01, retry_limit=2)
+        assert b'' == ser.read_bytes(size=1)
+        assert 0 == ser.write('abcdef')
```
wtgee: So does `write` ever return a valid number or is it always 0?

pocs/utils/rs232.py
```diff
@@ -12,37 +14,46 @@
...
 class SerialData(PanBase):
-    """
-    Main serial class
+    """SerialData wraps a PySerial instance for reading from and writing to a serial device.
+
+    Because POCS is intended to be very long running, and hardware may be turned off when unused
+    or to force a reset, this wrapper may or may not have an open connection to the underlying
+    serial device. Note that for most devices, is_connected will return true if the device is
+    turned off/unplugged after a connection is opened; the code will only discover there is a
+    problem when we attempt to interact with the device.
```
wtgee: If this is the case we should have a method on `pocs` or `observatory` that tests the connections and can be checked from the [ready](https://github.com/panoptes/POCS/blob/develop/pocs/state/states/default/ready.py) state. We can wait to do this when we write the methods that actually do the shutting down (probably in the [housekeeping](https://github.com/panoptes/POCS/blob/develop/pocs/state/states/default/housekeeping.py) state.)

pocs/utils/rs232.py
```diff
@@ -59,61 +70,88 @@ def __init__(self,
...
-        self.logger.debug('Serial connection established to {}'.format(
-            self.name))
-        return self.ser.isOpen()
+        """If disconnected, then connect to the serial port.
+
+        Raises:
+            BadSerialConnection if unable to open the connection.
+        """
+        if self.is_connected:
+            return
```
wtgee: Does it generate too much noise if we have a log message here saying connection was already open?

pocs/utils/rs232.py
```diff
@@ -135,28 +173,28 @@ def read(self, retry_limit=None, retry_delay=None):
...
         """
-
-        # TODO(jamessynge): Consider collecting the time (utc?) after the read completes,
-        # so that long delays reading don't appear to have happened much earlier.
         try:
+            # Get the timestamp after the read so that a long delay on reading doesn't make it
+            # appear that the read happened much earlier than it did.
+            line = self.read()
             ts = time.strftime('%Y-%m-%dT%H:%M:%S %Z', time.gmtime())
             info = (ts, self.read())
```
wtgee: Should this be `line` instead of another `self.read()`?

pocs/utils/rs232.py
```diff
@@ -135,28 +173,28 @@ def read(self, retry_limit=None, retry_delay=None):
...
         except IndexError:
+            # TODO(wtgee): When might this occur?
```
wtgee: Good question but I think it was when dealing the `threaded` and reading from a `queue`. Not entirely sure though so perhaps we should remove until becomes a problem. Or at least add a log message such that we can easily find when it does happen.

pocs/mount/serial.py
jamessynge: I've removed that (and will send another PR).

I'll create an issue to discuss the issue you raise.

pocs/tests/test_rs232.py
jamessynge: Rather than document all of those (mostly provided by PySerial), I documented what the scheme must be, and then violated that so that the exception will be raised.

pocs/tests/test_rs232.py
jamessynge: The no_op handler never returns anything, but other handles do report the size written.

pocs/utils/rs232.py
jamessynge: Submitted #185 to track this issue.

pocs/utils/rs232.py
jamessynge: Added appropriate message.

pocs/utils/rs232.py
jamessynge: Doh! Test added that detected this bug, and bug fixed.

pocs/utils/rs232.py
jamessynge: Removed try/except.

pocs/tests/test_rs232.py
```diff
@@ -43,34 +69,56 @@ def test_basic_no_op(handler):
...
+        # And we should be able to reconnect.
+        assert not ser.is_connected
+        ser.connect()
+        assert ser.is_connected
 
 
 def test_basic_io(handler):
-    protocol_buffers.ResetBuffers(b'abc\r\n')
-    ser = rs232.SerialData(port='buffers://', open_delay=0, threaded=False)
+    protocol_buffers.ResetBuffers(b'abc\r\ndef\n')
```
wtgee: So does this consume either `\r\n` or `\n`?

PR #182, opened 2017-12-09
Enable coverage of pocs/utils/rs232.py

jamessynge:


jamessynge:
The test serial handlers help, but only where used. We'll have to add tests for the iOptron mount and the peas sensor code.

wtgee:
> The test serial handlers help, but only where used. We'll have to add handlers for simulating the iOptron mount and the Arduino sketches.

Yes, I was more just wondering if you feel this gives us an appropriate mechanism to start creating valid tests. Seems like it. We can move discussion to separate Issue however.

Issue #134, opened 2017-11-26
Add support for Astrohaven dome


jamessynge:
Uses a Unitronics Vision 130 PLC to control opening and closing.
Has a very simple RS-232 protocol according to this extract the installation manual:

2. Using a serial communication program like Microsoft Windows HyperTerminal
set for 9600 baud, eight bit, no parity, and hardware control = off; type “A” or “a”
to make the side opposite the controller box go up and down. Similarly type “B”
or “b” to make the side of the controller box go up and down. You will note that
the controller replies with your command until the limit switches are encountered.
For the side opposite the controller encountering the upper limit switch will result
in replies of “X” while the encountering the lower limit switch will result in replies
of “x”. You will not similar replies for the side with the controller box except they
will be “Y” and “y”.
3. Test the full Close and full Open commands using “C” and “O” as inputs. Test
also that the “R” command will reset the BG Controller after the BG (Bumper Guard) Switch has
been bumped.

jamessynge:
Tested and debugged code with Astrohaven dome at Wheaton. My only remaining concern is about shutters not quite closing all the way (a little light coming through at the top). This may be a matter of timing of commands, momentum of the shutters, position of the "closed" sensors.

PR #187, opened 2017-12-12
Implement support for Astrohaven dome

jamessynge:
Addresses #134, though there is not yet any integration of domes into the state machine.


jamessynge:
Sorry, meant to squash first.

PR #188, opened 2017-12-12
Add support for Astrohaven dome

jamessynge:
Includes a simulator for the PLC (programmable logic controller) that controls the dome's motors.

Addresses #134, though there is not yet any integration of domes into the state machine.


pocs/dome/abstract_serial_dome.py
```diff
@@ -0,0 +1,81 @@
...
+            if self.ser:
+                self.ser.disconnect()
+        except NameError:
+            pass
+
+    @property
+    def is_connected(self):
+        """True if connected to the hardware or driver."""
+        if self.ser:
+            return self.ser.is_connected
```
wtgee: Seems like we should be setting the `self._is_connected` property and then returning that. Not that someone should be messing with the private variable, but this would get it out of sync.

pocs/dome/abstract_serial_dome.py
```diff
@@ -0,0 +1,81 @@
...
+        return False
+
+    def connect(self):
+        """ Connects to the device via the serial port (`self._port`)
```
wtgee: I still don't like advertising a private variable in the docstring. But I still kind of think the port shouldn't be private anyway.

pocs/dome/abstract_serial_dome.py
```diff
@@ -0,0 +1,81 @@
...
+            self.logger.error('No SerialData instance')
+        elif not self.is_connected:
+            self.logger.debug('Connecting to dome')
+            try:
+                self.ser.connect()
+                self.logger.info('Dome connected: {}'.format(self.is_connected))
+            except OSError as err:
+                self.logger.error("OS error: {0}".format(err))
+            except error.BadSerialConnection as err:
+                self.logger.warning('Could not create serial connection to mount.')
```
wtgee: Still refers to mount here and in line below.

pocs/dome/abstract_serial_dome.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            self.logger.debug('Already connected to dome')
+
+        return self.is_connected
+
+    def disconnect(self):
+        self.logger.debug("Closing serial port for dome")
+        self._is_connected = self.ser.disconnect()
+
+    def verifyConnected(self):
```
wtgee: `verify_connected`

pocs/tests/test_astrohaven_dome.py
```diff
@@ -0,0 +1,75 @@
+# Test the Astrohaven dome interface using a simulated dome controller.
+
+import copy
+import pytest
+import serial
+
+from pocs.dome import CreateDomeFromConfig
```
wtgee: This should be `create_dome_from_config` but I couldn't comment on file up above and must have skipped it in previous review.

pocs/dome/astrohaven.py
```diff
@@ -0,0 +1,127 @@
...
+        v = self.read_latest_state()
+        return v == Protocol.BOTH_CLOSED
+
+    def close(self):
+        self.full_move(Protocol.CLOSE_A, Protocol.A_CLOSE_LIMIT)
+        self.full_move(Protocol.CLOSE_B, Protocol.B_CLOSE_LIMIT)
+        return self.is_closed
+
+    @property
+    def state(self):
```
wtgee: Comment in other PR was about switching this to `status`. You seem to be using `read_latest_state` for all decision making so this seems like it is for a user.

pocs/dome/astrohaven.py
```diff
@@ -0,0 +1,127 @@
...
+        # TODO(jamessynge): Add ability to read the available input from self.ser without
+        # waiting if there is available input. The last received byte is good enough for our
+        # purposes... as long as we drained the input buffer before taking an action.
+        self.ser.reset_input_buffer()
+        data = self.ser.read_bytes(size=1)
+        if len(data):
+            return chr(data[-1])
+        return None
+
+    def nudge_shutter(self, send, target_feedback):
```
wtgee: Docstrings missing. Important here where there are parameters.

pocs/dome/astrohaven.py
```diff
@@ -0,0 +1,127 @@
...
+        self.ser.write(send)
+        # Wait a moment so that the response to our command has time to be emitted, and we don't
+        # get fooled by a status code received at about the same time that our command is sent.
+        time.sleep(0.1)
+        feedback = self.read_latest_state()
+        return feedback == target_feedback
+
+    def full_move(self, send, target_feedback):
+        """Send a command code until the target_feedback is recieved, or a timeout is reached."""
```
wtgee: Document the parameters. Not sure if `send` should be bytes or a string nor what `target_feedback` represents.

pocs/dome/astrohaven.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def nudge_shutter(self, send, target_feedback):
+        self.ser.write(send)
+        # Wait a moment so that the response to our command has time to be emitted, and we don't
+        # get fooled by a status code received at about the same time that our command is sent.
+        time.sleep(0.1)
+        feedback = self.read_latest_state()
+        return feedback == target_feedback
+
+    def full_move(self, send, target_feedback):
```
wtgee: Seems like this should just be named `open_shutter` or `open`. When I think of a dome moving I think of it rotating. For consistency with (eventual) other dome subclasses I think an `open` method makes the most sense. All domes open in some way.

pocs/dome/abstract_serial_dome.py
jamessynge: I've decided that it doesn't make great sense to put the _is_connected in the base class, AbstractDome, in part because of issues exactly like this. So, I'm inclined to respond to this feedback by stripping it out from AbstractDome, and implementing it in the sub-classes as appropriate to the means to establishing that connection.

pocs/dome/abstract_serial_dome.py
jamessynge: OK, will move it to public.

jamessynge: Though I assume you agree that port shouldn't be changed, right?

pocs/dome/abstract_serial_dome.py
jamessynge: Fixed locally, will push later.

pocs/dome/abstract_serial_dome.py
jamessynge: Done locally.

pocs/dome/astrohaven.py
jamessynge: Yes, it is for the end user. Will create another issue/branch/PR for general changes to the dome API.

jamessynge: Filed issue #189 for this.

pocs/dome/astrohaven.py
jamessynge: full_move goes in either direction, i.e. opens or closes, o it moves from whatever the current position is to the commanded position (i.e. until target_feedback is the output).

pocs/dome/astrohaven.py
jamessynge: done

pocs/dome/astrohaven.py
jamessynge: Done.

pocs/tests/test_astrohaven_dome.py
jamessynge: Done.

jamessynge:
PTAL

pocs/utils/rs232.py
```diff
@@ -210,7 +211,7 @@ def __del__(self):
             # If an exception is thrown when running __init__, then self.ser may not have
             # been set, in which case reading that field will generate a NameError.
```
wtgee: Update comment to reflect error.

pocs/dome/abstract_serial_dome.py
```diff
@@ -0,0 +1,82 @@
...
+        self._port = cfg.get('port')
+        if not self._port:
+            msg = 'No port specified in the config for dome: {}'.format(cfg)
+            self.logger.error(msg)
+            raise error.DomeNotFound(msg=msg)
+
+        baudrate = int(cfg.get('baudrate', 9600))
+
+        # Setup our serial connection to the given port.
+        self.ser = None
```
wtgee: Only other comment was about turning this into `self.serial` so we have `self.serial.ser` instead of `self.ser.ser`, which is more consistent with the mount. Any reason you want to keep it this way?

PR #192, opened 2017-12-14
Re-use PanLogger instances if the parameters haven't changed.

### LINKED ISSUE #190: get_root_logger creates a new logger on each call
Should return the same PanLogger instance if the params are the same. I have a fix in progress.

jamessynge:
Fixes #190.

### DIFF
--- a/pocs/__init__.py
+++ b/pocs/__init__.py
@@ -72,7 +72,6 @@ _check_environment()
 
 # Global vars
 _config = None
-_logger = None
 
 
 class PanBase(object):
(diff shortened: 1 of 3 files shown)


jamessynge:
It appears the build failed due to a network issue. Closing and reopening PR to trigger build again.

wtgee:
>It appears the build failed due to a network issue. Closing and reopening PR to trigger build again.

FYI, there is a Restart Build button on travis. Let me know if you don't have access to it.

jamessynge:
As long as importing works correctly, all the imports of pocs.utils.logger will reference the same one module instance, so they'll share the same global (module scope) variable `all_loggers`. And thus a single PanLogger is created.

wtgee:
Let's remove the global `_logger` and just use this setup. It is essentially the same thing.

jamessynge:
PTAL

PR #195, opened 2017-12-15
Observatory Dependency Injection

wtgee:
Simple change to create the `Observatory` outside of POCS and require it as a first parameter.

This will allow for a lot easier use by Huntsman and other observatories.

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -15,6 +15,7 @@ from astropy.io import fits
 from astropy.utils import console
 
 from pocs import POCS
+from pocs.observatory import Observatory
 from pocs.scheduler.field import Field
 from pocs.scheduler.observation import Observation
 from pocs.utils import current_time
@@ -125,7 +126,8 @@ class PocsShell(Cmd):
             simulator = []
 
         try:
-            self.pocs = POCS(simulator=simulator, messaging=True)
+            observatory = Observatory(simulator=simulator)
+            self.pocs = POCS(observatory, messaging=True)
             self.pocs.initialize()
         except error.PanError:
             pass
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -8,7 +8,6 @@ from multiprocessing import Queue
 from astropy import units as u
 
 from . import PanBase
-from .observatory import Observatory
 from .state.machine import PanStateMachine
 from .utils import current_time
 from .utils import get_free_space
@@ -25,6 +24,7 @@ class POCS(PanStateMachine, PanBase):
     the `get_ready()` method the transition that is responsible for moving to the initial state.
 
     Args:
+        observatory(Observatory): An instance of a `pocs.observatory.Observatory` class
         state_machine_file(str): Filename of the state machine to use, defaults to
             'simple_state_table'
         messaging(bool): If messaging should be included, defaults to False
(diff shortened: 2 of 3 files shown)


pocs/core.py
```diff
@@ -38,7 +38,12 @@ class POCS(PanStateMachine, PanBase):
...
 
-    def __init__(self, state_machine_file='simple_state_table', messaging=False, **kwargs):
+    def __init__(
+            self,
+            observatory,
+            state_machine_file='simple_state_table',
+            messaging=False,
+            **kwargs):
 
         # Explicitly call the base classes in the order we want
```
jamessynge: FWIW, this comment and the associated code seem to imply that we should use aggregation for PanStateMachine.

pocs/tests/test_pocs.py
```diff
@@ -9,14 +9,24 @@
 from pocs import POCS
 from pocs import _check_config
 from pocs import _check_environment
+from pocs.observatory import Observatory
```
jamessynge: Ditto on the import style.

pocs/tests/test_pocs.py
```diff
@@ -9,14 +9,24 @@
...
 from pocs.utils import error
 from pocs.utils.messaging import PanMessaging
 
 
 @pytest.fixture
-def pocs(config):
+def observatory():
+    observatory = Observatory(simulator=['all'])
+
```
jamessynge: Drop blank line.

pocs/tests/test_pocs.py
```diff
@@ -177,11 +187,13 @@ def test_is_weather_safe_no_simulator(pocs, db):
...
-def test_run_wait_until_safe(db):
+def test_run_wait_until_safe(db, observatory):
     os.environ['POCSTIME'] = '2016-08-13 23:00:00'
 
     def start_pocs():
-        pocs = POCS(simulator=['camera', 'mount', 'night'],
+        observatory.config['simulator'] = ['camera', 'mount', 'night']
+
+        pocs = POCS(observatory,
                     messaging=True, safe_delay=15)
```
jamessynge: FWIW, looks like safe_delay is one of several undocumented parameters to POCS.

bin/pocs_shell
```diff
@@ -15,6 +15,7 @@ from astropy.io import fits
 from astropy.utils import console
 
 from pocs import POCS
+from pocs.observatory import Observatory
```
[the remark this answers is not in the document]
wtgee: As I remarked in another PR, this was not something that we explicitly decided on, we only talked about using root relative syntax. Either style is correct and the thing that is usually stressed is to be consistent. Since 95% of the current code uses the `from` syntax I'm not sure what the arguments are for in favor of changing them all. It is slightly more obvious what package is being referred to so I could understand wanting to switch to it, but that is a conversation we have not had yet.

pocs/tests/test_pocs.py
wtgee: Created issue #198

bin/pocs_shell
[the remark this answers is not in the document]
wtgee: I think you made a comment but disappeared. I looked at the google style code a bit more and see what they are getting at. It's not explicitly saying not to use the `from` style but to use just the module without prefix, which is not what I am doing. Will fix accordingly.

pocs/core.py
jamessynge: I think this might be the comment that you missed, about the awkwardness of multiple inheritance.

PR #202, opened 2017-12-17
Hue lights

Physarah:
Created a file light_control with function Dome_Lights that connects to the hue bridge via a saved username and controls dome lights and desk lamp. Tested, and no longer changes settings on hue tap device.


pocs/utils/light_control.py
```diff
@@ -0,0 +1,141 @@
+#Basic Python interface for the qhue system.... ip for bridge -> "192.168.0.2"
+import qhue
+from os import path
+from qhue import Bridge, QhueException, create_new_username
```
lspitler: be sure to add `qhue` to `requirements.txt`

pocs/utils/light_control.py
```diff
@@ -0,0 +1,141 @@
...
+import qhue
+from os import path
+from qhue import Bridge, QhueException, create_new_username 
+
+#Setting Up Remote Access to Hue Bridge
+
+def ManualLogin(LED_State, Desk_State, ip):
+    
+    """
+    A way to manually login to the hue bridge, requires the button on the bridge to be pushed.
```
lspitler: Kind of unclear what the purpose of this is. Is it so someone can set the lights to a certain value using python? Why would auto login fail? Also, would it make sense to have 2 separate functions: one for setting up a manual login and one for setting the states? I can imagine the setting the states (given someone has already logged in) would be handy.

pocs/utils/light_control.py
```diff
@@ -0,0 +1,141 @@
...
+    username = qhue.create_new_username(ip)
+    #double check the user name - can be removed later
+    print("The username created is", username)
+    
+    b = qhue.Bridge(ip,username)
+    
+    b.lights[1].state(bri=250,hue=30000,on=LED_State)
+    b.lights[4].state(bri=250,hue=10000,on=Desk_State) 
+    
+    return()
```
lspitler: Here you could return `b` so someone can manipulate the bridge using other functions or even their own function.

Physarah:
closing for PR in AstroHuntsman/POCS

PR #204, opened 2017-12-17
Adding ability to reread the fields file on each new observation

### LINKED ISSUE #203: Re-read targets file
Huntsman added the ability to re-read the targets file so that targets could easily be updated without restart POCS. Want to merge that feature here.

wtgee:
Note that this is just for the dispatch scheduler. Should maybe add into base scheduler.

Closes #203


pocs/tests/test_dispatch_scheduler.py
```diff
@@ -82,6 +98,45 @@ def test_get_observation(scheduler):
     assert isinstance(best[1], float)
 
 
+@pytest.fixture
+def temp_file():
```
wtgee: Could maybe move this into conftest so it is available for all tests.

pocs/tests/test_dispatch_scheduler.py
```diff
@@ -82,6 +98,45 @@ def test_get_observation(scheduler):
...
+
+    yield temp_file
+    os.unlink(temp_file)
+
+
+def test_get_observation_reread(field_list, observer, temp_file):
+    time = Time('2016-08-13 10:00:00')
+
+    # Write out the field list
+    temp_field_file = temp_file
```
jamessynge: Why bother with _renaming_ the variable?

pocs/tests/test_dispatch_scheduler.py
```diff
@@ -82,6 +98,45 @@ def test_get_observation(scheduler):
...
+    with open(temp_field_file, 'w') as f:
+        f.write(yaml.dump(field_list))
+
+    scheduler = Scheduler(observer, fields_file=temp_field_file, constraints=constraints)
+
+    # Get observation as above
+    best = scheduler.get_observation(time=time)
+    assert best[0] == 'HD 189733'
```
jamessynge: FYI, I'm a bit disturbed by the assumptions in a number of these tests about time, location & fields. I think I've said before that I'd like separate, simple yaml files for tests, and preferably different such yaml files for different tests (or perhaps fragments combined together to reduce the number of such files needed).

pocs/scheduler/dispatch.py
```diff
@@ -22,18 +22,24 @@ def __init__(self, *args, **kwargs):
...
                 merit value, defaults to False to only get top value
+            reread_fields_file (bool, optional): If targets file should be reread
+                before getting observation, default False.
 
         Returns:
             tuple or list: A tuple (or list of tuples) with name and score of ranked observations
         """
+        if reread_fields_file:
+            self.logger.debug("Rereading fields file")
+            self.fields_file = self.fields_file
```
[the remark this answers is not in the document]
wtgee: The setter method for the property will re-read the file when set. I should add ac comment about it.

pocs/tests/test_dispatch_scheduler.py
wtgee: Remnant of when I was doing it a different way. Fixed. Thanks.

pocs/tests/test_dispatch_scheduler.py
wtgee: Agreed that it needs to be thought out and made a little more robust.

pocs/tests/test_dispatch_scheduler.py
wtgee: I've moved this into conftest.

PR #207, opened 2017-12-18
Experiment with type annotations in pocs/utils/messaging.py

jamessynge:
I started with:
```
$ monkeytype run `which pytest`
$ monkeytype stub pocs.utils.messaging
$ monkeytype apply pocs.utils.messaging
```
Then added the Message type, and used it in place of the computed value
of the message parameter to send_message and the message value portion
of the return type of receive_message.

### DIFF
--- a/pocs/utils/messaging.py
+++ b/pocs/utils/messaging.py
@@ -1,16 +1,18 @@
 import datetime
+from json import dumps
+from json import loads
 import logging
+from typing import Any, Dict, Union
 import zmq
-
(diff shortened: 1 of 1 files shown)


wtgee:
I like this and think looks good. I think what we should do a point release for POCS soon, before any typing, then aim to add the typing in starting with a new minor release.

jamessynge:
Sounds good. I'll drop this PR and hold off on type annotations for a bit.

PR #208, opened 2017-12-18
Applying a jack-hammer commit to pull in changes from AstroHuntsman/POCS

wtgee:
The two repos had diverged to a large extent and it was to much of a pain
to merge the two. This should be all the changes in Huntmsan that we are
going to pull into all at onces.

Note that this does not include some changes in the Hunstman version that
were done to `rs232.py` and `serial.py` so there might be a little work
after this (for the Huntsman folks).

### DIFF
--- a/pocs/camera/sbig.py
+++ b/pocs/camera/sbig.py
@@ -5,7 +5,6 @@ from astropy import units as u
 from astropy.io import fits
 
 from ..utils import current_time
-from ..utils import error
 from ..utils import images
 from .camera import AbstractCamera
 from .sbigudrv import INVALID_HANDLE_VALUE
@@ -74,7 +73,8 @@ class Camera(AbstractCamera):
 # Methods
 
     def __str__(self):
-        # For SBIG cameras uid and port are both aliases for serial number so shouldn't include both
+        # For SBIG cameras uid and port are both aliases for serial number so
+        # shouldn't include both
         try:
             return "{} ({}) with {} focuser".format(self.name, self.uid, self.focuser.name)
         except AttributeError:
@@ -87,7 +87,8 @@ class Camera(AbstractCamera):
         Gets a 'handle', serial number and specs/capabilities from the driver
 
         Args:
-            set_point (u.Celsius, optional): CCD cooling set point. If not given cooling will be disabled.
+            set_point (u.Celsius, optional): CCD cooling set point. If not
+                given cooling will be disabled.
         """
         self.logger.debug('Connecting to camera {}'.format(self.uid))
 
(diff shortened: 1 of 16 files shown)


pocs/mount/mount.py
```diff
@@ -292,6 +298,20 @@ def get_current_coordinates(self):
 
         return self._current_coordinates
 
+    def distance_from_target(self):
```
wtgee: This is only used by Huntsman so far but it doesn't hurt to have here. Could be convenient.

pocs/camera/sbig.py
```diff
@@ -190,11 +210,21 @@ def take_exposure(self, seconds=1.0 * u.second, filename=None, dark=False, block
 
         assert filename is not None, self.logger.warning("Must pass filename for take_exposure")
 
-        self.logger.debug(
-            'Taking {} second exposure on {}: {}'.format(
-                seconds, self.name, filename))
+        if self.focuser:
```
jamessynge: This should be integrated in a more generic fashion. For example, the base focuser could have a get_fits_headers method.

pocs/camera/sbig.py
```diff
@@ -106,7 +111,7 @@ def connect(self, set_point=None):
         else:
             self.filter_type = 'M'
 
-    def take_observation(self, observation, headers, **kwargs):
+    def take_observation(self, observation, headers=None, filename=None, *args, **kwargs):
```
jamessynge: args is missing from docstring. And why was it added? It doesn't appear to be used.

pocs/camera/sbigudrv.py
```diff
@@ -118,6 +122,17 @@ def __del__(self):
         _ctypes.dlclose(self._CDLL._handle)
         del self._CDLL
```
jamessynge: I'm curious if you ever see an AttributeError from this line when testing. The attribute isn't initialized directly in the ctor (e.g. to None), so an exception in the ctor (`__init__`) could then generate another exception in the dtor (`__del__`).

pocs/camera/sbigudrv.py
```diff
@@ -149,8 +164,7 @@ def assign_handle(self, serial=None):
             try:
                 index = self._handle_assigned.index(False)
             except ValueError:
-                # All handles already assigned, must be trying to intialising more cameras
-                # than are connected.
+                # All handles already assigned, must be trying to intialising more cameras than are connected.
```
jamessynge: c/intialising/initialize/  (or initialise for those raised with the Queen's spelling ;-)).

pocs/mount/mount.py
jamessynge: If you have more than one camera, what does it mean?

pocs/camera/sbig.py
AnthonyHorton: Yes. Reorganisation of the FITS header related code has been on the todo list for a while: https://github.com/AstroHuntsman/HuntsmanPOCSLegacy/issues/28

pocs/mount/mount.py
AnthonyHorton: In the case of Huntsman all the cameras are (roughly) pointing in the same direction so all have a common distance to target.

AnthonyHorton: More generally, as a method of the mount this would refer to the distance on the sky between the target coordinates and the 'mount coordinates', which will always be unique even in the case where multiple cameras are attached to a single mount.

In practice the mount coordinates would be defined by selecting one camera as the reference, and using that camera to do the pointing tests that are used to calibrate the mount pointing model and thereby calculate mount position.

pocs/camera/sbigudrv.py
AnthonyHorton: Initialise it should be :-)

pocs/camera/sbigudrv.py
AnthonyHorton: I don't recall seeing any.

`self._CDLL` is initialised on line 66 by calling the constructor of the `ctypes.CDLL` class, which should always have a `._handle` attribute.

If I recall correctly the destructor completes with no errors but still fails to do what it was added to do, which was to allow unloading and reloading of the SBIG driver library without having to restart the Python interpreter.

pocs/camera/sbig.py
wtgee: I get in the habit of adding `*args, **kwargs` to any method that might be overridden. Not sure that is a great idea but I've had to go back and add them more than I've ever had to remove them. I've looked for recommendations or best practices about this and haven't found anything.

I don't think there was a specific need here.

PR #209, opened 2017-12-18
State location needs to be configurable

wtgee:
Defaults to current location but allows state files to override.


jamessynge:
What happened to the required review?

PR #213, opened 2017-12-18
Let Observatory handle its own initialization

### LINKED ISSUE #212: Let `Observatory` do its own initialization
The `Observatory` object is now created outside of the POCS class so it should handle the initialization of things connected to it. This is just the mount in `pocs.core` right now.

wtgee:
Closes #212

### DIFF
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -130,8 +130,8 @@ class POCS(PanStateMachine, PanBase):
 
             try:
                 # Initialize the mount
-                self.logger.debug("Initializing mount")
-                self.observatory.mount.initialize()
+                self.logger.debug("Initializing observatory")
+                self.observatory.initialize()
 
             except Exception as e:
                 self.say("Oh wait. There was a problem initializing: {}".format(e))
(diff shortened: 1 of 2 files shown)


pocs/core.py
```diff
@@ -129,9 +129,8 @@ def initialize(self):
             self.say("Initializing the system! Woohoo!")
 
             try:
-                # Initialize the mount
-                self.logger.debug("Initializing mount")
-                self.observatory.mount.initialize()
+                self.logger.debug("Initializing observatory")
```
jamessynge: Perhaps adjust the comment on line 72 to not claim the observatory is being created.
And in the ctor docstring on line 29, could you clarify the condition that the observatory instance is supposed to be in (esp. not yet initialized)?

And FWIW, it would be nice to continue with dependency injection for the Observatory class.

wtgee: Thanks, got those updated. Agree with continuing on with the Observatory dependency injection. Perhaps that should be after a patch release, which I'd still like to get done relatively soon.

PR #216, opened 2017-12-19
Adding some docstrings for the config loaded and save

wtgee:
Also makes `_parse_config` private. `save_config` hasn't changed, just added docstring and moved it above private `_parse_config`.

### DIFF
--- a/pocs/utils/config.py
+++ b/pocs/utils/config.py
@@ -8,7 +8,46 @@ from warnings import warn
 
 
 def load_config(config_files=None, simulator=None, parse=True, ignore_local=False):
-    """ Load configuation information """
+    """Load configuation information
+
+    This function supports loading of a number of different files. If no options
+    are passed to `config_files` then the default `$POCS/conf_files/pocs.yaml`
+    will be loaded. See Notes for additional information.
+
+    Notes:
+        `config_files` supports a number of options:
+        * `config_files` is a list and loaded in order, so the first entry
+            will have any values overwritten by similarly named keys in
+            the second entry.
+        * Entries can be placed in the `$POCS/conf_files` folder and
+            should be passed as just the file name, e.g.
+            [`weather.yaml`, `email.yaml`] for loading
+            `$POCS/conf_files/weather.yaml` and `$POCS/conf_files/email.yaml`
+        * The `.yaml` extension will be added if not present, so list can
+            be writen as just ['weather', 'email']
+        * `config_files` can also be specified by an absolute path, which
+            can exist anywhere on the filesystem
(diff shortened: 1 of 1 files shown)


pocs/utils/config.py
```diff
@@ -8,7 +8,46 @@
...
 def load_config(config_files=None, simulator=None, parse=True, ignore_local=False):
-    """ Load configuation information """
+    """Load configuation information
+
+    This function supports loading of a number of different files. If no options
+    are passed to `config_files` then the default `$POCS/conf_files/pocs.yaml`
+    will be loaded. See Notes for additional information.
+
+    Notes:
+        `config_files` supports a number of options:
```
jamessynge: Should that be: `load_config` supports a number of options?

If not, then some other clarification is in order.

pocs/utils/config.py
```diff
@@ -8,7 +8,46 @@
...
+        * `config_files` is a list and loaded in order, so the first entry
+            will have any values overwritten by similarly named keys in
+            the second entry.
+        * Entries can be placed in the `$POCS/conf_files` folder and
+            should be passed as just the file name, e.g.
+            [`weather.yaml`, `email.yaml`] for loading
+            `$POCS/conf_files/weather.yaml` and `$POCS/conf_files/email.yaml`
+        * The `.yaml` extension will be added if not present, so list can
+            be writen as just ['weather', 'email'].
```
jamessynge: c/writen/written/

pocs/utils/config.py
```diff
@@ -8,7 +8,46 @@
...
+        * `config_files` can also be specified by an absolute path, which
+            can exist anywhere on the filesystem.
+        * Local versions of files can override built-in versions and are
+            automatically loaded if placed in the `$POCS/conf_files` folder.
+            The files have a `<>_local.yaml` name, where `<>` is the built-in
```
jamessynge: Seems like we should generate the local file name from the `path` variable at line 77, and not assume that the file was in config_dir. Specifically, replace in `path`  the trailing ".yaml" with "_local.yaml".

wtgee: Yes, I like the idea of saving to a local file more but would rather do it in a separate PR. 

> not assume that the file was in config_dir

Confused about this comment.

pocs/utils/config.py
wtgee: The note is specifically for all the options that `config_files` supports. I made a small change, does that help anything?

pocs/utils/config.py
wtgee: Create https://github.com/panoptes/POCS/issues/238

PR #220, opened 2017-12-19
Dark subtraction for autofocus

AnthonyHorton:
Also bug fixes.

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -222,7 +222,7 @@ class AbstractCamera(PanBase):
                                       focus_range=focus_range,
                                       focus_step=focus_step,
                                       keep_files=keep_files,
-                                      take_dark=take_dark
+                                      take_dark=take_dark,
                                       thumbnail_size=thumbnail_size,
                                       merit_function=merit_function,
                                       merit_function_kwargs=merit_function_kwargs,
@@ -231,10 +231,20 @@ class AbstractCamera(PanBase):
                                       blocking=blocking,
                                       *args, **kwargs)
 
-    def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_files=False, *arg, **kwargs):
+    def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_file=False, *arg, **kwargs):
         """
+        Takes an image and returns a thumbnail.
+
         Takes an image, grabs the data, deletes the FITS file and
-        returns a thumbnail from the centre of the iamge.
+        returns a thumbnail from the centre of the image.
+
+        Args:
+            seconds (astropy.units.Quantity): exposure time, Quantity or numeric type in seconds.
+            file_path (str): path to (temporarily) save the image file to.
+            thumbnail_size (int): size of the square region of the centre of the image to return.
+            keep_file (bool, optional): if True the image file will be deleted, if False it will
+                be kept.
+            *args, **kwargs: passed to the take_exposure() method
         """
         exposure = self.take_exposure(seconds, filename=file_path, *args, **kwargs)
         exposure.wait()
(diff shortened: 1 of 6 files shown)


pocs/camera/camera.py
```diff
@@ -221,12 +231,13 @@ def autofocus(self,
...
 
-    def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_files=False):
+    def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_files=False, *arg, **kwargs):
         """
         Takes an image, grabs the data, deletes the FITS file and
         returns a thumbnail from the centre of the iamge.
         """
-        self.take_exposure(seconds, filename=file_path, blocking=True)
+        exposure = self.take_exposure(seconds, filename=file_path, *args, **kwargs)
+        exposure.wait()
```
wtgee: Should we give this a timeout and do something after?

pocs/focuser/birger.py
```diff
@@ -56,7 +56,7 @@ def __init__(self,
                  name='Birger Focuser',
                  model='Canon EF-232',
                  initial_position=None,
-                 dev_node_pattern='/dev/tty.USA49WG*.?',
+                 dev_node_pattern='/dev/tty.USA49*.?',
```
wtgee: This is ultimately dependent on those acroname usb hubs, so we should make an appropriate comment on what this means.

pocs/focuser/focuser.py
```diff
@@ -228,6 +244,29 @@ def autofocus(self,
...
+                                                   self._camera.uid,
+                                                   start_time,
+                                                   "dark",
+                                                   self._camera.file_extension)
+            self.logger.debug('Taking dark frame {} on camera {}'.format(file_path, self._camera))
+            try:
+                dark_thumb = self._camera.get_thumbnail(seconds,
+                                                        file_path,
+                                                        thumbnail_size,
+                                                        keep_files=True,
```
wtgee: Should this always be `True` or should it be the `keep_files` param? Same with `dark` below, should maybe be `autofocus_take_dark`

pocs/camera/camera.py
```diff
@@ -213,6 +221,8 @@ def autofocus(self,
         return self.focuser.autofocus(seconds=seconds,
                                       focus_range=focus_range,
                                       focus_step=focus_step,
+                                      keep_files=keep_files,
+                                      take_dark=take_dark
```
wtgee: Oh right, the missing comma.

pocs/camera/camera.py
AnthonyHorton: This should have a timeout, put possibly the place for it is in the take_exposure() method of the camera class?

pocs/focuser/birger.py
AnthonyHorton: Yes, should have better documentation.  This is specific to the Tripplite Keyway USA-49 USB-serial adaptors, rather than the usb hubs. At least I made it an argument and only hard coded the default value...

pocs/focuser/focuser.py
AnthonyHorton: I'm causing confusion here by using the same argument name for two different purposes. In `get_thumbnail` `keep_files` simply determines whether the one FITS file it creates gets immediately deleted or not, but in the `autofocus` and `_autofocus` methods `keep_files` means keep _all_ files when `True`, and keep _some_ files when `False`.  The dark frame is one of the files I decided it should always keep, along with the initial and final focus images.

pocs/camera/camera.py
AnthonyHorton: Dammit.

pocs/focuser/focuser.py
AnthonyHorton: Argument of `get_thumbnail` changed from `keep_files` to `keep_file` to slightly reduce confusion, docstring rewritten.

pocs/focuser/birger.py
AnthonyHorton: Docstring written.

PR #221, opened 2017-12-19
Simple script to start message forwarder

wtgee:
Note: This is not ideal because of the hard-coded ports but the fix will
have to wait until we can figure out the port issue in general.

### DIFF
--- a/scripts/start_messenger.py
+++ b/scripts/start_messenger.py
@@ -5,7 +5,7 @@ from pocs.utils.messaging import PanMessaging
 from_port = 6510
 to_port = 6511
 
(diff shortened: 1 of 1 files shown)


scripts/start_messenger.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+
+from pocs.utils.messaging import PanMessaging
+
+from_port = 6510
+to_port = 6511
+
+print("Staring message forwarding, hit Ctrl-c to stop")
```
AnthonyHorton: Staring -> Starting

PR #225, opened 2017-12-20
Explicitly convert equinox to float for FITS header

wtgee:
https://archive.stsci.edu/fits/fits_standard/node40.html#SECTION00942270000000000000

Note: We assume jyear for these. Might need to be more robust in future


wtgee:
@lspitler You made some comments in https://github.com/AstroHuntsman/huntsman-pocs/issues/24#issuecomment-352971579, does this address those?

PR #226, opened 2017-12-20
Small bug when setting up simulators

wtgee:


wtgee:
> Looks fine. What problem was this solving?

I was about to answer your question and then realized the answer didn't make sense. There was an issue when using a camera simulator wherein it was looking for certain autofocus config items but not finding them (a KeyError). This would have handled that properly. I'll try to recreate issue before merging this.

wtgee:
This change is not necessary. See https://github.com/AstroHuntsman/huntsman-pocs/pull/27

PR #228, opened 2017-12-20
Share a single MongoClient instance

### LINKED ISSUE #193: PanBase creates a PanMongo instance for each created sub-class
This seems likely to be a bad idea.

jamessynge:
Caches clients using a weakref so that the client is discarded if there are no referrers.

Fixes #193.

### DIFF
--- a/pocs/utils/database.py
+++ b/pocs/utils/database.py
@@ -1,20 +1,35 @@
-import os
-import pymongo
-
-import gzip
-import json
-
-from bson import json_util
 from datetime import date
 from datetime import datetime
+import gzip
+import json
+from bson import json_util
+import os
+import pymongo
 from warnings import warn
+import weakref
 
 from pocs.utils import current_time
+
+_shared_mongo_clients = weakref.WeakValueDictionary()
+
+
+def get_shared_mongo_client(host, port, connect):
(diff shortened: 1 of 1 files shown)


pocs/utils/database.py
```diff
@@ -41,6 +57,7 @@ def __init__(self, db='panoptes', host='localhost', port=27017, connect=False, *
             'weather',
         ]
 
+        # Why not: db_handle = self._client.db
```
wtgee: Good question. Since no comment could probably be changed.

pocs/utils/database.py
```diff
@@ -1,34 +1,50 @@
-import os
-import pymongo
-
+from datetime import date
```
wtgee: Curious why these were moved to the top and `os` and `pymongo` moved down.

jamessynge: Putting standard python libraries at the top, then third party, then our own. A style I'm used to... from you know where.
 Probably should discuss whether we should do that globally before making such changes.

wtgee: I was mostly confused by moving the `os` below the `datetime`. Makes sense for `pymongo`. Not a huge deal.

PR #231, opened 2017-12-21
Add a 'dome' attribute to Observatory.

jamessynge:
Add logger kwarg to create_dome_from_config, enabling caller
to provide logger to be used.

### DIFF
--- a/pocs/dome/__init__.py
+++ b/pocs/dome/__init__.py
@@ -4,23 +4,19 @@ import pocs
 import pocs.utils
 import pocs.utils.logger as logger_module
 
-# A dome needs a config. We assume that there is at most one dome in the config,
-# i.e. we don't support two different dome devices, such as might be the case
-# if there are multiple independent actuators, for example slit, rotation and
-# vents.
 
-
-def create_dome_from_config(config):
+def create_dome_from_config(config, logger=None):
     """If there is a dome specified in the config, create a driver for it.
 
     A dome needs a config. We assume that there is at most one dome in the config, i.e. we don't
     support two different dome devices, such as might be the case if there are multiple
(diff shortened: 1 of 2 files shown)


pocs/observatory.py
```diff
@@ -488,6 +498,8 @@ def _create_mount(self, mount_info=None):
         Note:
             This does not actually make a serial connection to the mount. To do so,
             call the 'mount.connect()' explicitly.
+            TODO(jamessynge): Discuss this claim with Wilfred. SerialData automatically
```
wtgee: The claim is wrong as you point out and should be updated.

pocs/dome/__init__.py
```diff
@@ -30,10 +26,10 @@ def create_dome_from_config(config):
     else:
         brand = dome_config.get('brand')
         driver = dome_config['driver']
```
wtgee: This already made it's way in, but this will fail if `dome` is in config but `driver` is not.

wtgee: Perhaps that was intended but if so we should catch and warn.

jamessynge: Good point. This is certainly a fatal error in terms of creating the dome instance, and will most likely keep POCS from working as intended. Probably we should just raise an exception.

pocs/observatory.py
```diff
@@ -128,6 +135,9 @@ def status(self):
                     status['mount']['mount_target_ha'] = self.observer.target_hour_angle(
                         t, self.mount.get_target_coordinates())
 
+            if self.dome:
+                status['dome'] = self.dome.status
```
wtgee: I realize this has already been merged, but it's a little odd that `status` is a property here but a method for everything else. Why not make the dome status a method?

PR #241, opened 2017-12-22
Remove nested collection name

### LINKED ISSUE #237: Mongo collections are nested with a hardcoded name
The mongo collections are stored under `<db_name>.panoptes.<collection>` where the `<db_name>` comes from the config file and the `<collection` depends on the collection names. However it doesn't need the `panoptes` in the middle and shouldn't be hard-coded regardless.

wtgee:
Closes #237

### DIFF
--- a/pocs/utils/database.py
+++ b/pocs/utils/database.py
@@ -57,13 +57,12 @@ class PanMongo(object):
             'weather',
         ]
 
-        # Why not: db_handle = self._client.db
-        db_handle = getattr(self._client, db)
+        db_handle = self._client.db
 
         # Setup static connections to the collections we want
         for collection in self.collections:
             # Add the collection as an attribute
-            setattr(self, collection, getattr(db_handle, 'panoptes.{}'.format(collection)))
+            setattr(self, collection, getattr(db_handle, collection))
 
     def insert_current(self, collection, obj, include_collection=True):
         """Insert an object into both the `current` collection and the collection provided


pocs/utils/database.py
```diff
@@ -57,13 +57,12 @@ def __init__(self, db='panoptes', host='localhost', port=27017, connect=False):
...
 
-        # Why not: db_handle = self._client.db
-        db_handle = getattr(self._client, db)
+        db_handle = self._client.db
 
         # Setup static connections to the collections we want
         for collection in self.collections:
             # Add the collection as an attribute
-            setattr(self, collection, getattr(db_handle, 'panoptes.{}'.format(collection)))
+            setattr(self, collection, getattr(db_handle, collection))
```
jamessynge: This strikes me as a big change. Why are there not more files impacted?

And how does this impact operational systems, i.e. PAN001, etc.

wtgee: Everything accesses mongo via this file using the attributes that are set up so has not direct knowledge of the underlying collection name.  There might be a notebook I have that is impacted but it's not in one of the committed notebooks.

For PAN001 it will silently just start using a new collection. There is a [`renameCollection`](https://docs.mongodb.com/manual/reference/method/db.collection.renameCollection/) command that can easily rename the existing collections but since it is only PAN001 at this point it's not really worth create a repo script for it.

jamessynge: Why make this change? There is also a 2.5 months of weather data in PAN006. Note tremendously valuable, but I just don't understand why this change is being made.

For example, you could have made this more capable by checking for existing collections, and refusing to start until the user runs a tool that does the renameCollection commands.

wtgee: The change is being made because it is hard-coded value that is misleading and unnecessary. If anything it should be `pocs` so that the db can be `panoptes` and the collection namespace is `pocs`. That way it would read something like `panoptes.pocs.weather` rather than a redundant `panoptes.panoptes.weather`. However, it should not be hard-coded regardless. 

It will be much easier to make the change now than when there are more units.

This PR doesn't have that `pocs` level namespace feature but upon writing the above I think that maybe it should. It could also actually be something like `pan001` so full name was `db.unit.collection`, e.g. `panoptes.pan001.weather`.

> For example, you could have made this more capable by checking for existing collections, and refusing to start until the user runs a tool that does the renameCollection commands.

Because that would pollute the code with a one-time operation that applies to two people. The one-time operation takes less than one minute to implement. There's no reason to build that into the code. The old data can also be easily exported and imported into new name without losing anything.

jamessynge: I surrender.

wtgee: Have any thoughts on adding the unit name as mentioned above?

Also, I'm realizing there is no Issue created for this, but one of the identified projects for someone to work on is to totally redo how we store info in mongo, which will be a much larger change than this. Basically, how things are saved now is not optimal for time series data and we create a lot of records (one for each reading), which is not ideal.

We should probably work on getting that change in sooner than later if we decided it's necessary. I'll create an Issue for it shortly based of the project description.

jamessynge: I have a little concern about making the data store different on each unit, but it doesn't make too much difference since PanMongo has attributes such as weather, rather than <unitname>.weather.

So, should we have a small collection (a singleton) with info about the unit, i.e. the the name of the unit? I suppose that is dangerously similar to the config files, but it would travel with the database if copied. Hmm, not sure whether to add that small collection. But, go ahead with the unit name change.

Issue #199, opened 2017-12-17
Integrate dome support into POCS and Observatory


jamessynge:
For Wheaton (PAN006) and the Huntsman Array, we need to enable the addition of dome support into the POCS' state machine and Observatory's set of devices, but in a way which doesn't impact deployments without domes.

jamessynge:
#231 Added a dome attribute to Observatory, minimally.

jamessynge:
I've now tested (and debugged) this with the Astrohaven dome at Wheaton.

PR #248, opened 2017-12-24
Add minimal support for domes into POCS.

jamessynge:
Basically just support for **closing** the dome when done. Still need to figure out how to integrate the dome into the approach state transition. Advice sought.

Tests TBS.

Partially address #199.


pocs/core.py
```diff
@@ -41,6 +41,7 @@ class POCS(PanStateMachine, PanBase):
     Attributes:
         name (str): Name of PANOPTES unit
         next_state (str): The next state for the state machine
+            TODO(jamessynge): Why document this here? It is an attribute of the base class.
```
wtgee: At one point I was going to document all the attributes of PanBase and PanStateMachine here so this is the only place you would have to look. That never happened and isn't a good idea, so this is a remnant.

pocs/core.py
```diff
@@ -295,7 +299,11 @@ def is_safe(self, no_warning=False):
             if no_warning is False:
                 self.logger.warning('Unsafe conditions: {}'.format(is_safe_values))
 
-            # Not safe so park unless we are not active
+            # Not safe, so close the dome and park the mount.
+            if not self.observatory.close_dome():
+                self.logger.critical('Unable to close dome!')
+
+            # TODO(jamessynge): Determine why parking is conditional. Is self.park() idempotent?
```
wtgee: If you try to `park` from a state that doesn't support the transition it will complain. The `transitions` library has since added support for nested states as well as other features that could clean this up, but we should discuss in another PR.

pocs/core.py
```diff
@@ -223,6 +224,9 @@ def power_down(self):
                 "Shutting down {}, please be patient and allow for exit.".format(
                     self.name))
 
+            if not self.observatory.close_dome():
+                self.logger.critical('Unable to close dome!')
```
wtgee: This is perhaps one of those places where we need to somehow add email support.

pocs/core.py
```diff
@@ -488,6 +498,8 @@ def _check_messages(self, queue_type, q):
     def _interrupt_and_park(self):
         self.logger.info('Park interrupt received')
         self._interrupted = True
+        # TODO(jamessynge): Should we close the dome here? Or extend the state machine
+        # with dome related states and actions?
```
wtgee: This will call `park` from any state that it is currently in and move to the `parking` state. I think it makes most sense to close the dome in the `parking` state itself and leave this as is.

pocs/core.py
```diff
@@ -461,6 +469,8 @@ def _check_environment(self):
             os.makedirs("{}/logs".format(pandir))
 
     def _check_messages(self, queue_type, q):
+        # TODO(jamessynge): Where do these messages come from? PAWS?
+        # TODO(jamessynge): Determine what to do about the dome.
```
wtgee: Regarding comment below, I don't think this needs to be changed.

pocs/core.py
jamessynge: Removed next_state from docstring.

pocs/core.py
jamessynge: Makes sense. Perhaps we should extend the idea of the "say" method to include a "yell" method, designed to ask for help.

pocs/core.py
jamessynge: OK, I'll move it, but I'm concerned that POCS.power_down() assumes that the only device that needs to be dealt with for parking is the mount, and so makes decisions about whether to change state based on whether the mount is connected. @wtgee What do you suggest I do about that?

pocs/core.py
jamessynge: Do the messages come from PAWS?

pocs/core.py
jamessynge: Got it. I've been wondering about having separate state machines for the different equipment, since it doesn't make sense to continue expanding the base state machine.

Issue #246, opened 2017-12-23
_config in pocs/base.py makes tests non-hermetic


jamessynge:
The global variable _config gets updated based on the config= kwarg passed into PanBase. There is no way to reset that in tests.

We may want to introduce a PanConfig object, passed around to objects rather than having a global variable, and with the abilities of the config.py to load and save.

Assigning to @wtgee for consideration. I'm willing to implement.

PR #249, opened 2017-12-24
Make PanBase hermetic across tests by clearing _config each time

jamessynge:
A simple fix for issue #246.

### DIFF
--- a/pocs/base.py
+++ b/pocs/base.py
@@ -10,8 +10,8 @@ from pocs.utils.logger import get_root_logger
 _config = None
 
 
-def reset_config_for_test():
-    """Reset the shared _config to None.
+def reset_global_config():
+    """Reset the global _config to None.
 
     Globals such as _config make tests non-hermetic. Enable conftest.py to clear _config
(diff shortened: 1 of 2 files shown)


pocs/base.py
```diff
@@ -10,6 +10,16 @@
 _config = None
 
 
+def reset_config_for_test():
```
wtgee: What about just making this an option that is passed to PanBase, like `ignore_local_config`? I can imagine there might be a situation where someone needs to reset the config for whatever reason. Or just change this function to `reset_config`.

jamessynge: There are many instances of PanBase in a running POCS system, and it doesn't make sense (to me) for them to have different global configs.

I suggested elsewhere that we might have a PanConfig object that gets passed around, and represents whatever config has been loaded by the creator of POCS and Observatory. This fix is designed to allow us to make a decision about that latter, yet still get closer to hermetic tests now.

So, I'll rename to reset_global_config, with a comment that it is intended for use in tests, but might have other uses.

PR #254, opened 2017-12-27
Logging updates

### LINKED ISSUE #253: Remove PanLogger class
PanLogger exists to handle formatting in the more modern Python style, but it turns out that Python's logging.Formatter class takes a style kwarg on the ctor which controls whether legacy or modern formatting is used. The benefit of using a "real" logger is that the source (file name and line number) of the log messages should then be correct, and not always in pocs/utils/logger.py.

wtgee:
* Replace PanLogger with custom `LogRecord` class
	* Allow logger to use either `%` or `{}` (str.format) style for actual
log messages.
* Log messages show correct file/lineno
* Add timezone info to log line
* Reorder log.yaml

Fixes #253

Now you can:
```
pocs.logger.debug("Hello {}", 42)        # Hello 42
pocs.logger.debug("Hello {:.02f}", 42)   # Hello 42.00
pocs.logger.debug("Hello %f", 42)        # Hello 42.00000
pocs.logger.debug("Hello {:.02f}%", 42)  # Hello 42.00%
pocs.logger.debug("Hello {:.02f}%f", 42) # Hello {:.02f}42.000000 - See below
```

For the last example we could double-interpolate it, but I don't think it is too large of an issue.

Note that this performs legacy interpolation first (via `%`) and catches exceptions and then tries the `str.format`. Might be more efficient to reverse order but I don't think the impact is large.

Full log output example:
```
MainProcess(19834) MainThread 20171227015442GMT INFO     messaging.py        : 122:send_message              PANCHAT Hi there!
```

### DIFF
--- a/conf_files/log.yaml
+++ b/conf_files/log.yaml
@@ -3,26 +3,12 @@ logger:
     use_utc: True
     formatters:
       simple:
-        format: '%(asctime)s UTC - %(message)s'
+        format: '%(asctime)s - %(message)s'
         datefmt: '%H:%M:%S'
       detail:
-        format: '%(processName)s(%(process)d) %(threadName)s %(asctime)14s UTC %(levelname)8s %(filename)20s:%(lineno)4d:%(funcName)-25s %(message)s'
-        datefmt: '%Y%m%d%H%M%S'
-
-    loggers:
-      all:
-        handlers: [all]
-        propagate: true
-      info:
-        handlers: [info]
-        propagate: true
-      warn:
-        handlers: [warn]
-        propagate: true
-      error:
-        handlers: [error]
-        propagate: true        
-
+        style: '{'
+        format: '{processName}({process:d}) {threadName} {asctime:14s} {levelname:8s} {filename:20s}:{lineno:4d}:{funcName:25s} {message}'
+        datefmt: '%Y%m%d%H%M%S%Z'
     handlers:
       all:
         class: logging.handlers.TimedRotatingFileHandler
@@ -48,8 +34,19 @@ logger:
         formatter: detail
         when: W6
         backupCount: 4        
-
+    loggers:
+      all:
+        handlers: [all]
+        propagate: true
+      info:
+        handlers: [info]
+        propagate: true
+      warn:
+        handlers: [warn]
+        propagate: true
+      error:
+        handlers: [error]
+        propagate: true 
     root:
       level: DEBUG
       handlers: [all, warn]
-      propagate: true
--- a/conf_files/log.yaml
+++ b/conf_files/log.yaml
@@ -7,8 +7,9 @@ logger:
         datefmt: '%H:%M:%S'
       detail:
         style: '{'
-        format: '{processName}({process:d}) {threadName} {asctime:14s} {levelname:8s} {filename:20s}:{lineno:4d}:{funcName:25s} {message}'
-        datefmt: '%Y%m%d%H%M%S%Z'
+        # See FilenameLineFilter in logger.py for fileline description
+        format: '{levelname:.1s}{asctime}.{msecs:03.0f} {fileline:20s} {message}'
+        datefmt: '%m%d %H:%M:%S'
     handlers:
       all:
         class: logging.handlers.TimedRotatingFileHandler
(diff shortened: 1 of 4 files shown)


pocs/utils/logger.py
```diff
@@ -9,47 +9,30 @@
...
+        msg = str(self.msg)
+        if self.args:
+            try:
+                msg = msg % self.args
+            except (TypeError, ValueError):
+                msg = msg.format(*self.args)
+        return msg
 
 
 def get_root_logger(profile='panoptes', log_config=None):
```
jamessynge: Please document params, especially point to https://docs.python.org/3/library/logging.config.html#configuration-dictionary-schema to explain log_config.

pocs/utils/logger.py
```diff
@@ -87,11 +70,11 @@ def get_root_logger(profile='panoptes', log_config=None):
         full_log_fname = '{}/{}-{}.log'.format(log_dir, log_fname, handler)
         log_config['handlers'][handler].setdefault('filename', full_log_fname)
 
-        # Setup the TimeedRotatingFileHandler for middle of day
+        # Setup the TimedRotatingFileHandler for middle of day
```
jamessynge: A TODO for the future: remove explicit reference to TimedRotatingFileHandler, and instead check for atTime being a string in the log_config, and if so parsing it with some flexibility regarding the time format.

pocs/utils/logger.py
```diff
@@ -87,11 +70,11 @@ def get_root_logger(profile='panoptes', log_config=None):
...
         log_config['handlers'][handler].setdefault('atTime', datetime.time(hour=11, minute=30))
 
         if handler == 'all':
-            # Symlink the log file to $PANDIR/logs/panoptes.log
+            # Symlink the log file to log_symlink
```
jamessynge: I think of this as the other way around:

```python
# Create a symlink to the log file with just the name of the script, not the date and pid,
# as this makes it easier to find the latest such log file when debugging.
```

conf_files/log.yaml
```diff
@@ -3,26 +3,12 @@ logger:
...
-        handlers: [info]
-        propagate: true
-      warn:
-        handlers: [warn]
-        propagate: true
-      error:
-        handlers: [error]
-        propagate: true        
-
+        style: '{'
```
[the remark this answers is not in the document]
wtgee: Yes, good idea. My next step was to reduce the fixed portion and I like this suggestion.

One note, I did originally add the thread ID in but it was giving me a 12 digit string, which seemed odd. I'll play a little more.

pocs/utils/logger.py
```diff
@@ -9,47 +9,30 @@
...
 
-    def error(self, fmt, *args, **kwargs):
-        self.logger.error(self._process_str(fmt, *args, **kwargs))
+class StrFormatLogRecord(logging.LogRecord):
+    """ Allow for `str.format` style log messages
 
-    def critical(self, fmt, *args, **kwargs):
-        self.logger.critical(self._process_str(fmt, *args, **kwargs))
+    Even though you can select '{' as the style for the formatter class,
+    you still can't use {} formatting for your message. The custom
```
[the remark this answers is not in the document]
wtgee: No worries, I'm glad you suggested it as it forced me to dig up the correct method. I think the documents are very confusing in terms of what it allows. Honestly, I was very surprised there wasn't an existing module that would just do this, but I couldn't find anything.

And the incorrect file and line number was really starting to bother me, so it was a good time for it.

pocs/utils/logger.py
wtgee: Not sure I understand what that would get us but let's discuss in a separate Issue.

conf_files/log.yaml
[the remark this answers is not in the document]
wtgee: Confirmed, the thread id is a very large number. Current format with latest commit is:

```
I1228 10:05:01.572 logger.py:112        ****************************** Starting PanLogger ******************************                      
I1228 10:05:04.897 observatory.py:35    Initializing observatory                                                                     
I1228 10:05:04.897 observatory.py:38            Setting up location                                                                
D1228 10:05:04.897 observatory.py:483   Setting up site details of observatory   
```

Note that we lose time zone information by adding the msecs. Could get back in there with a little bit of work but not entirely sure how important it is.

PR #262, opened 2017-12-29
Fixing `EQUINOX` fits header so we don't see warnings

wtgee:
Changing from `'J2000   '` to `2000.0`


wtgee:
FYI: https://en.wikipedia.org/wiki/Equinox_(celestial_coordinates)

This PR was just following the FITS header convention of making EQUINOX a float.

Issue #267, opened 2017-12-29
peas_shell logging errors


jamessynge:
After updating POCS on PAN006 to include the piaa imaging fix, peas_shell is still spewing logging related errors.

```python
panoptes@Panmass:/var/panoptes/POCS$ bin/peas_shell 
Welcome to PEAS Shell! Type ? for help
PEAS > load_environment
Loading sensors
PEAS > load_weather
Loading AAG Cloud Sensor on /dev/ttyUSB1
--- Logging error ---
Traceback (most recent call last):
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 992, in emit
    msg = self.format(record)
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 838, in format
    return fmt.format(record)
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 578, in format
    s = self.formatMessage(record)
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 547, in formatMessage
    return self._style.format(record)
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 399, in format
    return self._fmt.format(**record.__dict__)
KeyError: 'fileline'
Call stack:
  File "bin/peas_shell", line 296, in <module>
    PanSensorShell().cmdloop()
  File "/home/panoptes/anaconda3/lib/python3.6/cmd.py", line 138, in cmdloop
    stop = self.onecmd(line)
  File "/home/panoptes/anaconda3/lib/python3.6/cmd.py", line 217, in onecmd
    return func(arg)
  File "bin/peas_shell", line 148, in do_load_weather
    self.weather = AAGCloudSensor(serial_address=port, use_mongo=True)
  File "/var/panoptes/POCS/peas/weather.py", line 127, in __init__
    self.logger.info('Connecting to AAG Cloud Sensor')
Message: 'Connecting to AAG Cloud Sensor'
Arguments: ()
--- Logging error ---
Traceback (most recent call last):
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 992, in emit
    msg = self.format(record)
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 838, in format
    return fmt.format(record)
  File "/home/panoptes/anaconda3/lib/python3.6/logging/__init__.py", line 578, in format
    s = self.formatMessage(record)
[...]

PR #269, opened 2017-12-29
Remove imports of Python's logging module from PEAS code in POCS.

jamessynge:
The import was interfering with the PANOPTES logger in peas_shell.
Add command history to peas_shell.

### DIFF
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -2,6 +2,7 @@
 import cmd
 import os
 import readline
+import sys
 
 from astropy.utils import console
 from pprint import pprint
@@ -11,15 +12,9 @@ from peas.sensors import ArduinoSerialMonitor
 from peas.weather import AAGCloudSensor
 from peas.webcam import Webcam
 
-import logging
-
 from pocs.utils.config import load_config
+from pocs.utils.logger import get_root_logger
 from pocs.utils.database import PanMongo
-
-log_level = {
-    'info': logging.INFO,
-    'debug': logging.DEBUG,
-}
 
 
 class PanSensorShell(cmd.Cmd):
@@ -86,6 +81,9 @@ class PanSensorShell(cmd.Cmd):
         Arguments:
             sensor {str} -- environment, weather, webcams
         """
+        # TODO(jamessynge): We currently use a single logger, not one per module or sensor.
+        # Figure out whether to keep this code and make it work, or get rid of it.
+        import logging
         get_level = {
             logging.DEBUG: logging.INFO,
             logging.INFO: logging.DEBUG,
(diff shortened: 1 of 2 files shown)


wtgee:
Does this fix #267 ?

bin/peas_shell
```diff
@@ -86,6 +81,9 @@ class PanSensorShell(cmd.Cmd):
         Arguments:
             sensor {str} -- environment, weather, webcams
         """
+        # TODO(jamessynge): We currently use a single logger, not one per module or sensor.
+        # Figure out whether to keep this code and make it work, or get rid of it.
```
wtgee: We are not really using the `name` (we call it `profile`) attribute of the logger, which we could start doing. This would allow us to write to one log file but filter based on the name, which could be a module or sensor name.

peas/weather.py
```diff
@@ -103,8 +103,7 @@ class AAGCloudSensor(object):
 
     def __init__(self, serial_address=None, use_mongo=True):
         self.config = load_config(config_files='peas')
-        self.logger = logging.getLogger('aag-cloudsensor')
-        self.logger.setLevel(logging.INFO)
+        self.logger = get_root_logger()
```
wtgee: As commented above, we could pass `aag-cloudsensor` as the profile here and then change the log output format to include this `name`, which would be easily filterable.

This file does output a lot of debug information each time it queries the weather, however, which isn't a bad thing but mostly ends up being noise.

jamessynge:
Yes, this fixes #267.

bin/peas_shell
jamessynge: When you say filter, do you mean grep the log file, or filter to decide whether the message is logged. That latter is what I'd like to get out of a change w.r.t. multiple loggers. In particular, it would be nice to be able to enable emitting of debug log messages of selected modules at run time (e.g. by a command-line flag).

peas/weather.py
jamessynge: I looked at the log file, and I agree, lots of ~~noise~~ details. :wink:

jamessynge:
I'll submit an issue for working on per-module loggers, enabling the filtering of log records before output at the module level and/or augmenting the output with the logger name so that filtering can be done after the fact. I'm not as keen on the latter because it makes the lines long, and slightly duplicates the file & line info we have in the log file already.

wtgee:
It seems like if we filter before output that would then make it impossible to go back and recover any information that might be missing unless you are also writing out a separate log file that has no filters. But yes, could discuss in another Issue. This is okay to merge.

jamessynge:
Submitted issue #270.

PR #279, opened 2017-12-30
Split image utils

wtgee:
I've been wanting to split this for a while as I think it will make maintenance and testing easier but it's not required. Let me know thoughts on whether this is useful or not.


pocs/utils/images/focus.py
```diff
@@ -0,0 +1,83 @@
+import numpy as np
+
+
+def focus_metric(data, merit_function='vollath_F4', **kwargs):
```
jamessynge: This seems to have hit the YAGNI level, as there is only a single merit function (which seems like it should be called metric_function).
Do you plan to add Vollath's F5 or other such metric?

pocs/utils/images/fits.py
```diff
@@ -0,0 +1,322 @@
...
+from astropy import units as u
+
+from pocs.utils import error
+
+
+def solve_field(fname, timeout=15, solve_opts=[], **kwargs):
+    """ Plate solves an image.
+
+    Args:
+        fname(str, required):       Filename to solve in .fits extension.
```
jamessynge: Astrometry's `solve_field` supports a variety of input image formats, including JPEG and PNG, not just FITS. Perhaps `solve_field()` and `get_solve_field()` should be in a location other than the fits specific file?

pocs/focuser/focuser.py
```diff
@@ -12,7 +12,7 @@
 
 from pocs import PanBase
 from pocs.utils import current_time
-from pocs.utils import images
+from pocs.utils.images import focus as focus_utils
```
jamessynge: I notice that you always end up renaming it to focus_utils. How about renaming the file to focus_utils.py?

pocs/images.py
```diff
@@ -10,7 +10,7 @@
 from collections import namedtuple
 
 from pocs import PanBase
-from pocs.utils import images as img_utils
+from pocs.utils.images import fits as fits_utils
```
jamessynge: Ditto with fits_utils.py.

pocs/tests/test_utils.py
```diff
@@ -7,7 +7,9 @@
 from astropy.io import fits
 
 from pocs.utils import current_time
-from pocs.utils import images
+from pocs.utils import images as img_utils
```
jamessynge: Do you plan to split the tests into multiple files?

pocs/focuser/focuser.py
wtgee: I thought that was a little odd since it was already in `utils` folder. I suppose it makes the most sense though.

pocs/tests/test_utils.py
wtgee: Yes, that's a good idea, I will do that as well.

pocs/utils/images/fits.py
wtgee: The `solve_field` program does support more options but our wrappers around them do not. In the exiting files there was some old code for how we handle `CR2`, but none of that was relevant any more so I removed it for this PR.

The `solve_field()` we have could be more generic, but `get_solve_field()` definitely assumes FITS (despite some of the comments in the docstring, which I just noticed and will clean).

We could put `solve_field` somewhere else and then change `get_solve_field` to `solve_fits_field`.

pocs/utils/images/focus.py
wtgee: Pinging @AnthonyHorton on this one. I think the plan was to add more merit functions originally, not sure if that is still the plan.

wtgee: But don't think it should be changed as part of this PR regardless.

jamessynge: Agreed, I'm just raising the issue now as this is the first time I've reviewed this code.

pocs/focuser/focuser.py
wtgee: The more I think about this the less I like renaming it. It could potentially be annoying left as is, but I think it's a cleaner name. I'm going to leave as it is now unless there are strong feelings either way.

wtgee:
Note that I added tests that close #265 

Also, switched a default param in the `solve_field` to be correct.

wtgee:
+2.03% Woohoo! :)

jamessynge:
Awesome coverage improvement!

pocs/focuser/focuser.py
jamessynge: Leave it.

bin/pocs_shell
```diff
@@ -310,7 +313,7 @@ Hardware names: {}   (or all for all hardware)'''.format(
 
         self.pocs = None
 
-    def do_polar_alignment_test(self, *arg):
+    def do_polar_alignment_utils_test(self, *arg):
```
jamessynge: I think this change was a mistake, i.e. a global search and replace of polar_alignment to polar_alignment_utils that went too far.

wtgee: Oops, yeah. Good case for why this file needs testing. :) I'll get it changed.

pocs/utils/images/focus.py
AnthonyHorton: @wtgee is correct. I expected to have to experiment with several different merit functions, and possibly even use more than one depending on the situation (e.g. one for coarse focus and a different one for fine focus). As it turned out the first thing I tried (Vollath's F4) seemed to work tolerably well, at least for fine focusing. It's not great though, and I do still expect to try some alternatives. https://github.com/AstroHuntsman/HuntsmanPOCSLegacy/issues/30

Good point on the nomenclature, though. Either metric and merit function would be appropriate in this context, but should settle on one of them instead of using both.

PR #280, opened 2017-12-31
Auto-detect Arduinos attached to USB serial ports

### LINKED ISSUE #277: Re-enable auto_detect of Arduinos
The Arduinos aren't always associated with the same /dev/ttyACM device, so the currently disabled auto_detect feature would be valuable.

jamessynge:
Fix #277.

This is only lightly tested. I still need to connect a suitable Arduino to my laptop for proper testing.

### DIFF
--- a/peas/sensors.py
+++ b/peas/sensors.py
@@ -9,13 +9,10 @@ from pocs.utils.rs232 import SerialData
 
 
 class ArduinoSerialMonitor(object):
+    """Monitors the serial lines and tries to parse any data recevied as JSON.
 
-    """
-        Monitors the serial lines and tries to parse any data recevied
-        as JSON.
-
-        Checks for the `camera_box` and `computer_box` entries in the config
-        and tries to connect. Values are updated in the mongo db.
+    Checks for the `camera_box` and `computer_box` entries in the config and tries to connect.
+    Values are updated in the mongo db.
     """
 
     def __init__(self, auto_detect=False, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


peas/sensors.py
```diff
@@ -25,45 +21,25 @@ def __init__(self, auto_detect=False, *args, **kwargs):
...
         self.db = None
         self.messaging = None
 
         # Store each serial reader
         self.serial_readers = dict()
 
-        if auto_detect:
+        if auto_detect or self.config['environment'].get('auto_detect', False):
+            self.logger.debug('Performing auto-detect')
             for port_num in range(9):
```
wtgee: We can also grab connected ports and just check those: https://pythonhosted.org/pyserial/tools.html?highlight=comport#serial.tools.list_ports.comports. We could then check vendor id and product id. The arduino uno even reports a `serial_number` correctly although I don't think the micros do.

That could probably help clean up some of the `_auto_detect_port` below.

peas/sensors.py
```diff
@@ -141,20 +150,20 @@ def capture(self, use_mongo=True, send_message=True):
                 if send_message:
                     self.send_message({'data': data}, channel='environment')
             except yaml.parser.ParserError:
```
wtgee: I think I didn't know this when I wrote this a long time ago, but you can pass a tuple for except:

```
except (yaml.parser.ParserError, ValueError, TypeError, Exception) as e:
```

However if we have `Exception` at the end and we are not doing anything different for the different exceptions then really we just need to capture the generic.

peas/sensors.py
jamessynge: Submitted #284 for this.

peas/sensors.py
jamessynge: Hmm, I didn't know that either, though as you say it doesn't do much to include both a class and super-class in the the same tuple.

PR #282, opened 2017-12-31
Use symlinks to get around Arduino IDE limitations

### LINKED ISSUE #281: Enable Arduino IDE to work with camera & telemetry boards
We have some of the required libraries copied into the POCS repo (but not all), however the Arduino IDE doesn't make it easy to use libraries that aren't installed in a user or system-wide Arduino directory.

Fix this by using symlinks. This works on Linux, should work on MacOS, and may even work on modern versions of Windows.

jamessynge:
Add dallas-temperature-control library, used by the telemetry board sketch.

Fixes #281.


resources/arduino_files/libraries/dallas-temperature-control/DallasTemperature.h
```diff
@@ -0,0 +1,242 @@
+#ifndef DallasTemperature_h
+#define DallasTemperature_h
+
+#define DALLASTEMPLIBVERSION "3.7.2"
```
wtgee: I notice that "3.7.9" is the latest for this. Any reason to start with an old one here? I realize the features probably haven't changed much, just wondering.

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -1,8 +1,9 @@
 #include <Wire.h>
```
wtgee: We are using `OneWire` below but `Wire` here, which are two different libraries that do the same thing. We should look into just use one of them.

jamessynge: Agreed. I'll submit an issue to consolidate.

resources/arduino_files/libraries/dallas-temperature-control/DallasTemperature.h
jamessynge: Just what I had already been using. Since I'm not doing extensive testing of this, I'd prefer to hold off changes until able to do so.

jamessynge:
I'll strip out the examples, then commit.

PR #291, opened 2018-01-03
Update camera and telemetry board sketches to those in use on PAN006

jamessynge:
These have the benefit that they produce their reports faster because they first gather the data needed before they start writing. This reduces the risk that POCS will timeout while reading a single report.

In addition, the parsing of relay changes in the telemetry sketch is improved so that it can't accidentally shutdown everything.

This includes PR #290, and should be reviewed after that is submitted.

### DIFF
--- a/resources/arduino_files/camera_board/camera_board.ino
+++ b/resources/arduino_files/camera_board/camera_board.ino
@@ -5,162 +5,261 @@
 #include "Adafruit_Sensor.h"
 #include "DHT.h"
 
-#define DHT_TYPE DHT22 // DHT 22  (AM2302)
-
-/* DECLARE PINS */
 const int DHT_PIN = 9;      // DHT Temp & Humidity Pin
-const int CAM_01_RELAY = 5;
-const int CAM_02_RELAY = 6;
+const int CAM_0_RELAY = 5;
+const int CAM_1_RELAY = 6;
 const int RESET_PIN = 12;
 
+// Utitlity Methods
 
-/* CONSTANTS */
-Adafruit_MMA8451 accelerometer = Adafruit_MMA8451();
+void toggle_led() {
+  digitalWrite(LED_BUILTIN, !digitalRead(LED_BUILTIN));
+}
 
-// Setup DHT22
-DHT dht(DHT_PIN, DHT_TYPE);
+void turn_pin_on(int pin_num) {
+  digitalWrite(pin_num, HIGH);
+}
 
-int led_value = LOW;
+void turn_pin_off(int pin_num) {
+  digitalWrite(pin_num, LOW);
+}
+
+// IO Handlers.
+
+class AccelerometerHandler {
+  public:
+    void init() {
+      ready_ = false;
+      has_reading_ = false;
+      ready();
+    }
+    void collect() {
+      if (ready()) {
+        accelerometer_.getEvent(&event_);
+        orientation_ = accelerometer_.getOrientation(); // Orientation
+        has_reading_ = true;
+      }
+    }
+    bool report() {
+      if (has_reading_) {
+        Serial.print(", \"accelerometer\":{\"x\":");
+        Serial.print(event_.acceleration.x);
+        Serial.print(", \"y\":");
+        Serial.print(event_.acceleration.y);
+        Serial.print(", \"z\":");
+        Serial.print(event_.acceleration.z);
+        Serial.print(", \"o\": "); Serial.print(orientation_);
+        Serial.print("}");
+        return true;
+      }
+      return false;
+    }
+    bool ready() {
+      if (!ready_) {
+        ready_ = accelerometer_.begin();
+        if (!ready_) {
+          Serial.println("Accelerometer not ready, or not present.");
+        } else {
+          // Check Accelerometer range
+          // accelerometer.setRange(MMA8451_RANGE_2_G);
+          // Serial.print("Accelerometer Range = "); Serial.print(2 << accelerometer.getRange());
+          // Serial.println("G");
+        }
+      }
+      return ready_;
+    }
+
+  private:
+    Adafruit_MMA8451 accelerometer_;
+    sensors_event_t event_;
+    uint8_t orientation_;
+    bool ready_;
+    bool has_reading_;
+} acc_handler;
+
+class DHTHandler {
+  public:
+    DHTHandler() : dht_(DHT_PIN, DHT22) {}  // AM2302
+
+    void init() {
+      dht_.begin();
+    }
+
+    void collect() {
+      // Force readHumidity to actually talk to the device;
+      // otherwise will read at most every 2 seconds, which
+      // is sometimes just a too far apart.
+      // Note that the underlying read() routine has some big
+      // delays (250ms and 40ms, plus some microsecond scale delays).
+      humidity_ = dht_.readHumidity(/*force=*/true);
+      // readTemperature will use the data collected by
+      // readHumidity, which is just fine.
+      temperature_ = dht_.readTemperature();
+    }
+
+    void report() {
+      Serial.print(", \"humidity\":");
+      Serial.print(humidity_);
+      Serial.print(", \"temp_00\":");
+      Serial.print(temperature_);
+    }
+
+  private:
+    DHT dht_;
+    float humidity_ = 0;
+    float temperature_ = 0;  // Celcius
+} dht_handler;
+
+unsigned long end_setup_millis;
+unsigned long next_report_millis;
+int report_num = 0;
+
+class LedHandler {
+  public:
+    void init() {
+      pinMode(LED_BUILTIN, OUTPUT);
+      digitalWrite(LED_BUILTIN, false);
+
+      // Provide a visible signal that setup has been entered.
+      if (Serial) {
+        // 2 seconds of fast blinks.
+        for (int i = 0; i < 40; ++i) {
+          delay(50);
+          toggle_led();
+        }
+        Serial.println("LED blink complete");
+      } else {
+        // 2 seconds of slow blinks.
+        for (int i = 0; i < 10; ++i) {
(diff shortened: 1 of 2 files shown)


resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+    led_handler.update();
+    dht_handler.collect();
+    led_handler.update();
+    bool cam0 = digitalRead(CAM_0_RELAY);
+    bool cam1 = digitalRead(CAM_1_RELAY);
+
+    // Format/output the results.
+    Serial.print("{\"name\":\"camera_board\", \"count\":");
+    led_handler.update();
+    Serial.print(millis() - end_setup_millis);
```
wtgee: `millis()` rolls over to 0 when long-running (docs say about 50 days) so this doesn't seem entirely robust.

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+    led_handler.update();
+    Serial.print(", \"num\":");
+    led_handler.update();
+    Serial.print(report_num);
+    led_handler.update();
+    Serial.print(", \"inputs\":");
```
wtgee: `inputs` was not at all clear to me until reading below. Maybe name the key `inputs_recvd` or something similar.

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
-      break;
+      case CAM_0_RELAY:
+      case CAM_1_RELAY:
+        if (pin_status == 1) {
+          turn_pin_on(pin_num);
+        } else if (pin_status == 0) {
+          turn_pin_off(pin_num);
+        }
+        break;
+      case LED_BUILTIN:
```
wtgee: We might be able to remove this since the led_handler will just clobber it anyway.

resources/arduino_files/telemetry_board/telemetry_board.ino
```diff
@@ -1,259 +1,603 @@
+/*
+   The PANOPTES Baseline Unit (January 2017) telemetry board controls 5 relays on
+   the V0 power board, and reads 4 current sensors, 3 temperature sensors and one
+   combined humidity & temperature sensor. Each pass through loop() it checks for
+   serial input regarding switching the relays, and every 2 seconds it prints
+   out the values of the sensors and the settings of the relays.
+
+   This program has a class for each type of sensor, with a common API:
+   1) An Init() method that is called from setup.
```
wtgee: Camera board has these methods as lower case names. We can make it conform in a PR that moves things to a shared library if you want.

resources/arduino_files/telemetry_board/telemetry_board.ino
```diff
@@ -1,259 +1,603 @@
...
 
-// Temperature chip I/O
 OneWire ds(DS18_PIN);
-DallasTemperature sensors(&ds);
-
-// Setup DHT22
-DHT dht(DHT_PIN, DHTTYPE);
 
-int led_value = LOW;
+// Utitlity Methods
```
wtgee: Copy and paste spelling error.

resources/arduino_files/telemetry_board/telemetry_board.ino
```diff
@@ -1,259 +1,603 @@
...
-      case WEATHER_RELAY:
-      case MOUNT_RELAY:
-        if (pin_status == 1) {
-          turn_pin_on(pin_num);
-        } else if (pin_status == 0) {
-          turn_pin_off(pin_num);
-        } else if (pin_status == 9) {
-          toggle_pin(pin_num);
+      }
+      // Sort the devices so we get a consistent order from run to run...
```
wtgee: I'm surprised we need to do this. Is it due to DallasTemperature or the underlying OneWire?

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
-
-/* DECLARE PINS */
 const int DHT_PIN = 9;      // DHT Temp & Humidity Pin
-const int CAM_01_RELAY = 5;
-const int CAM_02_RELAY = 6;
+const int CAM_0_RELAY = 5;
+const int CAM_1_RELAY = 6;
 const int RESET_PIN = 12;
 
+// Utitlity Methods
```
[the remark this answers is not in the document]
jamessynge: Fixed.

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+    void init() {
+      ready_ = false;
+      has_reading_ = false;
+      ready();
+    }
+    void collect() {
+      if (ready()) {
+        accelerometer_.getEvent(&event_);
+        orientation_ = accelerometer_.getOrientation(); // Orientation
+        has_reading_ = true;
```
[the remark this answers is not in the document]
jamessynge: `getEvent()` doesn't provide a status output, so we can't tell if something went wrong.

Do you mean can `has_reading_` be wrong?

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+      }
+      return false;
+    }
+    bool ready() {
+      if (!ready_) {
+        ready_ = accelerometer_.begin();
+        if (!ready_) {
+          Serial.println("Accelerometer not ready, or not present.");
+        } else {
+          // Check Accelerometer range
```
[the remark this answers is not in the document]
jamessynge: Let's leave it in for the moment so that it is in the history, and so that we can figure out "what next".

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+    DHTHandler() : dht_(DHT_PIN, DHT22) {}  // AM2302
+
+    void init() {
+      dht_.begin();
+    }
+
+    void collect() {
+      // Force readHumidity to actually talk to the device;
+      // otherwise will read at most every 2 seconds, which
+      // is sometimes just a too far apart.
```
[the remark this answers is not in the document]
jamessynge: done.

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+      // Note that the underlying read() routine has some big
+      // delays (250ms and 40ms, plus some microsecond scale delays).
+      humidity_ = dht_.readHumidity(/*force=*/true);
```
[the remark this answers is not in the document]
jamessynge: It is a way to provide semantics for the human reader. When we read a call with literal numbers and/or bools as args, we can't readily tell what the literal means. This style adds the name of the parameter before the actual arg value as a comment. We use it at work to help with readability.

resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -5,162 +5,261 @@
...
+      // readTemperature will use the data collected by
+      // readHumidity, which is just fine.
+      temperature_ = dht_.readTemperature();
+    }
+
+    void report() {
```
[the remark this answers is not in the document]
jamessynge: How about the other way: remove the bool return from `AccelerometerHandler::report()` as the value returned isn't currently used.

resources/arduino_files/camera_board/camera_board.ino
jamessynge: I'll add a TODO, as I don't want to make many changes while "just" bringing these already running versions, especially without doing extensive testing.

resources/arduino_files/camera_board/camera_board.ino
jamessynge: OK

resources/arduino_files/camera_board/camera_board.ino
jamessynge: OK

resources/arduino_files/telemetry_board/telemetry_board.ino
jamessynge: Good point. I'll clearly be forced to resolve that when extracting the shared classes.

resources/arduino_files/telemetry_board/telemetry_board.ino
jamessynge: Done.

resources/arduino_files/telemetry_board/telemetry_board.ino
jamessynge: I don't now recall if there is an inconsistency in the order from run-to-run, but I assume so since I went to the trouble. Fortunately, I did buy some more of those sensors and can test sometime Real Soon Now :registered:. The main point is that it doesn't make a lot of sense to have 3 temps listed, but not have any idea which is which.

resources/arduino_files/camera_board/camera_board.ino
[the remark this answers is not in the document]
wtgee: Yes, wondering if `getEvent()` could somehow fail and then `has_reading` would be outdated. Since `report()` just checks for `has_reading` it would just keep reporting the same values and we wouldn't be the wiser. Would be nice to reset `has_reading` to `false` each time somehow. Might not be necessary though.

resources/arduino_files/camera_board/camera_board.ino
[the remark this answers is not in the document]
wtgee: Got it, that makes sense. Thanks.

resources/arduino_files/camera_board/camera_board.ino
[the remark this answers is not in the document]
wtgee: Yes, sounds good. Just going for consistency so if not used let's definitely remove.

PR #296, opened 2018-01-04
Suppress Arduino's update check on startup

jamessynge:
This is an attempt to see if I can suppress some of the DNS error messages in the travis log.


jamessynge:
The arduino CLI does generate travis errors as we desire if the code is bad. In this case, a change to the arduino CLI in version 1.8.2 introduced the WARN messages related to DNS, so a known issue:

https://github.com/per1234/arduino-ci-script/issues/1
https://forum.arduino.cc/index.php?topic=469428

I'm going to try an alternate solution: Use 1.8.1.

PR #297, opened 2018-01-04
Use older Arduino build to avoid warnings

jamessynge:
Use older Arduino build that (apparently) doesn't emit noisy DNS related WARN messages.


jamessynge:
This works (i.e. doesn't have lots of noisy messages unrelated to our sketches), so please review.

jamessynge:
I've added an ARDUINO_VERSION variable. I'll merge after the tests pass.

Issue #289, opened 2018-01-03
Share code between the Arduino sketches


jamessynge:
We've got duplicated classes between the 3 sketches. Move those to a shared library.

jamessynge:
Fixed with several PRs that created symlinks from the sketch folders to files in arduino_files/shared.

PR #302, opened 2018-01-06
Share DHTHandler between sketches.

jamessynge:
More partial work on #289.
Note that the use in power_board.ino is minimal for now. I plan to make power_board.ino more like telemetry_board.ino soon.

### DIFF
--- a/resources/arduino_files/camera_board/camera_board.ino
+++ b/resources/arduino_files/camera_board/camera_board.ino
@@ -3,13 +3,16 @@
 
 #include "Adafruit_MMA8451.h"
 #include "Adafruit_Sensor.h"
-#include "DHT.h"
+#include "dht_handler.h"
 #include "PinUtils.h"
 
 const int DHT_PIN = 9;      // DHT Temp & Humidity Pin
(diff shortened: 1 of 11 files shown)


resources/arduino_files/camera_board/camera_board.ino
```diff
@@ -152,7 +125,7 @@ void setup(void) {
   turn_pin_on(CAM_1_RELAY);
 
   acc_handler.init();
-  dht_handler.init();
+  dht_handler.Init();
```
wtgee: Is the plan to make the shared libraries use the `Init` vs `init`?

jamessynge: Yes, if you don't mind. I'm comfortable with the Google C++ Style Guide, where accessors are lower case (e.g. `float humidity() const;` or `void set_humidity(float v);`, and other methods are upper case (e.g. `float ComputeSomeValue();`).

wtgee: Yeah, that's fine and good to know. I'm not familiar with good c++ coding styles.

PR #303, opened 2018-01-06
Extract DallasTemperatureHandler out of telemetry_board.ino.

jamessynge:
Not currently used elsewhere, but I expect it will be used
in power_board.ino soon.

### DIFF
--- a/resources/arduino_files/shared/dallas_temperature_handler.h
+++ b/resources/arduino_files/shared/dallas_temperature_handler.h
@@ -1,0 +1,98 @@
+#ifndef RESOURCES_ARDUINO_FILES_SHARED_DALLAS_TEMPERATURE_HANLDER_H
+#define RESOURCES_ARDUINO_FILES_SHARED_DALLAS_TEMPERATURE_HANLDER_H
+
+#include "OneWire.h"
+#include "DallasTemperature.h"
+
+// DallasTemperatureHandler collects temp values from Dallas One-Wire temp sensors.
+template <size_t kMaxSensors>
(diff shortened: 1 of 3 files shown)


resources/arduino_files/shared/dallas_temperature_handler.h
```diff
@@ -0,0 +1,98 @@
...
+    void Init() {
+      dt_.begin();
+      dt_.setWaitForConversion(true);
+
+      uint8_t devices = dt_.getDeviceCount();
+      for (uint8_t device_num = 0; device_num < devices; ++device_num) {
+        if (dt_.getAddress(devices_[device_count_].address, device_num)) {
+          devices_[device_count_].index = device_num;
+          ++device_count_;
+          if (device_count_ >= kMaxSensors) {
```
wtgee: If we didn't have `kMaxSensors` here would it work for an arbitrary amount? With how I set up the trucker board it's easy to wire in new sensors (they are chained in the wires and only have one input to board) so I was thinking about adding more since I have more. What's the benefit of hard-coding the 3 below if we can detect them this way?

jamessynge: We don't have dynamic memory allocation available to us on the Arduino, so we need to pre-allocate arrays, such as for storing info about the set of available devices or the sensor readings.
We could eliminate this storage if we reverted back to printing data as we read it from each sensor, but then we'd have the problems we encountered in the past.

wtgee: Got it, thanks for the explanation.

PR #306, opened 2018-01-06
Add completion of simulator names on setup_pocs command line.

jamessynge:
Add help (docstring) to commands in PocsShell where missing.
Cleanup comments.

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -82,7 +82,7 @@ class PocsShell(Cmd):
         i.cmdloop()
 
     def do_start_messaging(self, *arg):
-        """Starts the messaging system for the POCS ecosystem
+        """Starts the messaging system for the POCS ecosystem.
 
         This starts both a command forwarder and a message forwarder as separate
         processes.
@@ -169,7 +169,7 @@ Hardware names: {}   (or all for all hardware)'''.format(
         self.pocs = None
 
     def do_run_pocs(self, *arg):
-        """Make POCS `run` in a separate process
+        """Make POCS `run` in a separate process.
 
         Note:
             Run with `--simulator=all` for simulator
@@ -193,7 +193,7 @@ Hardware names: {}   (or all for all hardware)'''.format(
             print_warning('Please run `setup_pocs` before trying to run')
 
     def do_collect_status(self, *arg):
-        """ Just records the `status` for pocs
+        """Just records the `status` for pocs.
 
         This would mostly be used when you want to record information from
         POCS but might be driving the system manually
@@ -220,7 +220,7 @@ Hardware names: {}   (or all for all hardware)'''.format(
             print_warning('Please run `setup_pocs` before trying to run')
 
     def do_pocs_command(self, cmd):
-        """ Send a command to POCS instance
+        """Send a command to POCS instance.
 
         Arguments:
             cmd {str} -- Command to be sent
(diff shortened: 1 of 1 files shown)


bin/pocs_shell
```diff
@@ -75,14 +75,14 @@ class PocsShell(Cmd):
...
-        """Enter the drift alignment shell """
+        """Enter the drift alignment shell."""
         self.do_reset_pocs()
         print_info('*' * 80)
         i = DriftShell()
         i.cmdloop()
 
     def do_start_messaging(self, *arg):
-        """ Starts the messaging system for the POCS ecosystem
+        """Starts the messaging system for the POCS ecosystem
```
wtgee: Period here too.

I wish there were a way in reviews to make very simple edits like this...

bin/pocs_shell
```diff
@@ -156,11 +156,20 @@ simulate is a space-separated list of hardware to simulate.
 Hardware names: {}   (or all for all hardware)'''.format(
             ','.join(hardware.get_all_names())))
 
+    def complete_setup_pocs(self, text, line, begidx, endidx):
```
wtgee: Can we move this down to the other Utility methods section? The organization scheme is entirely my own across a lot of our files but I think makes sense. Let me know if disagree.

wtgee: I only see the `text` param being used. Are the others still required?

bin/pocs_shell
```diff
@@ -156,11 +156,20 @@ simulate is a space-separated list of hardware to simulate.
...
+        Does NOT park the mount, nor execute power_down.
+        """
         self.pocs = None
 
     def do_run_pocs(self, *arg):
-        """ Make POCS `run` in a separate process
+        """Make POCS `run` in a separate process
 
         Note:
             Run with `--simulator=all` for simulator
```
wtgee: Should update this comment to reflect how `simulator` is used.

bin/pocs_shell
```diff
@@ -233,32 +242,37 @@ Hardware names: {}   (or all for all hardware)'''.format(
...
         try:
             self.pocs.observatory.mount.park()
             self.pocs.say("Mount parked")
         except Exception as e:
-            print_warning('Problem unparking: {}'.format(e))
+            print_warning('Problem parking: {}'.format(e))
 
     def do_go_home(self, *arg):
-        """ Move the mount to home """
+        """Move the mount to home"""
```
wtgee: Tricky periods. :)

bin/pocs_shell
jamessynge: They are all passed in by Cmd when it asks for a set of completions. I assume that Python would complain if we didn't declare them.

jamessynge: Regarding moving the completion_* (and help_*) methods, I'd like to keep those methods grouped together by command name. Perhaps we should add a delineating comment (i.e. ####...) between commands?

bin/pocs_shell
jamessynge: I don't know what to update it with. The comment looks wrong, but I've not executed this command enough to know what to update this with... except the missing period at the end! ;-)

bin/pocs_shell
jamessynge: Done.

bin/pocs_shell
jamessynge: Done.

bin/pocs_shell
wtgee: Back in https://github.com/panoptes/POCS/commit/3cf2831b0dda9ecf3b8d70ab64069789d06a4c94#diff-f081d3c1efc3be3fbd0aa696e79e797a commit you had switched it do a space-separated list and without the `--simulator=`. The `Note:` here refers to commands to pass to `setup_pocs`, which is probably confusing.

PR #307, opened 2018-01-06
Fix mistake introduced in #279

wtgee:
See comment here: https://github.com/panoptes/POCS/pull/279#pullrequestreview-87077713


wtgee:
Dupe of #305

PR #308, opened 2018-01-08
Fix IntervalTimer so that it doesn't screw up the remainder; I found

jamessynge:
this happened when we paused for 30 seconds for a computer power-cycle.
It is possible this is also why PAN006's telemetry board is not
responding.

Extract SerialInputHandler out of telemetry_board.ino, adding the
ability to specify the pin to be changed by name.

If power-cycling the computer, continue printing reports, which
makes debugging easier.

### DIFF
--- a/resources/arduino_files/shared/PinUtils.h
+++ b/resources/arduino_files/shared/PinUtils.h
@@ -9,4 +9,7 @@ bool is_pin_on(int pin_num);
 void toggle_pin(int pin_num);
 void toggle_led();
 
+// Read the mode (INPUT, OUTPUT, or INPUT_PULLUP) for a pin.
+int pinMode(int pin);
+
 #endif  // RESOURCES_ARDUINO_FILES_SHARED_PIN_UTILS_H
--- a/resources/arduino_files/shared/PinUtils.h
+++ b/resources/arduino_files/shared/PinUtils.h
@@ -9,7 +9,7 @@ bool is_pin_on(int pin_num);
 void toggle_pin(int pin_num);
 void toggle_led();
 
-// Read the mode (INPUT, OUTPUT, or INPUT_PULLUP) for a pin.
+// Returns the mode (INPUT, OUTPUT, or INPUT_PULLUP) for a pin.
 int pinMode(int pin);
 
 #endif  // RESOURCES_ARDUINO_FILES_SHARED_PIN_UTILS_H
--- a/resources/arduino_files/shared/interval_timer.cpp
+++ b/resources/arduino_files/shared/interval_timer.cpp
@@ -1,0 +1,41 @@
+#include "interval_timer.h"
+
+#include "Arduino.h"
+
+IntervalTimer::IntervalTimer(millis_t interval_ms)
+    : IntervalTimer(interval_ms, interval_ms) {}
+
+IntervalTimer::IntervalTimer(millis_t interval_ms, millis_t remaining_ms)
+    : last_time_(millis()), remaining_(remaining_ms), interval_(interval_ms) {}
+
+void IntervalTimer::Reset() {
+  last_time_ = millis();
+  remaining_ = interval_;
+}
+
+bool IntervalTimer::HasExpired() {
+  millis_t now = millis();
+  millis_t elapsed;
+  if (now < last_time_) {
+    // We've had wrap around of the millisecond clock. When we do so, we keep things simple by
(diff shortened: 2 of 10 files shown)


resources/arduino_files/shared/PinUtils.h
```diff
@@ -9,4 +9,7 @@ bool is_pin_on(int pin_num);
 void toggle_pin(int pin_num);
 void toggle_led();
 
+// Read the mode (INPUT, OUTPUT, or INPUT_PULLUP) for a pin.
+int pinMode(int pin);
```
wtgee: Should the return value go somewhere?

resources/arduino_files/shared/interval_timer.cpp
```diff
@@ -0,0 +1,41 @@
+#include "interval_timer.h"
```
wtgee: Seems like this is something that should already exist publicly.

resources/arduino_files/shared/serial_input_handler.h
```diff
@@ -0,0 +1,112 @@
...
+      return false;
+    }
+
+    // Allow the input line to end with NL, CR NL or CR.
+    bool IsNewLine(int c) {
+      return c == '\n' || c == '\r';
+    }
+
+    void LineNotMatched(int reason) {
+      Serial.print("LINE NOT MATCHED, reason=");
```
wtgee: Since we are listening on the other end for valid JSON responses, should we be outputting some JSON here?

resources/arduino_files/telemetry_board/telemetry_board.ino
```diff
@@ -107,6 +112,22 @@ class BaseNameHandler {
       Serial.print("\":");
     }
 
+    bool NameEquals(const char* s, uint8_t len) {
```
wtgee: It's not all intuitive to me what is going on here and since this is in an actual file people might look at (as opposed to the shared files) maybe some comments or description of what is happening.

resources/arduino_files/telemetry_board/telemetry_board.ino
```diff
@@ -248,104 +273,66 @@ void Report(unsigned long now) {
...
-          if (!input_buffer_.Append(static_cast<char>(c))) {
-            wait_for_new_line_ = true;
-          }
-        } else {
-          // Input is not an acceptable character.
-          wait_for_new_line_ = true;
+bool restarting_computer = false;
+IntervalTimer restart_computer_timer(POWER_CYCLE_MS);
+
+void HandleNumEqNum(uint8_t pin_num, uint8_t pin_status) {
```
wtgee: So has the default changed from `1,1` to `1=1`? It looks like the serial input handler does both but you are using the latter version here.

jamessynge:
I like the JSON message idea. I'll do that in another PR. Submitted issue #309.

resources/arduino_files/shared/PinUtils.h
jamessynge: I don't understand. Does the comment/function make more sense if I documented it as:
```c++
// Return the mode (INPUT, OUTPUT, or INPUT_PULLUP) for a pin.
```
(i.e. change **Read** to **Return**).

resources/arduino_files/shared/interval_timer.cpp
jamessynge: It does, and I stole from it when I wrote the original code last summer. But it is part of a much larger library.

resources/arduino_files/shared/serial_input_handler.h
jamessynge: Agreed. I'll do that in a follow on PR.

resources/arduino_files/telemetry_board/telemetry_board.ino
jamessynge: Oh, good idea.

resources/arduino_files/telemetry_board/telemetry_board.ino
jamessynge: No, but I misnamed that method. Should have been HandleNumCommaNum.

resources/arduino_files/shared/serial_input_handler.h
wtgee: Note #310  that I just made.

wtgee:
> I like the JSON message idea. I'll do that in another PR. Submitted issue #309.

Note #310 that I just made.

Issue #43, opened 2016-06-29
Pretty Pictures


wtgee:
Determine best way to do pretty pictures. This could be manually, automatic for each image, only for first image, etc. Also figure out best techniques for generating pretty pictures automatically. Possible develop web-based interface for PAWS that would allow unit owners to dynamically alter images.

wtgee:
I've re-added our old pretty image process in #1258. This issue is mostly stale so is getting closed in favor of more specific work tasks to be done in the future as identified.

Issue #232, opened 2017-12-21
Fix make_pretty_image for FITS files


wtgee:
Image currently doesn't show anything and is far from pretty. Need to change the private function [`_make_pretty_from_fits`](https://github.com/AstroHuntsman/POCS/blob/develop/pocs/utils/images.py#L477).

See also #43

wtgee:
#319 in progress.

jermainegug:
#319, I believe is ready to be merged.

PR #319, opened 2018-01-12
add make_pretty_from_fits file to develop

jermainegug:
*creates norm variable
*checks wcs (**issue)
*changed cmap from cubealpha_r to inferno

`try:
    wcs = WCS(fname)
except:
    wcs = False`
This does not work as expected.
If we do:

`wcs = WCS(fname)
print(wcs)`

We get this as a result:

```WCS Keywords
Number of WCS axes: 2
CTYPE : ''  ''  
CRVAL : 0.0  0.0  
CRPIX : 0.0  0.0  
PC1_1 PC1_2  : 1.0  0.0  
PC2_1 PC2_2  : 0.0  1.0  
CDELT : 1.0  1.0  
NAXIS : 3352  2532

### CODE REVIEW

**Comment by wtgee** [comment]
Thanks @jermainegug. So when I told you to develop it I had you work with a separate script because you didn't have POCS installed yet. But know that you do...

What we want to do is take your `_make_pretty_from_fits` and repalce what is in: https://github.com/panoptes/POCS/blob/develop/pocs/utils/images/__init__.py#L87. Then you can just go ahead and delete this file.

**Comment by wtgee** [comment]
You can just keep working in the same branch and any new changes you push will automatically become part of this PR.

**Comment by AnthonyHorton** [comment]
Ah, OK, going to need to try something else to check for a valid WCS. Seems `WCS(fname)` will just create a mostly empty template in the absence of a full WCS in the FITS headers. Check the `astropy.wcs` documentation, there must be a function for checking whether a WCS is present/valid or not.

Also, for the plot using WCS use RA & dec for the axes (not an overlay) and don't bother with Galactic coordinates, at least not by default.  Could make the choice of coordinate systems an option, of course.

**Comment by wtgee** [comment]
Just about to put the comment in review for WCS. We do this elsewhere.  Basically:

```
w = wcs.WCS(filename)
assert w.is_celestial
```

File: pocs/utils/images/__init__.py [hunk 1]
```diff
@@ -92,7 +95,39 @@ def _make_pretty_from_fits(fname, **kwargs):
...
 
     data = getdata(fname)
-    plt.imshow(data, cmap='cubehelix_r', origin='lower')
+
+    norm = ImageNormalize(interval=MinMaxInterval(), stretch=LogStretch())
+
+    wcs = False
+
+    """
+    try:
```
wtgee:
```
try:
    w = wcs.WCS(filename)
    assert w.is_celestial
except AssertionError:
    pass
```

wtgee: Or just skip the `try` and use `if wcs.is_celstial` below.

pocs/utils/images/__init__.py
```diff
@@ -6,7 +6,10 @@
 from matplotlib import pyplot as plt
 from warnings import warn
 
+from astropy.io import fits
```
wtgee: I don't think you are using this.

jermainegug: Nope, using that in my jupyter notebook.

pocs/utils/images/__init__.py
jermainegug: Get this error message `AttributeError: 'bool' object has no attribute 'WCS'` for `w = wcs.WCS(filename)`.

AnthonyHorton: @jermainegug That's because in your code you're setting `wcs` to `False`, whereas @wtgee's code suggestion is assuming that `wcs` is `astropy.wcs` (i.e. you're doing `from astropy import wcs`)

wtgee: Yes, I didn't adjust it for your code, I just copied what we have elsewhere.

You import `WCS` directly so use that instead of `wcs.WCS`.

But, you are also mixing types. You create `wcs = False` (a boolean) up above and then assign `wcs=WCS(filename)` (a `WCS` class) to it. I would just skip the `try` and use the `is_celestial` in the `if` statement directly after getting the wcs.

jermainegug: What is `wcs` in this case? `wcs=WCS(fname)`?

AnthonyHorton: R.T.D.

It's an `WCS` object, i.e. an instance of the `astropy.wcs.WCS` class.

http://docs.astropy.org/en/stable/api/astropy.wcs.WCS.html#astropy.wcs.WCS

AnthonyHorton:
@jermainegug Have a look through the documentation on this under the `astropy.visualization` module: http://docs.astropy.org/en/stable/visualization/wcsaxes/index.html

For this you should be using `matplotlib`s object orientated interface, rather than the `pyplot` interface, see the second example in the above link.

That will then make it easier to properly format the tick labels for axis, see: http://docs.astropy.org/en/stable/visualization/wcsaxes/ticks_labels_grid.html  You're going to want to do this in order to make the Right Ascension axis use hour units instead of the default degrees.

Also, you need to make the axis labels more explicit. I'd prefer to spell out Right Ascension & declination (if there's room) but more importantly you should specify the coordinate system and (if applicable) equinox that are being used. The WCS object should have this information (and also allow you to transform to others).  It's probably either FK5 J2000 or ICRS.

pocs/utils/images/__init__.py
```diff
@@ -89,7 +90,7 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
 def _make_pretty_from_fits(fname, **kwargs):
     config = load_config()
 
-    title = '{} {}'.format(kwargs.get('title', ''), current_time().isot)
+    title = '{} {} {}'.format(kwargs.get('title', ''), current_time().isot, 'WCS') # is WCS the coordinate system?
```
wtgee: You can do `current_time(pretty=True)` for a better time output here.

wtgee: And WCS = "world coordinate system", but you don't just want the string there unless you are only putting it there when `is_celestial` is true or something.

pocs/utils/images/__init__.py
```diff
@@ -103,14 +104,22 @@ def _make_pretty_from_fits(fname, **kwargs):
         ax = plt.subplot(projection=wcs)
 
         ax.coords.grid(True, color='white', ls='solid')
-        ax.set_xlabel('RA')
-        ax.set_ylabel('Dec')
+
+        ra = ax.coords[0]
```
wtgee: This is not really an `ra`, it is an `ra_axis`. Same with the dec. Better to be explicit about the name because if you had another variable (which maybe you will in future) that actually holds an RA value you would get confused by this.

pocs/utils/images/__init__.py
```diff
@@ -103,14 +104,22 @@ def _make_pretty_from_fits(fname, **kwargs):
...
+        dec = ax.coords[1]
+
+        ra.set_axislabel('Right Ascension')
```
wtgee: I tend to do `label [units]` (although I think @AnthonyHorton does something different), in which case this would be `Right Ascension [hms]` and `Declination [dms]`.

pocs/utils/images/__init__.py
```diff
@@ -103,14 +104,22 @@ def _make_pretty_from_fits(fname, **kwargs):
...
+        ra.set_major_formatter('hh:mm:ss')
+        dec.set_major_formatter('dd:mm:ss')
     else:
         ax = plt.subplot()
 
         ax.grid(True, color='white', ls='solid')
-        ax.set_xlabel('Pixel x-axis')
-        ax.set_ylabel('Pixel y-axis')
+
+        ax.coords[0].set_axislabel('Pixel x-axis')
```
wtgee: As per comments above, I would just do `X [pixels]` and `Y [pixels]`.

pocs/utils/images/__init__.py
```diff
@@ -103,14 +104,22 @@ def _make_pretty_from_fits(fname, **kwargs):
         ax = plt.subplot(projection=wcs)
 
         ax.coords.grid(True, color='white', ls='solid')
```
wtgee: This line could be moved below `if` clause so it is not repeated.

pocs/utils/images/__init__.py
AnthonyHorton: Yes, I tend to prefer the `label / units` convention (i.e. the numbers are the quantity in question divided by the given unit).  Either is fine, though.

In this particular case putting units in the axis labels should be redundant though, the tick label formats include units, see https://github.com/AstroHuntsman/HuntsmanPOCSLegacy/blob/develop/examples/notebooks/dice9.png for an example.

pocs/utils/images/__init__.py
AnthonyHorton: Or `X / pixels`, `Y / pixels` ;-)

pocs/utils/images/__init__.py
AnthonyHorton: Even then it will be obvious from the rest of the plot whether a WCS from the FITS image has been used or not, from the axis labels, tick labels and grid. Don't need to put it in the title too.

pocs/utils/images/__init__.py
```diff
@@ -87,12 +90,39 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
...
+        ax.coords.grid(True, color='white', ls='solid')
+
+        ra = ax.coords[0]
+        dec = ax.coords[1]
+
+        ra.set_axislabel('Right Ascension')
+        dec.set_axislabel('Declination')
+
+        ra.set_major_formatter('hh:mm:ss')
+        dec.set_major_formatter('dd:mm:ss')
```
AnthonyHorton: For Huntsman, PANOPTES or in fact most other optical instruments the field of view will be large enough that including seconds in the major tick labels is unnecessary. `hh:mm` and `dd:mm` is probably better.

pocs/utils/images/__init__.py
jermainegug: So we will just be able to tell the coordinate system based off what the image looks like?

pocs/utils/images/__init__.py
jermainegug: I have written them in for now. Will make sure that the units are in the axis label when I run the code with a fits file that has wcs., then I will get rid of the units in the label.

pocs/utils/images/__init__.py
```diff
@@ -87,12 +90,39 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
...
 
-    title = '{} {}'.format(kwargs.get('title', ''), current_time().isot)
+    title = '{} {} {}'.format(kwargs.get('title', ''), current_time().isot, 'WCS') # is WCS the coordinate system?
 
     new_filename = fname.replace('.fits', '.jpg')
 
     data = getdata(fname)
-    plt.imshow(data, cmap='cubehelix_r', origin='lower')
+
+    norm = ImageNormalize(interval=MinMaxInterval(), stretch=LogStretch())
```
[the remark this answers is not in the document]
jermainegug: What sort of percentage are we thinking? Something like 0.1%?

pocs/utils/images/__init__.py
```diff
@@ -93,33 +93,30 @@ def _make_pretty_from_fits(fname, **kwargs):
...
-
     data = getdata(fname)
 
-    norm = ImageNormalize(interval=MinMaxInterval(), stretch=LogStretch())
+    percent_value = 99.81 #(?) | Why this % value? --> Gave a very nice image.
 
+    norm = ImageNormalize(interval=PercentileInterval(percent_value),
+                          stretch=LogStretch())
     wcs = WCS(fname)
+    ax = plt.subplot(projection=wcs)
```
wtgee: This line might need to stay inside the `if`. I'm not sure what happens if you pass a `projection` but don't have a valid WCS. Have you tried on the original images that weren't solved yet?

wtgee:
This has nothing to do with your code as it was already in the function, but we shouldn't actually be doing the symlink stuff here since we don't do it in `_make_pretty_from_cr2`.  Can you remove all those lines since you are updating the function anyway? Thanks!

pocs/utils/images/__init__.py
```diff
@@ -87,12 +90,34 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
 def _make_pretty_from_fits(fname, **kwargs):
     config = load_config()
 
-    title = '{} {}'.format(kwargs.get('title', ''), current_time().isot)
+    title = '{} {}'.format(kwargs.get('title', ''), current_time(pretty=True).isot)
```
wtgee: Tests are failing on this. Doing the `pretty=True` turns it into a string, so then you can't call `.isot` on it. You want one or the other. They are basically the same except `pretty` has spaces for easier reading.

pocs/utils/images/__init__.py
wtgee: Just for fun I just ran it on an image that a PANOPTES unit just took that has a pretty empty star field and a meteor streak. Looks like it handles the lack of `wcs` fine.

test

wtgee: There's a lot of dead white space in there around the edges. I think you can do a [`fig.tight_layout()`](https://matplotlib.org/users/tight_layout_guide.html) to get rid of some of that.

jermainegug: I tested it on my Jupyter notebook and it seems to work just fine. If I don't assign anything to `projection` then I can't use `coords[]` for the pixel coordinates. I would have to use this for `else`
```ax = plt.subplot()
ax.set_xlabel('X / pixels')
ax.set_xlabel('X / pixels')```
But I think what I have should work just fine :) 
If not then we can change it to this.

Awesome, image looks good, is this with `percent_value = 99.81`? Only asking because maybe we should make that value able to be modified since when I was trying different percentages, and even changing it by 0.1% it changes quite drastically.

**Comment by wtgee** [comment]
Yes, that image is from using your branch directly (I removed the isot
thing to make it work). So yes, using that same percentage. It's to some
extent going to depend on each image but we probably want to  find
something that works most of the time and then pass an option via kwargs.

File: pocs/utils/images/__init__.py (hunk 10 above)
Comment (jermainegug) [reply pocs/utils/images/__init__.py]: And I will add the `fig.tight_layout()` soon and let you know how it looks.

**Comment by jermainegug** [comment]
horsehead_inferno_new_method

I used the example image and the wcs is working great! Like @AnthonyHorton mentioned, we wont need the units in the axis label since they are written to values on the image.

**Comment by wtgee** [comment]
Note that you are failing on some codestyle issues. See https://travis-ci.org/panoptes/POCS/builds/328330892#L4924

With atom or sublime you can use an autopep8 tool that should correct most things.

**Comment by jermainegug** [comment]
autopep8 doesn't change anything in my file..

**Comment by wtgee** [comment]
> autopep8 doesn't change anything in my file..

They were related to the comments you deleted anyway.

Can you post a pic with the tight_layout? Also, in pick above I see the title but not a date. Seems like there should be both from this code.

**Comment by jermainegug** [comment]
fornax_inferno_best_image

The code in my Jupyter notebook doesn't print the time, I just manually put in the title for the images I get. But here is the picture with the tight_layout.

**Comment by wtgee** [comment]
> I just manually put in the title for the images I get. But here is the picture with the tight_layout.

You should be able to read both the `FIELD` and the `OBSTIME` from the fits headers. Something like:

```
from astropy.io import fits

data = fits.getdata(fn)
header = fits.getheader(fn)

title = kwargs.get('title', header.get('FIELD', ''))
```

**Comment by jermainegug** [comment]
`header.get('FIELD', ' ')` gives me Fornax and 'OBSTIME doesn't seem to have anything.

**Comment by jermainegug** [comment]
Got it, was `DATE-OBS`, I will change it so it replaces the T with a space.

**Comment by jermainegug** [comment]
fornax_inferno_best_image

**Comment by jermainegug** [comment]
When the image is captured,does this run straight away? If it doesn't then does `current_time(pretty=True)` give us the time when the image was taken or just when we call on the method?

**Comment by wtgee** [comment]
> When the image is captured,does this run straight away? If it doesn't then does current_time(pretty=True) give us the time when the image was taken or just when we call on the method?

If it is pulling the `OBS-TIME` then that is recorded when image is taken.

**Comment by jermainegug** [comment]
Did we want to use any of that header stuff in our actual code?

**Comment by jermainegug** [comment]
Check seems to give us this:
```
pocs/tests/utils/test_image_utils.py::test_clean_observation_dir
  /home/travis/build/panoptes/POCS/pocs/utils/images/__init__.py:140: UserWarning: Can't link latest image: [Errno 2] No such file or directory: '/tmp/tmppbxfbjvv/solved.jpg' -> '/var/panoptes/images/latest.jpg'
    warn("Can't link latest image: {}".format(e))
```

**Comment by wtgee** [comment]
See my comment above about removing all the link stuff: https://github.com/panoptes/POCS/pull/319#issuecomment-357238190

**Comment by jermainegug** [comment]
Oh yes, I forgot to delete those. Thanks.

**Comment by wtgee** [comment]
> Oh yes, I forgot to delete those. Thanks.

There is other stuff that goes with it. For instance once you delete that I don't think you need the `config` or the `image_dir`. Make sure to do a full cleanup please. Thanks!

**Comment by wtgee** [comment]
Regarding the title and time from the fits headers, yes I would think so. The example I showed up lets a `title` in the kwargs override that so it would just be a backup.

**Comment by jermainegug** [comment]
Ignore the recent commit.

**Comment by wtgee** [comment]
You want to keep the kwargs. That lets someone pass something in to override. So if I am using `make_pretty_image` interactively from a jupyter console (which I do freqently) I can say `make_pretty_image(title="test_01")` and have it work. That should take precedence over the fits header because I am explicitly asking for it.

**Comment by wtgee** [comment]
> Ignore the recent commit.

Too late. :)

File: pocs/utils/images/__init__.py [hunk 12]
```diff
@@ -85,30 +89,47 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
...
+                           date_time)
 
+    new_filename = fname.replace('.fits', '.jpg')
     data = getdata(fname)
-    plt.imshow(data, cmap='cubehelix_r', origin='lower')
-    plt.title(title)
-    plt.savefig(new_filename)
 
-    image_dir = config['directories']['images']
+    percent_value = 99.9
```
wtgee: Just to make it flexible for the future I would pull this from kwargs too. So `kwargs.get('normalize_clip_percent', 99.9)`

pocs/utils/images/__init__.py
```diff
@@ -85,30 +89,47 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
...
+        dec_axis.set_axislabel('Declination')
+        dec_axis.set_major_formatter('dd:mm')
+    else:
+        ax = plt.subplot()
+        ax.grid(True, color='white', ls='-', alpha=0.3)
+
+        ax.set_xlabel('X / pixels')
+        ax.set_ylabel('Y / pixels')
+
+    ax.imshow(data, norm=norm, cmap='inferno', origin='lower')
```
wtgee: Also make this slightly more flexible by doing a `cmap=kwargs.get('cmap', 'inferno')`. Then you can play with different colormaps in runtime without touching code.

jermainegug:
Yay! All checks passed :)

pocs/utils/images/__init__.py
```diff
@@ -85,30 +88,45 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
 
 
 def _make_pretty_from_fits(fname, **kwargs):
-    config = load_config()
+    header = getheader(fname)
+    data = getdata(fname)
 
-    title = '{} {}'.format(kwargs.get('title', ''), current_time().isot)
+    title = kwargs.get('title', header.get('FIELD', ''))
+    date_time = header.get('DATE-OBS', current_time()).replace('T', ' ', 1)
```
wtgee: This might need to be tested. `current_time` will return a time object and then the `replace` will probably fail. If you leave at `current_time(pretty=True)` it will return a string and the replace will work on a string (it just won't replace anything since the `T` has already been removed, but that's fine), which is probably safer and means you don't need to bother testing that.

jermainegug: I have changed it now but I will test anyways (want to make sure that it doesn't get rid of the T in UTC).

jermainegug: Oh, current_time doesn't print out UTC. So it is all good then.

jermainegug:
Indeed we have!

No problem but I do not have write access so the merge button does not show up.

Not sure if you can do this but did you want me to say it also fixes #43 along with #232 ?

wtgee:
Ok, I'll go ahead and merge. #43 stays open as it is more of a question of what we want to do with the pretty pictures than it is about getting pretty pictures working.

Issue #284, opened 2017-12-31
Improve Arduino enumeration


jamessynge:
In the review of PR #280, Wilfred suggested using PySerial to enumerate the set of ports:

> We can also grab connected ports and just check those: https://pythonhosted.org/pyserial/tools.html?highlight=comport#serial.tools.list_ports.comports. We could then check vendor id and product id. The arduino uno even reports a serial_number correctly although I don't think the micros do.
>
> That could probably help clean up some of the _auto_detect_port below.

For testing, I'd like to enable the use of serial handlers, which are probably not supported by list_ports.

wtgee:
> For testing, I'd like to enable the use of serial handlers, which are probably not supported by list_ports.

Not sure I follow on this comment. The auto detect feature wouldn't work with the tests anyway, right?

jamessynge:
Fixed with #320.

Issue #181, opened 2017-12-09
Support pyserial options in yaml files


jamessynge:
PySerial supports getting all properties as a dictionary, and replacing some or all properties with a dictionary. I'm proposing that we provide the ability to supply those properties via the yaml file. For example, instead of this (from `conf_files/pocs.yaml`):

```
mount:
    brand: ioptron
    model: 30
    driver: ioptron
    port: /dev/ttyUSB0
    non_sidereal_available: True
```

We'd instead have:
```
mount:
    brand: ioptron
    model: 30
    driver: ioptron
    non_sidereal_available: True
    serial:
       port: /dev/ttyUSB0
       baudrate: 9600
       timeout: 2.0
```

Besides adding the ability to set all of the PySerial properties, this has the added benefit of addressing the pylint complaint that the SerialData constructor takes too many arguments.

wtgee:
Yes, I like this more.

jamessynge:
I note that Observatory.initialize is supposed to be responsible for setting up the communication with devices. I'm thinking we can extend the serial section with info about what to do when the rs232 device is created. For example:
```yaml
mount:
    brand: ioptron
    model: 30
    driver: ioptron
    non_sidereal_available: True
    serial:
       port: /dev/ttyUSB0
       baudrate: 9600
       timeout: 2.0
       validate_port: True   # If true, then the port must exist at create time.
       open_port: False      # If true, then opens port immediately.
```

`open_port: True` implies `validate_port: True` simply because the port must exist in order to be opened. And since Observatory.initialize is supposed to setup communication, I'm proposing that `open_port` be `False` by default, but that `validate_port` be `True` by default.

I propose to create method `rs232.create_serial_data_from_config(serial_config, name)`, where name is something like 'mount' or 'ioptron', determined by the caller. The serial_config param is just the contents of the serial section in the config. This `rs232.create_serial_data_from_config` will be called by `dome.create_dome_from_config`, before creating the Dome instance.

@wtgee Does this make sense to you?

wtgee:
Yes, sounds good and makes sense, but is seems like the caller would already be inside the `mount` or `ioptron` so the `name` param would be redundant. It seems like `create_serial_data_from_config` shouldn't really care what it is creating a connection for, it should just return and create connection. The caller should be responsible for passing it the relevant portion of the config. Unless you had something else in mind that happens inside `create_serial_data_from_config`. 

[...]

PR #320, opened 2018-01-12
Add a simple (and incomplete) PySerial handler that emulates the output

jamessynge:
behavior of our Arduino sketches.

### DIFF
--- a/peas/sensors.py
+++ b/peas/sensors.py
@@ -226,14 +226,14 @@ def auto_detect_port(port, logger):
     """
     logger.debug('Attempting to connect to serial port: {}'.format(port))
     try:
-        serial_reader = SerialData(port=port)
+        serial_reader = SerialData(port=port, baudrate=9600)
         serial_reader.connect()
         logger.debug('Connected to {}', port)
     except Exception as e:
         logger.warning('Could not connect to port: {}'.format(port))
         return None
     try:
-        reading = get_and_parse_reading(serial_reader, logger=logger)
+        reading = serial_reader.get_and_parse_reading()
         if not reading:
             return None
         (ts, data) = reading
@@ -248,49 +248,6 @@ def auto_detect_port(port, logger):
     finally:
         if serial_reader:
             serial_reader.disconnect()
-
-
-
-def get_and_parse_reading(serial_reader, logger=None, num_tries=5):
-    """Try several times to get a parseable reading, which is then returned.
-
-    Usually when we open a serial port connected to one of the PANOPTES arduino sketches we find
-    a partial line of output when we first start reading, which then fails to be parsed. This
-    function reads several times until we get a full reading.
-
-    Returns: (timestamp, parsed_reading), where timestamp is a datetime.datetime and parsed_reading
-        is the parsed version of the reading (i.e. the decoded JSON data).
-    """
-    if not logger:
-        logger = serial_reader.logger
-    for try_num in range(num_tries):
-        try:
-            logger.debug('Reading from {} (try_num={})', serial_reader.name, try_num)
-            raw_reading = serial_reader.get_reading()
-        except Exception as e:
-            logger.warning('get_reading failed: %r', e)
-            continue
-        reading = parse_reading(raw_reading, logger)
-        if reading:
-            logger.debug('Parsed reading: {}', reading)
-            return reading
-        # Don't emit a warning for the first reading.
-        if try_num > 0:
-            logger.warning('On try_num {}, JSON parsing failed for {!r}', try_num, raw_reading)
-    return None
-
-
-def parse_reading(reading, logger):
-    if not reading:
-        return None
-    ts, data = reading
-    try:
-        parsed = json.loads(data)
-        return (ts, parsed)
-    except Exception as e:
-        logger.debug('Exception while parsing JSON: %r', e)
-        logger.debug('Erroneous JSON: %r', data)
-        return None
 
 
 # Support testing by just listing the available devices.
--- a/peas/sensors.py
+++ b/peas/sensors.py
@@ -1,4 +1,7 @@
+import json
 import os
+from serial.tools.list_ports import comports as list_comports
+import sys
 import yaml
 
 from pocs.utils.database import PanMongo
@@ -6,6 +9,7 @@ from pocs.utils.config import load_config
 from pocs.utils.logger import get_root_logger
 from pocs.utils.messaging import PanMessaging
 from pocs.utils.rs232 import SerialData
+
 
 
 class ArduinoSerialMonitor(object):
@@ -69,21 +73,29 @@ class ArduinoSerialMonitor(object):
         else:
             self.logger.debug('Port {} not found', port)
             return None
+        try:
+            serial_reader = self._connect_serial(port)
+            if serial_reader is None:
+                return None
+        except Exception:
+            self.logger.debug('Failed to connect on {}', port)
+            return None
+        # Try to get a reading with a full line. Sometimes the first line is
+        # partial, so we need to read a second line to find a line with the
+        # name of the board in the reading.
         num_tries = 5
         for _ in range(num_tries):
-            self.logger.debug('Trying to connect on {}', port)
-            serial_reader = None
             try:
-                serial_reader = self._connect_serial(port)
-                self.logger.debug('Parsing the reading from {}', port)
-                data = serial_reader.get_reading()
-                parsed = json.loads(data[1])
-                if 'name' in parsed:
+                self.logger.debug('Reading from {}', port)
+                data = self._get_parsed_reading(serial_reader)
+                if data and len(data) == 2 and 'name' in data[1]:
+                    parsed = data[1]
                     sensor_name = parsed['name']
                     self.logger.debug('Found name {}', sensor_name)
                     result = (sensor_name, serial_reader)
                     serial_reader = None
                     return result
+                self.logger.debug("'name' not found in reading")
             except Exception:
                 pass
             finally:
@@ -92,16 +104,30 @@ class ArduinoSerialMonitor(object):
         return None
 
     def _connect_serial(self, port):
-        if port is not None:
-            self.logger.debug('Attempting to connect to serial port: {}'.format(port))
-            serial_reader = SerialData(port=port)
-            self.logger.debug(serial_reader)
-            try:
-                serial_reader.connect()
-                self.logger.debug('Connected to {}', port)
-            except Exception as e:
-                self.logger.warning('Could not connect to port: {}'.format(port))
+        self.logger.debug('Attempting to connect to serial port: {}'.format(port))
+        serial_reader = SerialData(port=port)
+        try:
+            serial_reader.connect()
+            self.logger.debug('Connected to {}', port)
             return serial_reader
+        except Exception as e:
+            self.logger.warning('Could not connect to port: {}'.format(port))
+            return None
+
+    def _get_parsed_reading(self, serial_reader):
+        try:
+            data = serial_reader.get_reading()
+        except Exception as e:
+            self.logger.debug('get_reading failed: %r', e)
+            return None
+        try:
+            self.logger.debug('Parsing the reading from {}', serial_reader.name)
+            parsed = json.loads(data[1])
+            return (data[0], parsed)
+        except Exception as e:
+            self.logger.debug('Failed to parse reading: %r', e)
+            self.logger.debug('Erroneous reading: %r', data)
+            return None
 
     def disconnect(self):
         for sensor_name, reader_info in self.serial_readers.items():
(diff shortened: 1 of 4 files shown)


jamessynge:
Note that this is not yet used. I tested it using PySerial's miniterm.py.

This is a request for a code review, but it isn't necessary to merge until there is a test that uses this.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+        cmd = q.get_nowait()
+    return cmd  # Present just for debugging.
+
+
+class ArduinoSimulator:
+    """Simulates the serial behavior of the PANOPTES Arduino sketches.
+
+    The RS-232 connection is simulated with an input and output queue of bytes. This class provides
+    a run function which can be called from a Thread to execute. Every two seconds while running it
+    will generate another json output line,
```
wtgee: Something weird going on with line breaks in this comment.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+    at around 1000 bytes per second (near 9600 buad). The bytes  (actually chunked up in batches, so not emitting a single byte
```
wtgee: `buad` -> `baud`. And the last sentence just needs some correction.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+                "millis":1507497677,
+                "report_num":7835,
+                "inputs":6,
+                "camera_00":1,
+                "camera_01":1,
+                "accelerometer": {"x":-7.02, "y":6.95, "z":1.70, "o": 6},
+                "humidity":59.60, "temp_00":12.50
+            }
+            """)
+        # TODO(jamessynge): Update telemetry_board message to reflect latest changes.
```
wtgee: We should update telemetry, camera, and power board to have consistent interface. For instance in telemetry the power is all under the `power` key but in camera it is just `camera_00`, which doesn't really explain what it is. Both of them need a version as well.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+        self.telemetry_board = json.loads("""
+            {
+                "name":"telemetry_board",
+                "millis":155581294,
+                "report_num":4801,
+                "ver":"2017-09-23",
+                "power": {"computer":1, "fan":1, "mount":1, "cameras":1, "weather":1, "main":1},
+                "current": {"main":387,"fan":28,"mount":34,"cameras":27},
+                "amps": {"main":1083.60,"fan":50.40,"mount":61.20,"cameras":27.00},
```
wtgee: I realize you are mocking what is there now, but I thought we wanted to drop the amps from the arduino code.

I guess I say that because the worry here, and with comment above, is that this test just gets out of line with what we are actually testing for. Not sure there is any real way around that.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+
+    def read_relay_queue_until(self, next_time):
+        """Wait for a little bit for relay input, not more than the time until the next action."""
+        now = datetime.datetime.now()
+        if now >= next_time:
+            # Already reached the time for the next main loop event,
+            # so return to repeat the main loop.
+            return
+        remaining = (next_time - now).total_seconds()
+        assert remaining > 0
```
wtgee: It doesn't look like you are catching this assert anywhere which means if it every happens you will get a runtime error.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+        self.logger.debug('output_next_chunk -> {}', chunk)
+
+    def generate_next_message(self, now):
+        """Append the next message to the pending bytearray and scheduled the next message."""
+        b = self.generate_next_message_bytes(now)
+        self.pending_json_bytes.extend(b)
+        self.next_message_time = datetime.datetime.now() + self.message_delta
+
+    def generate_next_message_bytes(self, now):
+        """Generate the next message (report) from the simulated Arduino."""
```
wtgee: Not necessary for now but might be nice to change other values besides `millis` and `report_num`.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+                self.parity == serialutil.PARITY_NONE and not self.rtscts and not self.dsrdtr)
+
+    def _read1(self, timeout_obj):
+        """Read 1 byte of input, of type bytes."""
+        if not self.is_open:
+            raise serialutil.portNotOpenError
+        if not self.json_bytes:
+            try:
+                entry = self.json_queue.get(block=True, timeout=timeout_obj.time_left())
+                assert isinstance(entry, bytes)
```
wtgee: I would catch this assertion also.

peas/tests/serial_handlers/protocol_arduino_simulator.py
```diff
@@ -0,0 +1,422 @@
...
+            _drain_queue(self.relay_queue)
+            _drain_queue(self.json_queue)
+            self.json_bytes.clear()
+            self.stop.clear()
+            self.simulator_thread = threading.Thread(
+                name='Device Simulator', target=lambda: self.device_simulator.run())
+            self.simulator_thread.start()
+        elif self.simulator_thread and not need_thread:
+            self.stop.set()
+            self.simulator_thread.join(timeout=30.0)
```
wtgee: Do we need this long of a timeout here?

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: Agreed, will file an issue.

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: But that makes it a failure in a test, not prod.

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: I figured that would be gilding the lily. And yes, I thought of it. 😁

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: Where?

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: It shouldn't get that long, unless you're running in the debugger. But, again, this is intended to be used from tests. I don't want to make it so short that on an underpowered or overloaded system (like Travis) we start to get flakey results.

peas/tests/serial_handlers/protocol_arduino_simulator.py
wtgee: Ok, shouldn't be an issue then.

peas/tests/serial_handlers/protocol_arduino_simulator.py
wtgee: Probably not necessary for the test I suppose. Same thing as above.

peas/tests/serial_handlers/protocol_arduino_simulator.py
wtgee: Sounds good.

jamessynge:
I'll get started on a test that uses it.

peas/sensors.py
```diff
@@ -167,3 +193,115 @@ def capture(self, use_mongo=True, send_message=True):
...
+
+
+def auto_detect_port(port, logger):
+    """Open a port and determine which type of board its producing output.
+
+    Returns: (name, serial_reader) if a recognizable device is connected, else None.
+    """
+    logger.debug('Attempting to connect to serial port: {}'.format(port))
+    try:
+        serial_reader = SerialData(port=port)
```
wtgee: If I add a `baudrate=9600` here then your script works. I think. I just have an Uno but nothing hooked up to it, so I get the json response but have `nan` for all values so it doesn't parse. But that is different than with no baudrate in which case I see the null arrays.

peas/sensors.py
```diff
@@ -60,52 +57,9 @@ def __init__(self, auto_detect=False, *args, **kwargs):
...
-                pass
-            finally:
-                if serial_reader:
-                    serial_reader.disconnect()
-        return None
-
     def _connect_serial(self, port):
         self.logger.debug('Attempting to connect to serial port: {}'.format(port))
-        serial_reader = SerialData(port=port)
+        serial_reader = SerialData(port=port, baudrate=9600)
```
wtgee: I'm guessing it's not worth making this a variable?

peas/sensors.py
```diff
@@ -151,46 +90,35 @@ def capture(self, use_mongo=True, send_message=True):
             sensor_data (dict):     Dictionary of sensors keyed by sensor name.
         """
 
+        # Read from all the readers; we send messages with sensor data immediately, but accumulate
+        # data from all sensors before storing in the db.
+        # Note that there is no guarantee that these are the LATEST reports emitted by the sensors,
+        # as the
```
wtgee: Comment trails off here. However, regarding the latest, we can clear the input buffer and then it should be the latest, right?

peas/sensors.py
```diff
@@ -151,46 +90,35 @@ def capture(self, use_mongo=True, send_message=True):
...
+        # as the 
         sensor_data = dict()
-
-        # Read from all the readers
         for sensor_name, reader_info in self.serial_readers.items():
             reader = reader_info['reader']
 
-            # Get the values
-            self.logger.debug('Reading next serial value')
+            # Get the values before attempting to re
```
wtgee: Seems like your comments are having troubles finishing themselves today. :)

peas/sensors.py
```diff
@@ -125,45 +90,103 @@ def capture(self, use_mongo=True, send_message=True):
...
+        logger = get_root_logger()
+    result = []
+    for port in comports:
+        v = auto_detect_port(port, logger)
+        if v:
+            result.append(v)
+    return result
+
+
+def find_arduino_devices():
```
wtgee: This probably makes the most sense here but just wanted to point out that we have a [`list_connected_cameras`](
https://github.com/panoptes/POCS/blob/develop/pocs/utils/__init__.py#L74) function in the utils. We should aim for some consistency of putting these kinds of functions either in the file like you have here or create a whole device discovery utility.

peas/sensors.py
```diff
@@ -125,45 +90,103 @@ def capture(self, use_mongo=True, send_message=True):
...
+        return None
+    except Exception as e:
+        logger.error('Exception while auto-detecting port {!r}: {!r}'.format(port, e))
+    finally:
+        if serial_reader:
+            serial_reader.disconnect()
+
+
+# Support testing by just listing the available devices.
+if __name__ == '__main__':
```
wtgee: FWIW, and regarding comment above about separate utils for discovering devices, I have mostly been making the module callable via `if __main__` _unless_ they are in the utils folder. However, that would make development a little annoying as you would have to run a util file to test the sensors file.  Still, could end up looking something like:
```
python pocs/utils/devices.py --list_arduinos
python pocs/utils/devices.py --list_cameras
python pocs/utils/devices.py --list_attached_hardware
```
etc.

jamessynge:
With the addition of the main handling code in sensors.py, I now see this on PAN001:
```
(panoptes-env) panoptes@Panmass:/var/panoptes/POCS (test-auto-detect)$ python peas/sensors.py
Arduino devices: /dev/ttyACM1, /dev/ttyACM0
Device /dev/ttyACM1 has name telemetry_board
Device /dev/ttyACM0 has name camera_board
(panoptes-env) panoptes@Panmass:/var/panoptes/POCS (test-auto-detect)$ 
```
I'm psyched!

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: Done.

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: Done.

peas/tests/serial_handlers/protocol_arduino_simulator.py
jamessynge: Agreed, I don't see a good way around that... besides an awesome arduino simulator, with mock sensors attached!

peas/sensors.py
jamessynge: We should probably address this along with #181 (i.e. even without the port being specified in the yaml file, we could have other serial params there, including baudrate).

peas/sensors.py
jamessynge: Clearing the buffer gets rid of all accumulated input, including the first few bytes of the next line (if it has started emitting them), at which point we'll end up throwing away the "current" reading. I think we can easily solve this problem by splitting the reading of these into two threads, one for each, which disentangles their timing.

peas/sensors.py
jamessynge: My thoughts are leaping all over the ...

peas/sensors.py
jamessynge: Good idea. I'll add a comment or issue to that effect.

jamessynge:
Note, this closes #284, but doesn't actually create a list_devices util.

peas/sensors.py
jamessynge: #339 entered.

peas/tests/serial_handlers/protocol_arduino_simulator.py
wtgee: I dream about building a whole unit and storing it in a lab at Subaru and then hooking the testing up via Jenkins so we can run real automated hardware tests. But that's a long way off.

jamessynge:
I'm going to merge this in now, even though I don't yet have use of the simulator. I'll work on that.

PR #323, opened 2018-01-12
Somehow missed these during the image split. 

wtgee:
Discovered on PAN001


jamessynge:
Does "some how missed these" really mean "they don't have test coverage"?

wtgee:
> Does "some how missed these" really mean "they don't have test coverage"?

Correct. That changes were mostly to the gphoto2 file itself, which is actually ignored completely by the coverage.

PR #327, opened 2018-01-12
Add some docs explaining mongo db and collection auto-creation

### LINKED ISSUE #324: Mongo db name incorrect
Things are going into a db named `db` rather than what is in the config.

wtgee:
Closes #324

### DIFF
--- a/pocs/utils/database.py
+++ b/pocs/utils/database.py
@@ -36,6 +36,17 @@ class PanMongo(object):
         starts and can be read and updated as the project is running. The server
         is a wrapper around a mongodb collection.
 
+        Note:
+            Because mongo can create collections at runtime, the pymongo module
+            will also lazily create both databases and collections based off of
+            attributes on the client. This means the attributes do not need to
+            exist on the client object beforehand and attributes assigned to the
+            object will automagically create a database and collection.
+
+            Because of this, we manually store a list of valid collections that
+            we want to access so that we do not get spuriously created collections
+            or databases.
+
         Args:
             db (str, optional): Name of the database containing the PANOPTES collections.
             host (str, optional): hostname running MongoDB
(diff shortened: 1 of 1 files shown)


pocs/utils/database.py
```diff
@@ -57,9 +69,10 @@ def __init__(self, db='panoptes', host='localhost', port=27017, connect=False):
             'weather',
         ]
 
-        db_handle = getattr(self._client, db)
+        # Create an attribute on the client with the db name.
+        db_handle = self._client[db]
 
-        # Setup static connections to the collections we want
+        # Setup static connections to the collections we want.
```
jamessynge: Does the `getattr(db_handle, collection)` create the collection? Or is that lazy too (i.e. maybe pymongo doesn't do anything until we try to read or write from the collection)?

No need to change this, but if the operation isn't lazy, and we wanted it to be, we could instead add a `__getattr__` method that validates that the attr is a member of `self.collections` before doing the setattr.

pocs/utils/database.py
```diff
@@ -46,6 +57,7 @@ def __init__(self, db='panoptes', host='localhost', port=27017, connect=False):
         # Get the mongo client
         self._client = get_shared_mongo_client(host, port, connect)
 
+        # Pre-defined list of collections that are valid.
```
jamessynge: Is self.collections intended for use by the caller?

pocs/utils/database.py
```diff
@@ -36,6 +36,17 @@ def __init__(self, db='panoptes', host='localhost', port=27017, connect=False):
         starts and can be read and updated as the project is running. The server
         is a wrapper around a mongodb collection.
 
+        Note:
```
jamessynge: Thank you!

pocs/utils/database.py
wtgee: I didn't think it needed to be private in case, but I suppose it could be.

pocs/utils/database.py
wtgee: It's lazy and happens upon first read/write.

wtgee: I'm just going to leave as is for now. Probably a few more things we could do to improve overall but nothing necessary right now.

PR #329, opened 2018-01-13
Only the primary camera should have images saved in list

wtgee:
The `expsoure_list` is used to get the `last_exposure`, which is then used for pointing comparisons. We only want to track the primary camera for this.

### DIFF
--- a/pocs/camera/canon_gphoto2.py
+++ b/pocs/camera/canon_gphoto2.py
@@ -143,7 +143,8 @@ class Camera(AbstractGPhotoCamera):
         proc = self.take_exposure(seconds=exp_time, filename=file_path)
 
         # Add most recent exposure to list
-        observation.exposure_list[image_id] = file_path.replace('.cr2', '.fits')
+        if self.is_primary:
+            observation.exposure_list[image_id] = file_path.replace('.cr2', '.fits')
 
         # Process the image after a set amount of time
         wait_time = exp_time + self.readout_time


pocs/camera/canon_gphoto2.py
```diff
@@ -143,7 +143,8 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
         proc = self.take_exposure(seconds=exp_time, filename=file_path)
 
         # Add most recent exposure to list
-        observation.exposure_list[image_id] = file_path.replace('.cr2', '.fits')
+        if self.is_primary:
+            observation.exposure_list[image_id] = file_path.replace('.cr2', '.fits')
```
jamessynge: Is the point that we're trying to have a single canonical image to represent each exposure, even if we have multiple cameras? I'm not sure I understand why. Perhaps a comment?

wtgee: The exposure list is used for tracking purposes so the closed-loop compares the pointing image to the `last_exposure`. We always want to compare just the primary camera to itself.

I was going to expand `exposure_list` so it contained all exposures, which would make the cleanup and uploading to cloud a bit easier. So this might change going forward. 

I'll add an explanation.

wtgee: I took the easy out and made #342 for now. I want to make the whole thing more robust.

PR #330, opened 2018-01-13
Clear existing offset info when parking

wtgee:
This is only an issue if the system temporarily parks because of bad weather then starts back up and selects the same target. Without this the mount will try to adjust tracking after taking a pointing picture.

Fix a bad log call

### DIFF
--- a/pocs/state/states/default/parking.py
+++ b/pocs/state/states/default/parking.py
@@ -4,13 +4,14 @@ def on_enter(event_data):
 
     # Clear any current observation
     pocs.observatory.current_observation = None
+    pocs.observatory.current_offset_info = None
 
     pocs.next_state = 'parked'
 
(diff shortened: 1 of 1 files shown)


pocs/state/states/default/parking.py
```diff
@@ -4,13 +4,14 @@ def on_enter(event_data):
...
     pocs.observatory.current_observation = None
+    pocs.observatory.current_offset_info = None
 
     pocs.next_state = 'parked'
 
     if pocs.observatory.has_dome:
         pocs.say('Closing dome')
         if not pocs.observatory.close_dome():
-            self.logger.critical('Unable to close dome!')
+            pocs.logger.critical('Unable to close dome!')
```
jamessynge: Doh! Thanks for catching that.

PR #332, opened 2018-01-13
Wait for tracking adjustment

wtgee:
The tracking adjustment for ioptron uses `move_ms_<direction>`, which places
the mount in `guide` mode (as indicated by the status) but is non-blocking.
If the tracking adjustment is too long (~ > 4-5 seconds) and no blocking
is done the machine will try to transition to next state. There is a
condition on the state machine that the mount must be in `tracking`
to `observe` so it gets stuck and then tries again. After 5 iterations
of same state without successful transition the machine will give up
and park the mount.

Here we just wait out for up to 30 seconds until back in `tracking`
mode.

Note: Ideally we should never have to have an adjustment of 4-5 seconds,
so not entirely sure what is going on there. Seems to happen most on
first adjustment after pointing but also depends on where mount is poining.
This is a problem with PAN001 that we have had and might not be an issue
on other units.

### DIFF
--- a/pocs/mount/mount.py
+++ b/pocs/mount/mount.py
@@ -504,7 +504,7 @@ class AbstractMount(PanBase):
             offset (astropy.units.Angle): Offset in arcseconds
 
         Returns:
-             astropy.units.Second: Offset in milliseconds at current speed
+             astropy.units.Quantity: Offset in milliseconds at current speed
         """
 
         rates = {
--- a/pocs/mount/mount.py
+++ b/pocs/mount/mount.py
@@ -501,10 +501,10 @@ class AbstractMount(PanBase):
         """ Get offset in milliseconds at current speed
 
         Args:
-            offset (float): Offset in arcseconds
+            offset (astropy.units.Angle): Offset in arcseconds
 
         Returns:
-            float: Offset in milliseconds at current speed
+             astropy.units.Second: Offset in milliseconds at current speed
         """
 
         rates = {
--- a/pocs/mount/simulator.py
+++ b/pocs/mount/simulator.py
@@ -100,7 +100,7 @@ class Mount(AbstractMount):
             offset (astropy.units.Angle): Offset in arcseconds
 
         Returns:
-             astropy.units.Second: Offset in milliseconds at current speed
+             astropy.units.Quantity: Offset in milliseconds at current speed
         """
 
         offset = 25 * u.arcsecond  # Fake value
--- a/pocs/mount/simulator.py
+++ b/pocs/mount/simulator.py
@@ -1,5 +1,7 @@
 import time
 
+from astropy import units as u
+
 from pocs.utils import current_time
 from pocs.mount import AbstractMount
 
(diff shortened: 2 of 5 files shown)


wtgee:
Figuring out why this dropped covered so much. DO NOT MERGE yet.

pocs/mount/mount.py
```diff
@@ -501,10 +501,10 @@ def get_ms_offset(self, offset, axis='ra'):
         """ Get offset in milliseconds at current speed
 
         Args:
-            offset (float): Offset in arcseconds
+            offset (astropy.units.Angle): Offset in arcseconds
```
jamessynge: Support for using type annotations. :-)

pocs/mount/simulator.py
```diff
@@ -91,6 +93,20 @@ def move_direction(self, direction='north', seconds=1.0):
...
         time.sleep(seconds)
 
+    def get_ms_offset(self, offset, axis='ra'):
+        """ Fake offset in milliseconds
+
+        Args:
+            offset (astropy.units.Angle): Offset in arcseconds
+
+        Returns:
+             astropy.units.Second: Offset in milliseconds at current speed
```
jamessynge: How does the type (Second) match with the comment saying it is milliseconds?

pocs/state/states/default/tracking.py
```diff
@@ -1,14 +1,31 @@
...
+
+
 def on_enter(event_data):
     """ The unit is tracking the target. Proceed to observations. """
     pocs = event_data.model
-    pocs.say("Checking our tracking")
+    pocs.next_state = 'parking'
+
+    # If we came from pointing then don't try to adjust
+    if event_data.transition.source != 'pointing':
```
jamessynge: I'm confused about pointing vs tracking vs observing w.r.t. moving / adjusting the mount.

pocs/state/states/default/tracking.py
```diff
@@ -1,14 +1,31 @@
...
+        pocs.say("Checking our tracking")
+        try:
+            ra_info, dec_info = pocs.observatory.update_tracking()
```
jamessynge: Seems like we should move more of this code into POCS or Observatory. i.e. these on_enter functions should have little logic of their own. Thoughts?

pocs/mount/mount.py
wtgee: Actually the astropy units was one of the things I was concerned about. I'm hoping it works seamlessly for them.

pocs/mount/simulator.py
wtgee: Actually this should be a`Quantity` anyway. I'll get it changed.

pocs/state/states/default/tracking.py
wtgee: In what sense? For what the mount reports or for the actual state? Maybe some documentation of the states will eventually help with this.

pocs/state/states/default/tracking.py
wtgee: Yes, good point. Doing that now.

wtgee: I've done this but removed the `say` output that says what correction was. It's been useful for me the last few days to watch it but it's also a little too much information for the `say` box. Ideally we should be plotting this in real-time within PAWS. The info is delivered via the `status` update so information is there.

This was mostly because I've been keeping any `say` calls exclusively in the states (except for the setup in `core.py`). Not sure if that is necessary or not.

PR #333, opened 2018-01-14
Fix open_dome command so that it doesn't require that it be dark,

jamessynge:
just that the weather be safe.

Re-write loop in AstrohavenDome._full_move so that it doesn't wait long
when reading from the dome. The old code was too slow, so the dome moved
in a little then stopped, then repeated that, but not fast enough for
closing to occur.

Fix a flakiness problem: once a limit has been reached (i.e. one shutter
all the way opened or all the way closed), the PLC will likely output
more than one of the corresponding limit characters before then
outputting the new stable character. So added code for reading the
state until it has reached a stable value.

### DIFF
--- a/pocs/dome/astrohaven.py
+++ b/pocs/dome/astrohaven.py
@@ -68,7 +68,11 @@ class AstrohavenDome(abstract_serial_dome.AbstractSerialDome):
     def open(self):
         self._full_move(Protocol.OPEN_A, Protocol.A_OPEN_LIMIT)
         self._full_move(Protocol.OPEN_B, Protocol.B_OPEN_LIMIT)
-        return self.is_open
+        v = self._read_state_until_stable()
+        if v == Protocol.BOTH_OPEN:
+            return True
+        self.logger.warning('AstrohavenDome.open wrong final state: {!r}', v)
+        return False
 
     @property
     def is_closed(self):
@@ -78,7 +82,11 @@ class AstrohavenDome(abstract_serial_dome.AbstractSerialDome):
     def close(self):
         self._full_move(Protocol.CLOSE_A, Protocol.A_CLOSE_LIMIT)
         self._full_move(Protocol.CLOSE_B, Protocol.B_CLOSE_LIMIT)
-        return self.is_closed
+        v = self._read_state_until_stable()
+        if v == Protocol.BOTH_CLOSED:
+            return True
+        self.logger.warning('AstrohavenDome.close wrong final state: {!r}', v)
+        return False
 
     @property
     def status(self):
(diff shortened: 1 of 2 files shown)


pocs/dome/astrohaven.py
```diff
@@ -68,7 +68,11 @@ def is_open(self):
     def open(self):
         self._full_move(Protocol.OPEN_A, Protocol.A_OPEN_LIMIT)
         self._full_move(Protocol.OPEN_B, Protocol.B_OPEN_LIMIT)
-        return self.is_open
+        v = self._read_state_until_stable()
```
wtgee: Not quite sure if the call to open A is blocking or not but since `A_IS_OPEN` is one of the `stable` states it seems like this could open A but not B and still return `True`. Is that intended?

pocs/dome/astrohaven.py
```diff
@@ -148,14 +153,32 @@ def _full_move(self, send, target_feedback):
...
-            if time.time() < end_by:
-                continue
-            self.logger.error('Timed out moving the dome. Check for hardware or communications ' +
-                              'problem. send=%r latest_state=%r', send, self._read_latest_state())
-            return False
-        return True
+        # Set a short timeout on reading, so that we don't open or close slowly.
+        # In other words, we'll try to read status, but if it isn't available,
+        # we'll just send another command.
+        saved_timeout = self.serial.ser.timeout
```
wtgee: I wish `pyserial` supported timeouts on individual `read` and `write`.

pocs/dome/astrohaven.py
```diff
@@ -148,14 +153,32 @@ def _full_move(self, send, target_feedback):
...
+                        return True
+                    if c != send:
+                        self.logger.debug('Unexpected value from dome: {!r}', data)
+                if time.time() < end_by:
+                    continue
+                self.logger.error(
+                    'Timed out moving the dome. Check for hardware or communications ' +
+                    'problem. send={!r} expected={!r} actual={!r}', send, target_feedback, data)
+                return False
+        finally:
```
wtgee: A `try` without an `except` but with a `finally` seems a little odd to me. I'm surprised that doesn't complain.

pocs/dome/astrohaven.py
jamessynge: Yes, intended in this case: _full_move is "blocking" (i.e. it doesn't return until it accomplishes its goal or times out), so ideally it fully opens A, then fully opens B.

I'm thinking though of changing full_move so that it sends at a faster rate, and only stops after it gets several of the limit responses, as I noticed that the dome this morning was very slightly open (a little light leaking in). I'll do that in a separate PR though.

pocs/dome/astrohaven.py
jamessynge: Ditto. I'll file an issue with pyserial.

jamessynge: Filed https://github.com/pyserial/pyserial/issues/304

pocs/dome/astrohaven.py
jamessynge: finally also executes if a return is executed. See pep-341

wtgee: I forgot about that. That's potentially even more confusing. I see that this was merged already but maybe a comment explaining that.

PR #334, opened 2018-01-14
Fix timeout interval for pointing

wtgee:
Also move the log message inside if state. Really the whole for loop
is redundant, could just get the primary camera directly. However if
we ever want to take pointing images with both cameras it's nice to
have this.

### DIFF
--- a/pocs/state/states/default/pointing.py
+++ b/pocs/state/states/default/pointing.py
@@ -59,7 +59,7 @@ def on_enter(event_data):
                 pocs.status()
 
                 if wait_time > timeout:
-                    raise error.Timeout
+                    raise error.Timeout("Timeout waiting for pointing image")
 
                 sleep(wait_interval)
                 wait_time += wait_interval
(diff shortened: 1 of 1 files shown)


pocs/state/states/default/pointing.py
```diff
@@ -59,8 +62,8 @@ def on_enter(event_data):
                     'Waiting for images: {} seconds'.format(wait_time))
                 pocs.status()
 
-                if wait_interval > timeout:
-                    raise error.Timeout
+                if wait_time > timeout:
```
jamessynge: I've been writing a bunch of similar timeout code. I think we probably need a utility so that we get this right all the time. PySerial contains a timeout class.

wtgee: Yes, I was going to mention the exact same thing. This particular code isn't dealing with pyserial though and it's already a separate process. Could be made more generic and useful for sure.

PR #337, opened 2018-01-14
Fixing EQUINOX header (hopefully for the last time)

wtgee:
* `Observation.status` will return string, e.g. `J2000`.
* `Observatory.get_standard_headers` will convert to float, e.g. `2000.`
* Always assume a J2000 and don't allow `None` to be set.
* Clean up some comments and make some of the line breaks easier to read.


jamessynge:
We have a ways to go on coverage. Sigh.

pocs/observatory.py
```diff
@@ -379,6 +380,14 @@ def get_standard_headers(self, observation=None):
         # Add observation metadata
         headers.update(observation.status())
 
+        # Explicitly convert EQUINOX to float
+        try:
+            equinox = float(headers['equinox'].replace('J', ''))
+        except BaseException:
+            equinox = 2000.  # We assume J2000
+        finally:
```
jamessynge: Nit: Doesn't need to be a finally clause.

pocs/scheduler/field.py
```diff
@@ -13,13 +13,20 @@ def __init__(self, name, position, equinox='J2000', **kwargs):
...
 
         Arguments:
-            name {str} -- Name of the field, typically the name of object at center `position`
-            position {str} -- Center of field, can be anything accepted by `~astropy.coordinates.SkyCoord`
-            **kwargs {dict} -- Additional keywords to be passed to `astroplan.ObservingBlock`
+            name {str} -- Name of the field, typically the name of object at
+                center `position`
+            position {str} -- Center of field, can be anything accepted by
+                `~astropy.coordinates.SkyCoord`
+            **kwargs {dict} -- Additional keywords to be passed to
```
jamessynge: Looks like you might be wrapping at under 100.

pocs/state/states/default/pointing.py
```diff
@@ -68,23 +68,24 @@ def on_enter(event_data):
...
+                pocs.say("Ok, I've got the pointing picture, let's see how close we are.")
 
-                pocs.logger.debug("Pointing Coords: {}".format(
-                    pointing_image.pointing))
-                pocs.logger.debug("Pointing Error: {}".format(
-                    pointing_image.pointing_error))
+                pocs.logger.debug("Pointing Coords: {}", pointing_image.pointing)
+                pocs.logger.debug("Pointing Error: {}", pointing_image.pointing_error)
 
         # separation = pointing_image.pointing_error.magnitude.value
```
jamessynge: Commented out code (yes, I'm guilty of the same).

pocs/state/states/default/pointing.py
```diff
@@ -104,5 +105,4 @@ def on_enter(event_data):
         pocs.next_state = 'tracking'
 
     except Exception as e:
-        pocs.say(
-            "Hmm, I had a problem checking the pointing error. Sending to parking. {}".format(e))
+        pocs.say("Hmm, I had a problem checking the pointing error. Sending to parking. {}", e)
```
jamessynge: It feels like "sending to parking" is action at a distance given the next_state = parking way at the top.

wtgee: Sort of yes. All states (should) specify `parking` as the next state at the top of the file and only by proper action does it get changed to a different state. This was aiming to bring consistency.

pocs/observatory.py
wtgee: I was mostly trying to make that a contained block. Can remove.

PR #345, opened 2018-01-15
Remove output of temp sensor device info.

jamessynge:
Remove call to PrintDeviceInfo.

Change DallasTemperatureHandler::PrintDeviceInfo to output JSON.
Extract DeviceInfo out into a top-level struct, just to make things a
bit easier to manage.


jamessynge:
I wasn't that happy with the behavior when I tried that, so I put that idea on hold.

PR #352, opened 2018-01-15
Speed up rate of sending commands.

jamessynge:
And continue sending commands a bit after first getting the desired
response, in case that helps with momentum.

### DIFF
--- a/pocs/dome/astrohaven.py
+++ b/pocs/dome/astrohaven.py
@@ -43,8 +43,10 @@ class AstrohavenDome(abstract_serial_dome.AbstractSerialDome):
     """
     # TODO(jamessynge): Get these from the config file (i.e. per instance), with these values
     # as defaults, though LISTEN_TIMEOUT can just be the timeout config for SerialData.
-    LISTEN_TIMEOUT = 3  # Max number of seconds to wait for a response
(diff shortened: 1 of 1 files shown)


jamessynge:
Will test with real hardware later today after the snow showers are done over Wheaton.

pocs/dome/astrohaven.py
```diff
@@ -157,7 +159,8 @@ def _full_move(self, send, target_feedback):
         # In other words, we'll try to read status, but if it isn't available,
         # we'll just send another command.
         saved_timeout = self.serial.ser.timeout
-        self.serial.ser.timeout = 0.2
+        self.serial.ser.timeout = max(0.0000001, AstrohavenDome.MOVE_LISTEN_TIMEOUT)
```
wtgee: Just to guarantee timeout is never 0?

jamessynge: Yes, though I now see that 0 does not imply forever. None is used for that. 0 means non-blocking (i.e. read as much as available AND requested, but don't wait if requested amount/terminator is not there).

jamessynge:
FYI, still cloudy at Wheaton. Next clearing appears to be 10am tomorrow.

jamessynge:
Worked great at Wheaton this afternoon.

Issue #351, opened 2018-01-15
Investigate PAN001 tracking failure


wtgee:
Tried to correct Dec for over 30 seconds even though total correction should have be 9ms. Maybe too little of a correction? But the mount should have reported `Tracking` in status for a skip.

This was on the 43rd image and had been working fine until then. Target still about 45 minutes from reaching meridian.

```
D0115 10:44:06.673 observatory.py:277   Offset Info: OffsetError(delta_ra=<Angle 12.562124399983077 arcsec>, delta_dec=<Angle 0.04534232400104088 arcsec>, magnitude=<Angle 12.438572528277144 arcsec>)
D0115 10:44:06.674 machine.py:258       After calling analyze. Now in analyzing state
D0115 10:44:06.676 machine.py:300       Source: analyzing        Dest: tracking
D0115 10:44:06.676 machine.py:161       Transition method: adjust_tracking
D0115 10:44:06.676 machine.py:200       Checking safety for adjust_tracking
D0115 10:44:06.775 core.py:328          Dark Check: True
D0115 10:44:06.777 core.py:364          Weather Safety: True [11 sec old - 2018-01-15 10:43:55.597000]
D0115 10:44:10.937 machine.py:244       Before calling adjust_tracking from analyzing state
I0115 10:44:18.302 messaging.py:122     PANCHAT Checking our tracking
D0115 10:44:18.303 observatory.py:304   Updating the tracking
I0115 10:44:18.304 observatory.py:331   Adjusting Dec: north 9.04 ms 0.05 arcsec
D0115 10:44:29.979 observatory.py:342   Waiting for Dec tracking adjustment
D0115 10:44:34.639 observatory.py:342   Waiting for Dec tracking adjustment
D0115 10:44:39.290 observatory.py:342   Waiting for Dec tracking adjustment
D0115 10:44:43.951 observatory.py:342   Waiting for Dec tracking adjustment
D0115 10:44:48.610 observatory.py:342   Waiting for Dec tracking adjustment
D0115 10:44:53.260 observatory.py:342   Waiting for Dec tracking adjustment
W0115 10:44:57.921 tracking.py:15       Problem adjusting tracking: Trying to adjust Dec tracking for more than 30 seconds

```

wtgee:
[...]

Issue #350, opened 2018-01-15
Determine if mount guiding corrections are non-blocking


wtgee:
When we correct the tracking we do a `move_ms` command in a specified direction (east or west for RA and north or south for Dec) but we issue the Dec command first and then immediately move to the RA then start the waiting timer. 

During observations on PAN001 it seemed like an error in Dec wasn't getting corrected so it steadily drifted further off. Changing the timer to wait for the Dec before issuing RA commands seemed to help.

Need to determine (in the lab) if you can move both axis simultaneously.

PR #355, opened 2018-01-15
Tracking Updates

wtgee:
* Wait for the Dec axis to move before moving RA axis
* Only update tracking if > 50ms correction
* Some long line clean up

Note: Most of this `update_tracking` code should move into the mount
class and accept `current_offset_info` as a param

Corrects for small adjustments as reported in #351 
Relates to #350

### DIFF
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -1,4 +1,5 @@
 import os
+import time
 
 from collections import OrderedDict
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


pocs/observatory.py
```diff
@@ -148,12 +149,15 @@ def status(self):
                 status['observation']['field_ha'] = self.observer.target_hour_angle(
                     t, self.current_observation.field)
 
+            evening_astro_time = self.observer.twilight_evening_astronomical(t, which='next')
```
jamessynge: What does next produce if at midnight? Tomorrow night?

pocs/observatory.py
```diff
@@ -316,36 +320,48 @@ def update_tracking(self):
             else:
                 ra_direction = 'east'
```
jamessynge: Could you add some more explanation to the docstring, for those of us who are ignorant of mounts. For example, what does 50 milliseconds measure here? An angle? A time?

pocs/observatory.py
wtgee: Correct. The default is `nearest`, which would go backwards. With `next` we always get the upcoming.

pocs/observatory.py
wtgee: Yes, will add.

wtgee: Done. I'm going to merge but let me know if docstring doesn't make sense. I'd like to test these additions on PAN001 tonight.

PR #360, opened 2018-01-16
Add more tests of peas/sensors.py.

jamessynge:
Test auto_detect_port, find_arduino_devices, and
auto_detect_arduino_devices.

### DIFF
--- a/peas/sensors.py
+++ b/peas/sensors.py
@@ -1,6 +1,10 @@
 import json
 import os
+
+# Note: list_comports is modified by test_sensors.py, so if changing
+# this import, the test will also need to be updated.
 from serial.tools.list_ports import comports as list_comports
+
 import sys
 import yaml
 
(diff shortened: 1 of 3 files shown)


jamessynge:
Looks like we need to include peas/ in coverage counting.

peas/sensors.py
```diff
@@ -150,21 +154,24 @@ def find_arduino_devices():
     return [p.device for p in comports if 'Arduino' in p.description]
 
 
-def auto_detect_port(port, logger):
+def auto_detect_port(port, logger=None):
```
wtgee: I think I was confused by the name of the method. We are not detecting the port but the board type (since you are passing a port).

peas/tests/test_sensors.py
```diff
@@ -65,3 +79,55 @@ def test_create_telemetry(serial_handlers):
...
+    assert sensors_module.auto_detect_port('loop://') is None
+
+
+def test_no_handler_installed():
+    """Can't find our simulator, so fails."""
+    assert sensors_module.auto_detect_port('arduinosimulator://?board=telemetry') is None
+
+
+def test_auto_detect_port_telemetry(serial_handlers):
+    """Auto-detect will fail because loop:// handler doesn't print anything."""
```
wtgee: Copied docstring is incorrect.

peas/tests/test_sensors.py
```diff
@@ -65,3 +79,55 @@ def test_create_telemetry(serial_handlers):
...
+
+def test_auto_detect_port_fail():
+    """Auto-detect will fail because loop:// handler doesn't print anything."""
+    #pytest.set_trace()
+    assert sensors_module.auto_detect_port('loop://') is None
+
+
+def test_no_handler_installed():
+    """Can't find our simulator, so fails."""
+    assert sensors_module.auto_detect_port('arduinosimulator://?board=telemetry') is None
```
wtgee: Sorry, why is this one `None`? Seems same as below that work.

peas/sensors.py
jamessynge: Fixed.

peas/tests/test_sensors.py
jamessynge: Comments updated to explain.

peas/tests/test_sensors.py
jamessynge: Fixed.

Issue #362, opened 2018-01-16
Add support for FLI cameras


AnthonyHorton:
Add support for cameras from Finger Lake Instruments (FLI).

AnthonyHorton:
Closed by #363

PR #363, opened 2018-01-16
 Add support for FLI cameras

AnthonyHorton:
This adds support for cameras from Finger Lake Instruments (Issue #362).

In the process I've also taken the opportunity to remove a lot of the duplication of code that had crept into the various camera and focuser classes.


pocs/camera/camera.py
```diff
@@ -155,14 +172,99 @@ def CCD_cooling_power(self):
...
+                                                                          **kwargs)
+
+        exposure_event = self.take_exposure(seconds=exp_time, filename=file_path, *args, **kwargs)
+
+        # Add most recent exposure to list
+        if self.is_primary:
+            observation.exposure_list[image_id] = file_path
+
+        # Process the exposure once readout is complete
+        t = Thread(target=self.process_exposure, args=(metadata, camera_event, exposure_event))
```
wtgee: I've been wondering lately if we should return this thread process along with the `Event`. Might be nice to be able to hold on to it in case we need to join or kill or something. Any experience with needing that on the SBIGs or FLIs?

pocs/camera/camera.py
```diff
@@ -155,14 +172,99 @@ def CCD_cooling_power(self):
...
+                when the exposure is complete, triggering the processing.
+        """
+        # If passed an Event that signals the end of the exposure wait for it to be set
+        if exposure_event is not None:
+            exposure_event.wait()
+
+        image_id = info['image_id']
+        seq_id = info['sequence_id']
+        file_path = info['file_path']
+        self.logger.debug("Processing {} {}".format(image_id))
```
wtgee: There are two `{}` in the string but only one thing passed to `format`

pocs/camera/camera.py
```diff
@@ -155,14 +172,99 @@ def CCD_cooling_power(self):
...
+
+        if info['is_primary']:
+            self.logger.debug("Adding current observation to db: {}".format(image_id))
+            self.db.insert_current('observations', info, include_collection=False)
+        else:
+            self.logger.debug('Compressing {}'.format(file_path))
+            fits_utils.fpack(file_path)
+
+        self.logger.debug("Adding image metadata to db: {}".format(image_id))
+        self.db.observations.insert_one({
```
wtgee: We have notes to work on this elsewhere, but if `info` contains items that cannot be serialized (like was happening with `EQUINOX`) then this command fails in which case the `signal_event.set()` never happens and the whole system hangs. We want to clean up that serialization but I would add a `try/except` around this call in the base class.

pocs/camera/camera.py
```diff
@@ -251,9 +353,105 @@ def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_file=False, *ar
         image = fits.getdata(file_path)
         if not keep_file:
             os.unlink(file_path)
-        thumbnail = images.crop_data(image, box_width=thumbnail_size)
+        thumbnail = img_utils.crop_data(image, box_width=thumbnail_size)
         return thumbnail
 
+    def _fits_header(self, seconds, dark=None):
```
wtgee: Could probably move this to `fits_utils` file.

pocs/camera/camera.py
```diff
@@ -251,9 +353,105 @@ def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_file=False, *ar
...
+        header = fits.Header()
+        if isinstance(seconds, u.Quantity):
+            seconds = seconds.to(u.second)
+            seconds = seconds.value
+        header.set('INSTRUME', self.uid, 'Camera serial number')
+        header.set('DATE-OBS', Time.now().fits)
```
wtgee: This doesn't look right with the `.fits` at the end.

pocs/camera/fli.py
```diff
@@ -0,0 +1,236 @@
...
+# Properties
+
+    @AbstractCamera.uid.getter
+    def uid(self):
+        # Unlike Canon DSLRs 1st 6 characters of serial number is *not* a unique identifier.
+        # Need to use the whole thing.
+        return self._serial_number
+
+    @property
+    def CCD_temp(self):
```
wtgee: Should be avoiding the capital letters in methods and properties.

pocs/camera/fli.py
```diff
@@ -0,0 +1,236 @@
...
+        # Create an instance of the FLI Driver interface
+        self._FLIDriver = libfli.FLIDriver()
+        self.connect()
+
+        self.filter_type = filter_type
+
+        if self.is_connected:
+            if set_point is None:
+                # FLI *always* thermoregulates, all you can do is set a high set point
+                set_point = 25 * u.Celsius
```
wtgee: Should this come from a config somewhere just in case? I see it repeated below as well.

pocs/camera/fli.py
```diff
@@ -0,0 +1,236 @@
...
+        # Unlike Canon DSLRs 1st 6 characters of serial number is *not* a unique identifier.
+        # Need to use the whole thing.
+        return self._serial_number
+
+    @property
+    def CCD_temp(self):
+        return self._FLIDriver.FLIGetTemperature(self._handle)
+
+    @property
+    def CCD_set_point(self):
```
wtgee: Docstring of what happens on set would be nice here.

pocs/camera/fli.py
```diff
@@ -0,0 +1,236 @@
...
+                      *args,
+                      **kwargs):
+        """
+        Take an exposure for given number of seconds and saves to provided filename.
+
+        Args:
+            seconds (u.second, optional): Length of exposure
+            filename (str, optional): Image is saved to this filename
+            dark (bool, optional): Exposure is a dark frame (don't open shutter), default False
+
```
wtgee: Missed the `blocking` param.

pocs/camera/fli.py
```diff
@@ -0,0 +1,236 @@
...
+            except RuntimeError as err:
+                self.logger.error('FLI camera readout error: expected {} rows, got {}!'.format(
+                    image_data.shape[0], i
+                ))
+                self.logger.error(err)
+                break
+
+        self._exposure_lock.release()
+
+        fits_utils.write_fits(image_data, header, filename, self.logger, exposure_event)
```
wtgee: Do you want to `write_fits` if you get the `RuntimeError` above?

pocs/camera/libfliconstants.py
```diff
@@ -0,0 +1,122 @@
+# c_long flidev_t
```
wtgee: My eyes glazed over at this and I'm just going to trust you. :)

pocs/camera/sbig.py
```diff
@@ -60,12 +58,19 @@ def CCD_set_point(self):
     @CCD_set_point.setter
     def CCD_set_point(self, set_point):
```
wtgee: Seeing this repeated makes me wonder if we should have a `CCDCamera` that inherits from the base and then the SBIG and FLI inherit from `CCDCamera`. Not sure if that would help though.

pocs/camera/sbig.py
```diff
@@ -221,91 +138,35 @@ def take_exposure(self,
...
-        signal_event.set()
+        # Unbinned. Need to chance if binning gets implemented.
+        readout_mode = 'RM_1X1'
+
+        header.set('CAM-FW', self._info['firmware version'], 'Camera firmware version')
+        header.set('XPIXSZ', self._info['readout modes'][readout_mode]['pixel width'].value,
+                   'Microns')
+        header.set('YPIXSZ', self._info['readout modes'][readout_mode]['pixel height'].value,
+                   'Microns')
+        header.set('EGAIN', self._info['readout modes'][readout_mode]['gain'].value,
```
wtgee: I want to say we have a `GAIN` that comes of the Canon DSLR. Is `EGAIN` a better name?

pocs/tests/test_camera.py
```diff
@@ -222,7 +206,7 @@ def test_exposure(camera, tmpdir):
     # A one second normal exposure.
     camera.take_exposure(filename=fits_path)
     # By default take_exposure is non-blocking, need to give it some time to complete.
-    time.sleep(5)
+    time.sleep(10)
```
wtgee: Seems like a long timeout. Is this whole file skipped if camera is not detected?

pocs/utils/images/fits.py
```diff
@@ -320,3 +320,50 @@ def fpack(fits_fname, unpack=False, verbose=False):
...
+def update_headers(filepath, info):
+    with fits.open(file_path, 'update') as f:
+        hdu = f[0]
+        hdu.header.set('IMAGEID', info.get('image_id', ''))
+        hdu.header.set('SEQID', info.get('sequence_id', ''))
+        hdu.header.set('FIELD', info.get('field_name', ''))
+        hdu.header.set('RA-MNT', info.get('ra_mnt', ''), 'Degrees')
+        hdu.header.set('HA-MNT', info.get('ha_mnt', ''), 'Degrees')
+        hdu.header.set('DEC-MNT', info.get('dec_mnt', ''), 'Degrees')
+        hdu.header.set('EQUINOX', info.get('equinox', 2000.))  # Assume J2000
```
wtgee: I think this changed again so we just need to make sure it's okay in the end.

pocs/camera/camera.py
```diff
@@ -140,6 +147,16 @@ def CCD_cooling_enabled(self):
         """
         raise NotImplementedError
 
+    @CCD_cooling_enabled.setter
+    def CCD_cooling_enabled(self, enabled):
+        """
+        Set status of the camera's image sensor cooling system (enabled/disabled).
+
+        Note: this only needs to be implemented for cameras which have cooled image sensors,
```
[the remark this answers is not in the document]
AnthonyHorton: Always returning `False` makes sense. The corresponding setter is still there to return `NotImplementedError` to make it clear when the camera doesn't support cooling.

pocs/camera/camera.py
AnthonyHorton: Oops!

pocs/camera/camera.py

pocs/camera/fli.py
AnthonyHorton: What I'm doing here is using a set point of 25C as a stand in for turning the cooling off, because you can't actually disable the cooling on the FLI cameras.  I suppose we could (should?) make the 'no cooling' set point an (optional) parameter in the config file.

pocs/camera/camera.py
AnthonyHorton: I put it here because this is the function to generate the FITS header entries that (a) relate to the camera itself and (b) can be generated by the base class. There are similar functions in the derived classes that add additional camera related headers for which the implementation requires camera type specific code, and in both the base and Birger focuser classes.

I could move the base class methods to `fits_utils` but then I'd have to pass it the camera and focuser objects, and I'd still need FITS header generation methods in at least the derived camera and focuser classes to handle the hardware specific aspects.

pocs/camera/fli.py
AnthonyHorton: OK.

pocs/camera/fli.py
AnthonyHorton: Torn between two conventions :-) Should be lower case from Python conventions point of view, should be upper case as 'CCD' is an acronym. I'll lower them if you prefer.

pocs/camera/fli.py
AnthonyHorton: Will add.

pocs/camera/fli.py
AnthonyHorton: Arguable. The error could occur part way through readout, in which case you've still got part of an image that you might want to look at. Doesn't hurt to try to save regardless?

pocs/camera/libfliconstants.py
AnthonyHorton: Not much more of a copy and paste from the header of the C library! Just had to change syntax from C to Python.

pocs/camera/sbig.py
AnthonyHorton: I don't think they've got a lot of common code, now that I've moved a bunch of stuff into the base class.  Worth considering, though.

pocs/camera/sbig.py
AnthonyHorton: `EGAIN` (electrical gain? electron gain? electronic gain?) is what's used in SBIG's (unofficial) extension to the FITS standard (http://diffractionlimited.com/sbig-fits-extensions-sbigfitsext-1-0/) so that's what I used here.

pocs/tests/test_camera.py
AnthonyHorton: The long timeout turned out to be necessary for the FLI, which has a very long readout time. This test will run with the camera simulator even with no real cameras present, and the simulator needs less time. I should probably parameterise the timeout to avoid unnecessary delays.

pocs/utils/images/fits.py
AnthonyHorton: I took the latest version from `cr2_to_fits` but yes, would be good to test.

pocs/camera/camera.py
AnthonyHorton: I don't recall any issues with the `process_exposure` thread getting stuck. The readout threads are a different matter, though, seen those get stuck many times due to USB problems.

pocs/camera/camera.py
AnthonyHorton: It's a bit ugly. Better as
```
now = Time.now()
header.set('DATE-OBS', now.fits)
```
?

pocs/camera/camera.py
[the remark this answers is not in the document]
AnthonyHorton: Done

pocs/camera/fli.py
AnthonyHorton: The second use of this was a hang over from the way the SBIG class worked, where in order to disable the cooling you set the set point to `None`. I'd already changed that counter intuitive behaviour of the SBIG class so the corresponding bit from FLI too.  Now just have 25C in there as a simple default value for the FLI camera set point.

AnthonyHorton:
One more thing I'd like to do to this before merge: add the option to search device nodes for an FLI camera with a specified serial number.  This would work very similarly to the code for the Birger focusers and, like the SBIG camera class, allow us to specify FLI cameras in the config file by their serial numbers instead of having to directly specify the device node, which is non-deterministic.

pocs/camera/camera.py
wtgee: Makes sense, especially upon seeing what else was implemented further down in the PR.

pocs/camera/fli.py
wtgee: Good point. Let's leave for now.

pocs/camera/sbig.py
wtgee: I was just wondering. I'll double-check on what goes in. We're also not reading it back out for anything at the moment.

wtgee:
> One more thing I'd like to do to this before merge: add the option to search device nodes for an FLI camera with a specified serial number. This would work very similarly to the code for the Birger focusers and, like the SBIG camera class, allow us to specify FLI cameras in the config file by their serial numbers instead of having to directly specify the device node, which is non-deterministic.

This is similar to some work that @jamessynge has been doing for the arduino sensors and I think is generally a good way to go.  We should probably do the same for the DSLRs as they can vacillate between which one is Cam00 and Cam01.

AnthonyHorton:
OK, I've now implemented the ability to specify FLI cameras by serial number instead of device node. Appears to be Working As Intended (there's a demo in the example notebook: https://github.com/AstroHuntsman/POCS/blob/FLI-camera/examples/notebooks/FLI%20camera%20control.ipynb).  While I was doing that I also tweaked the corresponding section of the Birger focuser class, and fixed a bug I discovered in `libfli`. As far as I'm concerned this is now feature complete, as least for now.  Would be good to test it with DSLRs before merging through, just in case I've broken something that's not showing up in the tests. A little concerned that I have no SBIG cameras available to test with, too.

This is now passing the test suite on my computer, though emitting 1 warning which seems to be due to a failure to make a pretty image from the simulated cameras test observation.

AnthonyHorton:
Something broken in my local install at present, `test_codestyle` always passes despite violations that show up in Travis CI. Frustrating.

wtgee:
> Something broken in my local install at present, test_codestyle always passes despite violations that show up in Travis CI. Frustrating.

We added coverage to the `peas` directory without first cleaning up the directory, so there were some errors appearing there. Merging from upstream should resolve that if the issue.

> Would be good to test it with DSLRs before merging through, just in case I've broken something that's not showing up in the tests.

I'll be in lab today so will test this and then will approve.

pocs/camera/canon_gphoto2.py
```diff
@@ -86,60 +86,12 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
...
-            'is_primary': self.is_primary,
-            'sequence_id': sequence_id,
-            'start_time': start_time,
-        }
-        metadata.update(headers)
-        exp_time = kwargs.get('exp_time', observation.exp_time.value)
+        exp_time, file_path, image_id, metadata = self._setup_observation(observation,
+                                                                          headers,
+                                                                          filename,
+                                                                          *args,
```
wtgee: `take_observation` for gphoto2 needs `*args` added up above.

pocs/camera/camera.py
```diff
@@ -155,14 +172,102 @@ def CCD_cooling_power(self):
...
+
+        try:
+            self.logger.debug("Extracting pretty image")
+            img_utils.make_pretty_image(file_path,
+                                        title=info['field_name'],
+                                        primary=info['is_primary'])
+        except Exception as e:
+            self.logger.warning('Problem with extracting pretty image: {}'.format(e))
+
+        self._process_fits(filepath, info)
```
wtgee: `filepath` -> `file_path`

pocs/camera/camera.py
```diff
@@ -251,9 +356,106 @@ def get_thumbnail(self, seconds, file_path, thumbnail_size, keep_file=False, *ar
...
+            'is_primary': self.is_primary,
+            'sequence_id': sequence_id,
+            'start_time': start_time,
+        }
+        metadata.update(headers)
+        exp_time = kwargs.get('exp_time', observation.exp_time.value)
+
+        return exp_time, file_path, image_id, metadata
+
+    def _process_fits(self, filepath, info):
```
wtgee: `filepath` -> `file_path`

pocs/camera/canon_gphoto2.py
```diff
@@ -197,52 +149,11 @@ def take_exposure(self, seconds=1.0 * u.second, filename=None):
...
-
-        Converts the CR2 to a FITS file. If the camera is a primary camera, extract the
-        jpeg image and save metadata to mongo `current` collection. Saves metadata
-        to mongo `observations` collection for all images
-
-        Args:
-            info (dict): Header metadata saved for the image
-            signal_event (threading.Event): An event that is set signifying that the
-                camera is done with this exposure
+    def _process_fits(self, filepath, info):
```
wtgee: `filepath` -> `file_path`

wtgee:
Woohoo! Other than comments above about `filepath` this seems to be working great. Just tested with my two DSLRs in the lab. Not only did it work, but it worked from a Raspberry Pi in the camera box being controlled wirelessly by another Pi! 

This is a good code improvement so let's get merged before too out of sync (there is one merge conflict to resolve - I can send you a patch for my working system if you want but shouldn't be too bad).

AnthonyHorton:
OK @wtgee, I fixed the `filepath/file_path` thing, and the missing `*args`, then did another merge from `develop` into `FLI-camera` so that I could address the merge conflict. I spotted the two changes in `canon_gphoto2.py` and applied them to the base class `process_observation()` and `_setup_observation_methods()`. Should be good to go!

pocs/camera/camera.py
```diff
@@ -155,14 +172,101 @@ def CCD_cooling_power(self):
...
+
+        try:
+            self.logger.debug("Extracting pretty image")
+            img_utils.make_pretty_image(file_path,
+                                        title=info['field_name'],
+                                        primary=info['is_primary'])
+        except Exception as e:
+            self.logger.warning('Problem with extracting pretty image: {}'.format(e))
+
+        self._process_fits(file_path, info)
```
wtgee: For the gphoto2 class, this method puts the name of the fits file into `info`, replacing the cr2 file. However down below on line 259 the `file_path` (still `.cr2`) is passed into `fpack` as opposed to `info['file_path']`. Either need to return the file here so `file_path = self._process_fits(file_path, info)` or to duplicate line 238 below this line. Or change 259 to `...fpack(info['file_path']`.

That's the only change I had to make and everything else is looking good!

AnthonyHorton: Ah, OK. I opted to have `_process_fits()` return the FITS file path, and changed this line to `file_path = self._process_fits(file_path, info)`.

wtgee: Yes I think that makes the most sense. Will test shortly.

wtgee:
Other than the fact that it takes 2.5 minutes to convert the CR2 to FITS on a Raspberry Pi, this looks good. :)

This is a biggish change and we have a few other changes coming through soon, so I'd like to get a point release out before merging this in. Will try to do so tonight. I'll approve after that just so it doesn't accidentally get merged before.

jamessynge:
My apologies, I've not reviewed this yet. Do you want a review now, or are there changes coming in response to Wilfred's review.

AnthonyHorton:
@jamessynge No further changes expected. Last few commits were just fixing merge conflicts that had arisen due to changes on `develop`.

wtgee:
Sorry, was waiting until I did a release to merge but then it got hung up.
We will probably just start tagging releases off main branch soon so this
could be merged. Buying more weatherproof items now so will get to later
today

wtgee:
@AnthonyHorton after merging this I now get core dumps while test_camera.py is running and @jamessynge sees the same thing. Not sure why it was passing on travis. I seem to remember seeing these before though. Can you test on your system?

wtgee:
Ah, I think this is because I took out the `Agg` matplotlib backend by default. I'll try to fix that first but I think that is what is causing the issue.

AnthonyHorton:
It was matplotlib backend issues last time we got core dumps.

wtgee:
Yes, sorry, forgot to update. It was an incorrect matplotlib backend because I removed the explicit `use('Agg')` statement in favor of an env variable that points to `$POCS/.matplotlib/` and contains an rc file. Still not sure this is ideal. But the camera code is fine, sorry for false alarm.

AnthonyHorton:
Closes #362

Issue #367, opened 2018-01-16
Horizon Constraints


wtgee:
This is an umbrella issue for a number of things that can change with the horizon constraints:

- [ ] Account for entire Field of View #383 
- [ ] Determine how Altitude score should work #369 
- [ ] Test Altitude for entire time duration #365
- [ ] Horizon from image #364
- [ ] ~~Mark current image as obstructed (maybe via PAWS?) #392~~
- [x] ~~Potentially combine Altitude with Duration #366~~
- [x] Accepts lists instead of tuples #394
- [x] Read horizon info from config file #416

jermainegug:
We can now tick off issue #394 :)

wtgee:
I closed it as part of the merge, which is typically how you want to handle
it.

Issue #364, opened 2018-01-16
Get horizon/obstructions from image


wtgee:
@brendan-o had done some work to automatically find the horizon images from a picture, it would be nice to get that back in.

wtgee:
This is labelled as a good first issue but is not necessarily simple.

The basic idea would be to take an image that has a visible obstruction in it and figure out what the altitude and azimuth of that location is given the WCS (or other mount information).

Example obstruction (the Iron Giant!)

![Iron Giant](https://forum.projectpanoptes.org/uploads/default/original/1X/0fd4f3e3a7b5cbc2c676b2e16c6a9b4dc69aa096.jpeg)

PR #368, opened 2018-01-16
Basic horizon constraint features

wtgee:
This is an attempt to get basic support for a custom horizon into the codebase. This is neither feature complete nor robust but hopefully is enough of a stepping stone such that others could work on the code. 

Basic fixes and suggestions are welcome here but I'd prefer to get this merged soon. To that end, there is an umbrella issue #367 and any additional features, tests, etc. should get their own issue added to that umbrella issue. Hopefully those smaller issues can then be doled out to individuals to work on.

See the [notebook](https://github.com/wtgee/POCS/blob/c705276958bd885cbbae300270afce16006f89c1/examples/notebooks/Horizon%20Obstruction%20Limits.ipynb) first for an example of what is going on.

Would appreciate basic comments and suggestions for features/testing from @joshwalawender @AnthonyHorton @lspitler @jamessynge. Mostly just on the overall idea. Please go ahead and create any related issues.


pocs/scheduler/constraint.py
```diff
@@ -146,11 +107,11 @@ def get_score(self, time, observer, observation, **kwargs):
         except KeyError:
             self.logger.error("Moon must be set")
 
-        moon_sep = moon.separation(observation.field.coord).value
+        moon_sep = observation.field.coord.separation(moon).value
```
wtgee: I'm not sure why this changed, I'll check it out.

pocs/observatory.py
```diff
@@ -734,8 +736,20 @@ def _create_scheduler(self):
...
+
+                horizon = horizon_utils.Horizon(
+                    obstructions=obstruction_list,
+                    base_horizon=default_horizon
+                )
+
                 # Simple constraint for now
-                constraints = [MoonAvoidance(), Duration(30 * u.deg)]
+                constraints = [
+                    Altitude(horizon=horizon),
```
wtgee: For now this will just create a flat line at the base horizon level so shouldn't affect current scheduler.

pocs/scheduler/constraint.py
```diff
@@ -146,11 +107,11 @@ def get_score(self, time, observer, observation, **kwargs):
...
 
         # This would potentially be within image
         if moon_sep < 15:
-            self.logger.debug("Moon separation: {}".format(moon_sep))
+            self.logger.debug("\t\tMoon separation: {:.02f}".format(moon_sep))
```
jamessynge: `\t\t` seems unnecessary.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +120,40 @@ def get_score(self, time, observer, observation, **kwargs):
 
     def __str__(self):
         return "Moon Avoidance"
+
+
+class Altitude(BaseConstraint):
```
jamessynge: The move down the file seems unnecessary, FWIW.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +120,40 @@ def get_score(self, time, observer, observation, **kwargs):
...
+
+    """ Implements altitude constraints for a horizon """
+
+    def __init__(self, horizon=None, *args, **kwargs):
+        """
+        Retrieves the coordinate list from the config file and validates it
```
jamessynge: Doesn't appear to retrieve anything.

pocs/scheduler/dispatch.py
```diff
@@ -53,15 +53,15 @@ def get_observation(self, time=None, show_all=False, reread_fields_file=False):
...
+            self.logger.info("Checking Constraint: {}".format(constraint))
             for obs_name, observation in self.observations.items():
                 if obs_name in valid_obs:
                     self.logger.debug("\tObservation: {}".format(obs_name))
 
                     veto, score = constraint.get_score(
                         time, self.observer, observation, **common_properties)
 
-                    self.logger.debug("\t\tScore: {}\tVeto: {}".format(score, veto))
+                    self.logger.debug("\t\tScore: {:.05f}\tVeto: {}".format(score, veto))
```
jamessynge: Again with the leading tabs. I suppose it makes it easier to find... sometimes...

pocs/observatory.py
```diff
@@ -734,8 +736,20 @@ def _create_scheduler(self):
                 module = load_module(
                     'pocs.scheduler.{}'.format(scheduler_type))
 
+                obstruction_list = self.config['location'].get('obstructions', list())
```
jamessynge: Is there a description of the values in the list somewhere?

pocs/utils/images/horizon.py
```diff
@@ -0,0 +1,88 @@
+import numpy as np
```
jamessynge: I'm thinking about how to produce these from other info, like the azimuth and height above "ground" of the obstruction, and the height above ground of the mount/cameras (probably the lowest height).

pocs/scheduler/constraint.py
wtgee: It helps with the output in this case and makes it consistent with the other scheduler items. If we don't want here we should remove everywhere.

pocs/scheduler/constraint.py
wtgee: This is a direct pep8 change: https://www.python.org/dev/peps/pep-0008/#blank-lines

pocs/utils/images/horizon.py
wtgee: You mean how to produce a Horizon automatically? That's partly what #364 is about.

pocs/scheduler/constraint.py
wtgee: Oops, forgot to update docstring, thanks.

pocs/observatory.py
wtgee: Just it the `Horizon` class. I haven't taken the step to add these to the config yet so will do that soon.

pocs/utils/images/horizon.py
jamessynge: Not precisely; I meant "get horizon from measurements" as opposed to "get horizon from image". If the latter is going to be easy to achieve, great. But if it will be a while, then I could ask one of the student's I'm working with to help with measuring the space so we can build the former. I could ask him about implementing it.

pocs/scheduler/constraint.py
jamessynge: No, I mean the order of the classes was changed, with Altitude moved down the file. It was above class Duration, now is below.

pocs/observatory.py
jamessynge: Maybe a comment here pointing to the Horizon class?

pocs/scheduler/constraint.py
jamessynge: Remove everywhere. I've been guilty of making these changes in the past, and they stay long after they are useful. Perhaps it would be better to use a custom log colorizer.

pocs/utils/images/horizon.py
jamessynge: Regarding why this might be useful: the big obstruction next to PAN006 has a constant height, but its altitude at different azimuths will be very different. We could encode it with 5 measurements:
1. Height of wall.
2. Compass angle and distance to start of wall.
3. Compass angle and distance to end of wall.

wtgee: Got it. We should capture that in a wiki page or document so have an Issue in #367.

pocs/scheduler/constraint.py
wtgee: Oh, got it. Yes, this was originally added to the bottom of the file as a new constraint named `Horizon`, but then I realized after writing it that it basically replaced the existing Altitude constraint (which wasn't being used at all) so I deleted the one  from above and renamed the new one I had written.

wtgee: Alphabetical order probably makes the most sense.

pocs/scheduler/constraint.py
wtgee: Agreed about the tabs but I do think they are useful here. Long ago I had them all over the logs and it became a nightmare. Then I started thinking about how to write some logger wrapper that would auto-figure out the indent. Luckily I wisely abandoned the whole idea and removed all indents...except for here in the scheduler call.

I know @AnthonyHorton has used some in the SBIG and FLI code and I've been meaning to mention removing them. ;)

I'm going to leave in here unless very serious concerns (see notebook for how output looks) and again the additions here were to make the whole scheduler output consistent as it otherwise looked pretty bad. 

Agree that we should try to avoid them elsewhere (i.e I get to be the exception but no one else. 😁).

lspitler:
Only aspect I don't see (it might be elsewhere in dispatch) is whether the field of view of the particular imaging system being used is factored into the horizon. For Huntsman this is tiny compared to Panoptes so we can go much closer to the horizon. Won't add a to-do for this as I don't know if it is already taken care of.

wtgee:
> Only aspect I don't see (it might be elsewhere in dispatch) is whether the field of view of the particular imaging system being used is factored into the horizon. For Huntsman this is tiny compared to Panoptes so we can go much closer to the horizon. Won't add a to-do for this as I don't know if it is already taken care of.

No, this is a good point and I mean to put an Issue in for it. We consider this to be the center of the image, which of course is not ideal for PANOPTES. I'll make the issue.

pocs/scheduler/constraint.py
jamessynge: I think you see the slippery slope all to well!

wtgee:
Going to merge this just because we definitely have a need on PAN001, PAN006, and Huntsman. :)

horizon

PR #370, opened 2018-01-16
Add coverage of peas/

jamessynge:


wtgee:
-15%. Ouch! :)

Of course, it's just exposing what wasn't covered so is a good thing.

PR #374, opened 2018-01-17
New sequence for scheduler

jermainegug:
This is far from working (load_config does not seem to be working properly so I can't test anything on my laptop) but it is a start.
Seems to me that we just need something that removes an observation from observation list then checks if the observation list is empty and if it is then shut down POCS. (Issue https://github.com/AstroHuntsman/POCS/issues/5)

Just need suggestions on how to implement this properly.


wtgee:
Stupid notebook diffs make it hard for me to tell what has changed.

In general though, we are going to want to make a new file based off `dispatch.py` and then alter the code in there. The notebook (or maybe a separate one) should just show an example of the observing session to demonstrate how it schedules.

jermainegug:
So I should make a new file that does what the Scheduler Playground does but can actually be used?

wtgee:
You should make a new file that is almost the same as `dispatch.py`. The Scheduler Playground is nothing but an example and a playground to test things out. It can't actually be used for anything. The Scheduler Playground should be an example of what the real code does, it should not be the real code.

PR #375, opened 2018-01-17
PEAS shell updates.

### LINKED ISSUE #371: peas_shell status cmd should show if running
There is a start command, but after that you can't tell if it is running. And IIUC, running start again doesn't complain that it is already running.

jamessynge:
Prevent start command from being repeated.
Prevent load_* commands while timer loop is running.
Add command completion for last_reading and toggle_relay.
Improve error reporting/robustness of change_delay.

Fixes #371.

### DIFF
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -41,6 +41,11 @@ class PanSensorShell(cmd.Cmd):
 
     def do_status(self, *arg):
         """ Get the entire system status and print it pretty like! """
+        if self._keep_looping:
+            console.color_print("{:>12s}: ".format('Loop Timer'), "default", "active", "lightgreen")
+        else:
+            console.color_print("{:>12s}: ".format('Loop Timer'), "default", "inactive", "yellow")
+
         for sensor_name in ['environment', 'weather', 'webcams']:
             if sensor_name in self.active_sensors:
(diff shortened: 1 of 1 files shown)


bin/peas_shell
```diff
@@ -49,18 +54,28 @@ class PanSensorShell(cmd.Cmd):
...
+            return
+        print_info('*' * 80)
+        print("{}:".format(device.upper()))
+
+        rec = None
+        if device == 'weather':
+            rec = self.db.current.find_one({'type': 'weather'})
+        elif device == 'environment':
+            rec = self.db.current.find_one({'type': 'environment'})
+
```
wtgee: Since you are here, might be nice to pull the date from the `rec` and give a report of how stale it is. Just a `x seconds old` at bottom.

bin/peas_shell
```diff
@@ -269,6 +331,8 @@ class PanSensorShell(cmd.Cmd):
             if not delay:
                 delay = self._loop_delay
 
+            # WARNING: It appears we have a single _timer attribute, but we create
```
wtgee: Yikes. Good catch. Want to change this now?

jamessynge: No, I want to figure out a better design first (e.g. a separate process for each board and another for the weather sensor).

PR #377, opened 2018-01-17
Fix codestyle of test_sensors.py.

jamessynge:


peas/tests/test_sensors.py
```diff
@@ -5,7 +5,7 @@
 import serial
 
 from peas import sensors as sensors_module
-#from peas.sensors import ArduinoSerialMonitor
+from peas.sensors import ArduinoSerialMonitor
```
wtgee: I don't see this used. Is it coming in your upcoming PR?

wtgee: Actually, I show that a few of these modules aren't used. One of my plugins highlights them:

unused-modules

jamessynge: Please let me know which plugin that is.

And I'll remove the currently unnecessary imports.

PR #378, opened 2018-01-17
Make sequencer constraint

jermainegug:
New PR for issue https://github.com/AstroHuntsman/POCS/issues/5. This replaces #374. 

Note: this is not to be merged yet! Much more work is to be done.


pocs/scheduler/constraint.py
```diff
@@ -159,3 +159,23 @@ def get_score(self, time, observer, observation, **kwargs):
 
     def __str__(self):
         return "Moon Avoidance"
+
+class Seqencer(BaseConstraint):
```
wtgee: Seqencer -> Sequencer

But let's maybe name that `AlreadyVisited` just to be really obvious. Sequencer seems to apply to more than one targets whereas a constraint technically only applies to one target at a time.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +159,23 @@ def get_score(self, time, observer, observation, **kwargs):
...
+        target = observation.field
+        observed = observation.observed_list
+
+        if target in observed:
+            veto = True
+
+    return veto, score * self.weight
+
+    def __str__(self):
+        return "Don't repeat observations"
```
wtgee: How this gets used it's better to actually just have something like "AlreadyVisited". An example of how it shows up is with the word `Altitude` here:

```
I0116 12:30:21.285 dispatch.py:56       Checking Constraint: Altitude
D0116 12:30:21.285 dispatch.py:59       	Observation: Kepler 1100
```

pocs/scheduler/constraint.py
```diff
@@ -170,10 +170,10 @@ def get_score(self, time, observer, observation, **kwargs):
         score = self._score
 
         target = observation.field
-        observed = observation.observed_list
+        observed_list = kwargs['observed_list']
```
wtgee: For normal operations you'll still want to pull from `observation.observed_list` as that's how the scheduler builds it up. So do a`kwargs.get('observed_list', observation.observed_list)` here.

pocs/tests/test_constraints.py
```diff
@@ -197,3 +198,24 @@ def test_moon_avoidance(observer):
...
+    observation1 = Observation(Field('HD189733', '20h00m43.7135s +22d42m39.0645s'))  # HD189733
+    observation2 = Observation(Field('Hat-P-16', '00h38m17.59s +42d27m47.2s'))  # Hat-P-16
+    observation3 = Observation(Field('Sabik', '17h10m23s -15d43m30s'))  # Sabik
+
+    observed_list = OrderedDict()
+
+    observed_list[observation1.seq_time] = observation1
+    observed_list[observation2.seq_time] = observation2
+
+    veto1, score1 = avc.get_score(time, observer, observation1, observed_list=observed_list)
```
wtgee: An alternative to passing the `observed_list` and using the kwargs as above is to just alter the `observer.observed_list` here.

wtgee:
The `MoonAvoidance` and `Duration` are bad examples, but can you add a docstring similar to `Altitude` that describes what this does: https://github.com/jermainegug/POCS/blob/2260d04f5fdaef8135082da61c814ae99678f722/pocs/scheduler/constraint.py#L38

jermainegug:
Haven't been able to test it yet... Windows problems.

wtgee:
You can check the travis build here on the PR. You have some spacing/indent issues it is saying.

jermainegug:
OK, cool. I will just use that to check issues.

pocs/scheduler/constraint.py
```diff
@@ -177,11 +177,12 @@ def get_score(self, time, observer, observation, **kwargs):
...
         score = self._score
 
-        target = observation
+        target = observation.field
         observed_list = kwargs.get('observed_list', observer.observed_list)
 
-        if target in observed_list:
-            veto = True
+        for seq_time, Observation in observed_list:
+            if target = observation:
```
wtgee: You are assigning (`=`) here instead of comparing (`==`). Which might not show an error.

pocs/scheduler/constraint.py
```diff
@@ -177,11 +177,12 @@ def get_score(self, time, observer, observation, **kwargs):
         veto = False
         score = self._score
 
-        target = observation
+        target = observation.field
         observed_list = kwargs.get('observed_list', observer.observed_list)
 
-        if target in observed_list:
-            veto = True
+        for seq_time, Observation in observed_list:
```
wtgee: `Observation` -> `previous_obs` You already have an `observation` object above so can't just use `observation` (no capital letters).

jermainegug:
OK, so I know what it going on now with the `OrderedDict()`. For some reason, which I can't find any explanation for, when I do this:
```
observer.observed_list[observation1.seq_time] = observation1
observer.observed_list[observation2.seq_time] = observation2
```
The previous entries are over written by a recent entry. I will try to just assign it as a normal dictionary but I can't see why he ordered dictionary is behaving in such a way.
If the same thing happens with a normal dictionary then I am completely lost.

wtgee:
> The previous entries are over written by a recent entry. I will try to just assign it as a normal dictionary but I can't see why he ordered dictionary is behaving in such a way.

Not sure I understand what you mean. But the `observed_list` is an `OrderedDict` so that is what the constraint will be given, so that is what needs to be tested.

wtgee:
Ah, I see. The `Observation` requires you to set a `seq_time`. You are not doing that so the `seq_time` is `None` for both entries so they get stored under the same dictionary key. You need something like

```
observation1.seq_time = some_time
observation1.seq_time = some_different_time
```
then put it in the OrderedDict

jermainegug:
Ahh, of course!

jermainegug:
Now we just need to implement this and we are good to merge!

pocs/scheduler/constraint.py
```diff
@@ -159,3 +159,30 @@ def get_score(self, time, observer, observation, **kwargs):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get_score(self, time, observer, observation, **kwargs):
+        veto = False
+        score = self._score
+
+        observed_list = kwargs.get('observed_list', observer.observed_list)
+
+        for seq_time, previous_obs in observed_list.items():
```
wtgee: Actually, now that the OrderedDict is sorted out I think you can switch this back to `if observation.seq_time in observed_list` and avoid the loop.

jermainegug: Awesome, didn't like the idea of the loop but I remembered that `seq_time` will be different for each `observation.

jermainegug:
Thank you! No problem, will do that soon. Good that we have Wasp-35 twice so we can make sure that it works properly in the Scheduler Playground. 

Of course, makes sense to make sure we aren't checking conditions that we won't be using.

wtgee:
@jermainegug Can you run all the cells in the notebook and then save and commit it? I want to be able to see the output in the bottom of the notebook where it vetoes targets because of constraint.

pocs/scheduler/constraint.py
```diff
@@ -159,3 +159,29 @@ def get_score(self, time, observer, observation, **kwargs):
...
+    visited then it will not be considered for a call to become the `current observation`.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get_score(self, time, observer, observation, **kwargs):
+        veto = False
+        score = self._score
+
+        observed_list = kwargs.get('observed_list', observer.observed_list)
```
wtgee: I told you wrong here. `observed_list` belongs to the `scheduler`, not to the `observer` (the `observer` also belongs to the `scheduler`).  You can remove the `observer.observed_list` here and just fetch from `kwargs`. Then in the `dispatch.py` scheduler you can pass the `observed_list` in as a `common_properties` ([here](https://github.com/panoptes/POCS/blob/develop/pocs/scheduler/dispatch.py#L52)):

```
        common_properties = {
            'end_of_night': self.observer.tonight(time=time, horizon=-18 * u.degree)[-1],
            'moon': get_moon(time, self.observer.location),
            'observed_list': self.observed_list
        }
```

The `common_properties` get passed into as `kwargs` down below that.

wtgee: But this means we should add a test. You have added tests to successfully cover just the constraint, now we should add something in `test_dispatch_scheduler.py` that uses it as part of a normal scheduling routine.

jermainegug:
How can I make the `test_already_visited` method work?

pocs/tests/test_constraints.py
```diff
@@ -229,3 +233,28 @@ def test_moon_avoidance(observer):
...
+    observed_list = OrderedDict()
+
+    observation1.seq_time = '01:00'
+    observation2.seq_time = '02:00'
+    observation3.seq_time = '03:00'
+
+    observed_list[observation1.seq_time] = observation1
+    observed_list[observation2.seq_time] = observation2
+
+    veto1, score1 = avc.get_score(time, observer, observation1)
```
wtgee: In `dispatch.py` you [can see](https://github.com/panoptes/POCS/blob/develop/pocs/scheduler/dispatch.py#L62) how common properties is passed to `get_score` by using the `**common_properties` syntax, which passes individual keys of the dictionary to the method (as opposed to passing the dictionary itself).

So in your `get_score` here you just need to pass the `observed_list` with a keyword, so `observed_list=observed_list`. Then the generic `kwargs` in the get_score you wrote will have it available as `kwargs.get('observed_list')`. This is one of the points of using `**kwags` as a parameter, it can pick up keyword arguments that are not otherwise explicitly stated. We have a `get_score` in all of the constraints and not all the keywords apply to all the constraint types so this is how you design for something like that.

jermainegug: Ahh, interesting. I wasn't too sure how `**kwargs` worked but that makes plenty of sense. Thanks!

PR #380, opened 2018-01-17
Fix the observing state wait time

wtgee:
Fixed the `pointing` state the other day but missed this.


pocs/state/states/default/observing.py
```diff
@@ -25,7 +25,7 @@ def on_enter(event_data):
             pocs.logger.debug('Waiting for images: {} seconds'.format(wait_time))
             pocs.status()
 
-            if wait_interval > timeout:
+            if wait_time > timeout:
```
jamessynge: How can we test this?

PR #387, opened 2018-01-18
Adding a few more targets (good for beginning of night for PAN006)

wtgee:
No harm in adding these to the default target list.

@jamessynge HD 23630 is in the Pleiades, which might make a nice pretty picture. Should be good at start of evening. You could bump the priority in order to ensure it is selected or just let it go and see what happens.


jamessynge:
Thanks.

PR #391, opened 2018-01-19
Changing message when setting/getting WCS

### LINKED ISSUE #389: Remove debug message about WCS from Image class.
Isn't a helpful message:
```
D0119 02:48:12.828 images.py:106        Can't get WCS from FITS file (try solve_field)
```

wtgee:
Closes #389

### DIFF
--- a/pocs/images.py
+++ b/pocs/images.py
@@ -102,8 +102,9 @@ class Image(PanBase):
 
                 self.wcs = w
                 self._wcs_file = filename
+                self.logger.debug("WCS loaded from image")
             except Exception:
-                self.logger.debug("Can't get WCS from FITS file (try solve_field)")
+                self.logger.debug("No WCS information present in FITS file")
 
     @property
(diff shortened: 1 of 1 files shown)


pocs/images.py
```diff
@@ -102,8 +102,9 @@ def wcs_file(self, filename):
 
                 self.wcs = w
                 self._wcs_file = filename
+                self.logger.debug("WCS loaded from image")
             except Exception:
-                self.logger.debug("Can't get WCS from FITS file (try solve_field)")
+                pass
```
jamessynge: This is an odd setter, right? It is a no-op if the file doesn't exist, isn't a FITS file or if the file doesn't contain the WCS data.

wtgee: Yes. This is actually an odd file altogether with how it works although it is useful. Josh did the original and then I cleaned it up a little but I think my cleanup (including things like this setter) made it oddly formed.

PR #396, opened 2018-01-19
Redo the layout of the log files so that there are only symlinks in t…

jamessynge:
…he top-level PANLOG dir, plus a single sub-dir.

### DIFF
--- a/pocs/utils/logger.py
+++ b/pocs/utils/logger.py
@@ -28,10 +28,16 @@ class StrFormatLogRecord(logging.LogRecord):
     def getMessage(self):
         msg = str(self.msg)
         if self.args:
-            try:
-                msg = msg % self.args
-            except (TypeError, ValueError):
-                msg = msg.format(*self.args)
+            if '{' in msg:
+                try:
+                    msg = msg.format(*self.args)
+                except (TypeError, ValueError):
+                    msg = msg % self.args
+            else:
+                try:
+                    msg = msg % self.args
+                except (TypeError, ValueError):
+                    msg = msg.format(*self.args)
         return msg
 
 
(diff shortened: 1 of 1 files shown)


pocs/utils/logger.py
```diff
@@ -28,10 +28,16 @@ class StrFormatLogRecord(logging.LogRecord):
     def getMessage(self):
         msg = str(self.msg)
         if self.args:
-            try:
-                msg = msg % self.args
-            except (TypeError, ValueError):
-                msg = msg.format(*self.args)
+            if '{' in msg:
```
wtgee: Were there cases that weren't being covered?

pocs/utils/logger.py
```diff
@@ -66,32 +72,40 @@ def get_root_logger(profile='panoptes', log_config=None):
...
+        log_fname_datetime = datetime.datetime.now().strftime('%Y%m%dT%H%M%S')
 
     # Setup log file names
     invoked_script = os.path.basename(sys.argv[0])
-    log_dir = '{}/logs'.format(os.getenv('PANDIR', gettempdir()))
-    log_fname = '{}-{}-{}'.format(invoked_script, os.getpid(), log_fname_datetime)
-    log_symlink = '{}/{}.log'.format(log_dir, invoked_script)
+    log_dir = os.getenv('PANLOG', '')
+    if not log_dir:
+        log_dir = os.path.join(os.getenv('PANDIR', gettempdir()), 'logs')
```
wtgee: Why the temp dir here? Seems like you could just put it in `$PANDIR/logs`

pocs/utils/logger.py
```diff
@@ -110,6 +124,8 @@ def get_root_logger(profile='panoptes', log_config=None):
     logger.addFilter(FilenameLineFilter())
 
     logger.info('{:*^80}'.format(' Starting PanLogger '))
+    # TODO(jamessynge) Output name of script, cmdline args, etc. And do son
```
wtgee: "and so on". Not a big deal.

pocs/utils/logger.py
jamessynge: That was there before... It handles the case where PANDIR has not been set (i.e. gettempdir is providing the default value in case PANDIR is unset).

pocs/utils/logger.py
jamessynge: Worth pointing out, thanks.

pocs/utils/logger.py
jamessynge: Yes. IIRC, it was when some non-primitive type (dict, I think) was passed in as the args (e.g. for {!r}), but as a result the % operator didn't complain.

wtgee: Got it. I never use the `{!r}` syntax and don't feel like I've seen the error but it's certainly more robuse.

PR #399, opened 2018-01-20
Adjustments for Power (trucker) Board - WIP

wtgee:
* Update peas_shell to use board if found
* Update power board to turn on all relays to start
* Flush arduino input before sending relay
* Make power board commands consistent with others


bin/peas_shell
```diff
@@ -216,16 +219,35 @@ class PanSensorShell(cmd.Cmd):
...
+            relay_lookup = powerboard_relay_lookup
+        else:
+            relay_lookup = telemetry_relay_lookup
+
         try:
             relay_info = relay_lookup[relay]
-            self.environment.serial_readers[relay_info['board']]['reader'].write("{},9".format(relay_info['pin']))
+            serial_connection = self.environment.serial_readers[relay_info['board']]['reader']
+
+            serial_connection.ser.flushInput()
```
wtgee: @jamessynge I'm curious about your thoughts on this. It seemed to help in my testing yesterday. Before this toggle commands would sometimes lag (10+ seconds).

jamessynge:
Sorry for the delay in reviewing. I'll try to get to it soon, but after the GSoC application.

bin/peas_shell
```diff
@@ -216,16 +219,35 @@ class PanSensorShell(cmd.Cmd):
             'cam_1': {'pin': 6, 'board': 'camera_board'},
         }
 
+        # NOTE: These are not pins but index numbers
```
jamessynge: This is why I'd like to have each board tell us it's set of relays as the names in the relays object in the JSON report.

bin/peas_shell
```diff
@@ -270,7 +292,7 @@ class PanSensorShell(cmd.Cmd):
 
     def do_change_delay(self, *arg):
         """Change the timing between reads from the named sensor."""
-        # NOTE: For at least the Arduinos, we should not need a delay and a timer, but
```
jamessynge: I'll need to change this command later to match the changes I'm making for reading from the Arduinos.

resources/arduino_files/power_board/power_board.ino
```diff
@@ -84,6 +84,14 @@ void setup() {
   
   digitalWrite(DSEL_0, LOW); // DSEL_0 LOW reads PROFET 0_0. DSEL_0 HIGH reades PROFET 0_1
   digitalWrite(DSEL_1, LOW); // DSEL_1 LOW reads PROFET 1_0. DSEL_1 HIGH reades PROFET 1_1  
+
+  // Turn on everything to start
```
jamessynge: Given the lack of pull up resistors, I'd be inclined to turn the relationship on as soon as possible

bin/peas_shell
wtgee: Agreed. I think during the `setup` we could output some info about the board. Or just deliver it along with each report. This means you would have to be good about having the correct plug in each relay but that's probably a good idea anyway.

PR #405, opened 2018-01-21
Fix the exp_time in metadata when passed as a param

### LINKED ISSUE #404: Pointing images have wrong exp_time in metadata info
The time is correct in the FITS headers but is incorrect for the mongo `observations` collection.

wtgee:
Closes #404

### DIFF
--- a/pocs/camera/canon_gphoto2.py
+++ b/pocs/camera/canon_gphoto2.py
@@ -139,7 +139,11 @@ class Camera(AbstractGPhotoCamera):
             'start_time': start_time,
         }
         metadata.update(headers)
+
+        # This can be different than for the headers because passed as param
         exp_time = kwargs.get('exp_time', observation.exp_time.value)
+        metadata['exp_time'] = exp_time
+
         proc = self.take_exposure(seconds=exp_time, filename=file_path)
 
         # Add most recent exposure to list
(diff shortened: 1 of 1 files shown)


pocs/camera/canon_gphoto2.py
```diff
@@ -139,7 +139,11 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
             'start_time': start_time,
         }
         metadata.update(headers)
+
+        # This can be different than for the headers because passed as param
```
jamessynge: Not quite sure how to interpret that comment.

pocs/camera/canon_gphoto2.py
```diff
@@ -139,7 +139,11 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
...
         exp_time = kwargs.get('exp_time', observation.exp_time.value)
```
jamessynge: I'm curious why exp_time=None is not explicitly in the parameter list of the function decl, rather than using kwargs as done here.

pocs/camera/canon_gphoto2.py
wtgee: Yeah, not me best comment. :) Will try to clarify.

pocs/camera/canon_gphoto2.py
wtgee: It probably could be but the method is supposed to just take an `observation`, which defines the exp_time in most scenarios. This could be changed but should happen after #363, which consolidates a lot of the camera methods so we only have to change it in one place.

PR #407, opened 2018-01-22
End of night shutdown improvements.

### LINKED ISSUE #390: End of night shutdown
At the end of the night the scheduler is not finding any valid observations, which is okay, but the system sort of treats this as an error and shuts down. This triggers a  "error" shut down that stops the state machine, meaning it won't come up automatically the next night. Output from logs is here:

```
I0118 15:41:30.891 messaging.py:122     PANCHAT Weather is good and it is dark. Something must have gone wrong. Stopping loop
I0118 15:41:30.892 machine.py:171       Stopping POCS states
```

See also #373

wtgee:
A number of fixes and changes to account for situation when no valid
observations are found but we don't want to shut down the state machine.

* If scheduler has a list of `observations` then it `has_valid_observations`.
	which is True when starting.

Process:
* `scheduling` state calls `observatory.get_observation()`
* `get_observation` first checks if `has_valid_observations is False`
	if so it will reread the field list.
* `get_observation` either returns an observation or, if none found,
	calls `clear_available_observations`, making `has_valid_observations = False`.
* If no observation is returned to `scheduling` state, send to `parking` (then to `sleeping`).
* In `sleeping`, if `has_valid_observations is False` then sleep for an hour
	then try again (or shut down if `run_once` is specified).
  If `has_valid_observations is True` and it is safe and dark, shut down because
  	of unknown error.

Other changes:
* Remove `reread_fields_file` from dispatch scheduler as it is handled at
the `observatory` level by `observatory.get_observation`.
* Misc small fixes

> ⚠️ Note: I think I've gotten all the logic correct so that Bad Things™ won't happen. I think. :)

Closes #390

### DIFF
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -186,8 +186,8 @@ class Observatory(PanBase):
         self.logger.debug("Getting observation for observatory")
 
         # If observation list is empty or a reread is requested
-        if self.scheduler.has_valid_observations is False or \
-                kwargs.get('reread_fields_file', False):
+        if (self.scheduler.has_valid_observations is False or
+                kwargs.get('reread_fields_file', False)):
             self.scheduler.read_field_list()
 
         self.scheduler.get_observation(*args, **kwargs)
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -184,9 +184,16 @@ class Observatory(PanBase):
         """
 
         self.logger.debug("Getting observation for observatory")
+
+        # If observation list is empty or a reread is requested
+        if self.scheduler.has_valid_observations is False or \
+                kwargs.get('reread_fields_file', False):
+            self.scheduler.read_field_list()
+
         self.scheduler.get_observation(*args, **kwargs)
 
         if self.scheduler.current_observation is None:
+            self.scheduler.clear_available_observations()
             raise error.NoObservation("No valid observations found")
 
         return self.current_observation
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -190,9 +190,10 @@ class Observatory(PanBase):
                 kwargs.get('reread_fields_file', False)):
             self.scheduler.read_field_list()
 
+        # This will set the `current_observation`
         self.scheduler.get_observation(*args, **kwargs)
 
-        if self.scheduler.current_observation is None:
+        if self.current_observation is None:
             self.scheduler.clear_available_observations()
             raise error.NoObservation("No valid observations found")
 
--- a/pocs/state/machine.py
+++ b/pocs/state/machine.py
@@ -121,6 +121,7 @@ class PanStateMachine(Machine):
 
             # If we are processing the states
             if self.do_states:
+
                 # If sleeping, wait until safe (or interrupt)
                 if self.state == 'sleeping':
                     if self.is_safe() is not True:
@@ -129,7 +130,8 @@ class PanStateMachine(Machine):
                 try:
                     state_changed = self.goto_next_state()
                 except Exception as e:
-                    self.logger.warning("Problem going to next state, exiting loop [{}]".format(e))
+                    self.logger.warning("Problem going from {} to {}, exiting loop [{!r}]".format(
+                        self.state, self.next_state, e))
                     self.stop_states()
                     break
 
@@ -143,6 +145,14 @@ class PanStateMachine(Machine):
                         self.sleep(with_status=False)
                 else:
                     _loop_iteration = 0
+
+                ########################################################
+                # Note that `self.state` below has changed from above
+                ########################################################
+
+                # If we are in ready state then we are making one attempt
+                if self.state == 'ready':
+                    self._nightly_retries -= 1
 
                 if self.state == 'sleeping' and self.run_once:
                     self.stop_states()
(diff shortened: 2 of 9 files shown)


pocs/state/states/default/sleeping.py
```diff
@@ -4,11 +4,17 @@ def on_enter(event_data):
     pocs.next_state = 'ready'
 
     # If it is dark and safe we shouldn't be in sleeping state
-    if pocs.is_dark() and pocs.is_safe():
+    if pocs.observatory.scheduler.has_valid_observations and \
```
jamessynge: pep8 strongly prefers wrapping with parens around expressions rather than using a backslash.

pocs/state/states/default/sleeping.py
```diff
@@ -4,11 +4,17 @@ def on_enter(event_data):
...
+            pocs.is_dark() and pocs.is_safe():
         if pocs.should_retry is False:
```
jamessynge: Property should_retry is very problematic. It looks like a property, but accessing it modifies its value!

pocs/tests/test_dispatch_scheduler.py
```diff
@@ -119,11 +119,12 @@ def test_get_observation_reread(field_list, observer, temp_file, constraints):
         f.write(yaml.dump([{
             'name': 'New Name',
             'position': '20h00m43.7135s +22d42m39.0645s',
-            'priority': 50
+            'priority': 5000
```
jamessynge: Out of curiosity, why the change in priority?

pocs/observatory.py
```diff
@@ -184,9 +184,16 @@ def get_observation(self, *args, **kwargs):
         """
 
         self.logger.debug("Getting observation for observatory")
+
+        # If observation list is empty or a reread is requested
+        if self.scheduler.has_valid_observations is False or \
```
jamessynge: pep8 strongly prefers wrapping with parens around expressions rather than using a backslash.

pocs/state/states/default/sleeping.py
wtgee: Yes, thanks for calling this out. It's also problematic in that it is never reset without stopping and starting whole system. So if it retries once per night by the third night it is out of retry attempts no matter what. Will fix.

wtgee:
I've removed a whole test in `test_pocs.py` that was hanging. Looking at the test I can't figure out how it was any different from the test below it. Based off of the name of the test, it was designed to be testing some of the interrupt code that a student added about a year ago (via Huntsman) and which has been removed during repository merge and cleanup because it needs some independent work.

Feel weird about just removing a test but I think it is the right course of action.

pocs/observatory.py
```diff
@@ -186,8 +186,8 @@ def get_observation(self, *args, **kwargs):
         self.logger.debug("Getting observation for observatory")
 
         # If observation list is empty or a reread is requested
-        if self.scheduler.has_valid_observations is False or \
-                kwargs.get('reread_fields_file', False):
+        if (self.scheduler.has_valid_observations is False or
+                kwargs.get('reread_fields_file', False)):
             self.scheduler.read_field_list()
 
         self.scheduler.get_observation(*args, **kwargs)
```
jamessynge: This looks like it should return something. Why not return what get_observation returns?

pocs/state/machine.py
```diff
@@ -129,7 +130,8 @@ def run(self, exit_when_done=False, run_once=False):
                 try:
                     state_changed = self.goto_next_state()
                 except Exception as e:
-                    self.logger.warning("Problem going to next state, exiting loop [{}]".format(e))
+                    self.logger.warning("Problem going from {} to {}, exiting loop [{!r}]".format(
```
jamessynge: Fine for now, but it would be nice to have a recovery mechanism, so that the system automatically tries to get to mount-parked/dome-closed/state-sleeping if things go bad, and then resumes... or waits until the next noon to restart.

pocs/state/states/default/parked.py
```diff
@@ -4,5 +4,25 @@ def on_enter(event_data):
     pocs = event_data.model
     pocs.say("I'm parked now. Phew.")
 
-    pocs.say("Cleaning up for the night!")
-    pocs.next_state = 'housekeeping'
+    has_valid_observations = pocs.observatory.scheduler.has_valid_observations
+
+    if (has_valid_observations and pocs.is_safe() is False):
```
jamessynge: Parentheses not needed here.

pocs/state/states/default/parked.py
```diff
@@ -4,5 +4,25 @@ def on_enter(event_data):
...
+        pocs.say("Cleaning up for the night!")
+        pocs.next_state = 'housekeeping'
+    elif (has_valid_observations and pocs.is_safe()):
```
jamessynge: Excess parens.

pocs/state/states/default/parked.py
jamessynge: Perhaps:
```python
if has_valid_observations:
  if not pocs.is_safe():
    ...
  else:  # pocs.is_safe
    ...
else: # not has_valid_observations
  ...
```

pocs/observatory.py
wtgee: Good idea to make it explicit, but `get_observation` actually also sets `current_observation`, which is then returned. Should maybe think about renaming this.

pocs/state/machine.py
wtgee: The `state_changed` is `False` it will do as you describe, which should be what happens under normal operations. We should probably do as you describe although I would want to do more testing and focus on it. Created #419 to address.

Issue #120, opened 2017-11-18
Decouple PanMongo from mongo (and rename to PanDB)


wtgee:
We shouldn't be tied to mongo for storing and getting info.

wtgee:
Closed by #414. Specific items should be opened as new Issues.

PR #408, opened 2018-01-22
DB access consistency

wtgee:
In preparation for #120, this is brining some consistency to the access
of DB items so it will be easier to decouple from mongo if needed.


bin/peas_shell
```diff
@@ -44,15 +43,19 @@ class PanSensorShell(cmd.Cmd):
     def do_status(self, *arg):
         """ Get the entire system status and print it pretty like! """
         if self._keep_looping:
-            console.color_print("{:>12s}: ".format('Loop Timer'), "default", "active", "lightgreen")
+            console.color_print("{:>12s}: ".format('Loop Timer'),
```
jamessynge: Nothing changed except the wrapping, right?
If so, why was this changed, as it was 100 chars wide, which I thought we'd agreed on.

pocs/utils/database.py
```diff
@@ -78,16 +78,18 @@ def __init__(self, db='panoptes', host='localhost', port=27017, connect=False):
             setattr(self, collection, getattr(db_handle, collection))
 
     def insert_current(self, collection, obj, include_collection=True):
-        """Insert an object into both the `current` collection and the collection provided
+        """Insert an object into both the `current` collection and the collection provided.
```
jamessynge: Thanks for adding the periods! :wink:

pocs/utils/database.py
```diff
@@ -101,14 +103,46 @@ def insert_current(self, collection, obj, include_collection=True):
...
+
+        Returns:
+            str: Mongo object ID of record in `collection`.
+        """
+        assert collection in self.collections, warn("Collection not available")
+
+        _id = None
+        try:
+            # If `data` key is present we assume it has "metadata" for object
+            if 'data' in obj:
```
jamessynge: if isinstance(obj, dict) and 'data' in obj:

pocs/utils/database.py
```diff
@@ -101,14 +103,46 @@ def insert_current(self, collection, obj, include_collection=True):
...
+            obj (dict or str): Object to be inserted.
+
+        Returns:
+            str: Mongo object ID of record in `collection`.
+        """
+        assert collection in self.collections, warn("Collection not available")
+
+        _id = None
+        try:
+            # If `data` key is present we assume it has "metadata" for object
```
jamessynge: What does this mean: it has "metadata" for object
What does it refer to there?
And what is the metadata?

pocs/utils/database.py
```diff
@@ -101,14 +103,46 @@ def insert_current(self, collection, obj, include_collection=True):
...
+                    'type': collection,
+                    'data': obj,
+                    'date': current_time(datetime=True),
+                }
+
+            # Insert record into db
+            col = getattr(self, collection)
+            _id = col.insert_one(obj).inserted_id
         except Exception as e:
             warn("Problem inserting object into collection: {}".format(e))
```
jamessynge: Could you add an additional message with the obj to be inserted?
And note that this is not written to the log file, which would really be useful!

pocs/tests/test_database.py
```diff
@@ -34,8 +34,29 @@ def test_insert_and_no_collection(db):
     db.current.remove({'type': 'config'})
 
 
+def test_simple_insert(db):
+    rec = {'test': 'insert'}
+    db.insert('config', rec)
+
+    record = db.config.find({'data.test': {'$exists': 1}})
```
jamessynge: This seems very mongo specific. Are you not attempting to address that ... yet?

pocs/utils/database.py
```diff
@@ -101,14 +103,46 @@ def insert_current(self, collection, obj, include_collection=True):
...
+
+    def insert(self, collection, obj):
+        """Insert an object into the collection provided.
+
+        Args:
+            collection (str): Name of valid collection within panoptes db.
+            obj (dict or str): Object to be inserted.
+
+        Returns:
+            str: Mongo object ID of record in `collection`.
```
jamessynge: Error behavior?

bin/peas_shell
wtgee: 100 char limit means it wraps at 99, which this line is. From the [pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length):
```
For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters)
```

pocs/tests/test_database.py
wtgee: Correct. I was thinking of actually renaming to `test_mongo.py`, which could be part of this PR. Actually addressing anything non-mongo is for another PR.

pocs/utils/database.py
wtgee: There currently isn't any error that is raised as we just catch the exceptions and give a warning. Not sure if that is ideal or not but it does allow operations to continue.

pocs/tests/test_database.py
wtgee: Renamed to `test_panmongo.py` for now.

pocs/utils/database.py
wtgee: Fixed this so now we have a logger in PanMongo and warnings will go to appropriate place.

PR #409, opened 2018-01-23
Export mongo to datastore

wtgee:
Tool for sending observation information up to Datastore in bulk.

Note: I will start a discussion as to exact format of Keys/Entities and that should be agreed upon before this is used in mass. However, I have used it to send all of PAN001 information to Datastore.

Note: this should be used for sending items that have not been stored already. The next PR will involve sending of observations as they are record so we will have real-time information.

Note: This **does not** send the entity for the Unit as that should be part of a registration script. However, the unit as parent key will still work, just missing lat/long information about unit.

Edited: See also https://github.com/orgs/panoptes/teams/cloud-devs/discussions/1

### DIFF
--- a/scripts/data/observations_to_datastore.py
+++ b/scripts/data/observations_to_datastore.py
@@ -10,23 +10,30 @@ from pocs.utils import current_time
 
 
 class ObservatonsExporter(PanBase):
-    """ A simple class for exporting observations to Google Datastore
 
-    Observations are pulled from the corresponding mongo collection.
-    """
+    def __init__(self, start_date=None, project_id='panoptes-survey'):
+        """A simple class for exporting observations to Google Datastore
 
-    def __init__(self, date=None, project_id='panoptes-survey'):
-        super(ObservatonsExporter, self).__init__()
+        Observations are pulled from the corresponding mongo collection.
+
+        Args:
+            start_date (None|str, optional): Observations later than provided date
+                will be exported. Defaults to None, in which case the previous
+                date will be used. If provided, should be str, e.g. `2018-01-01`
+            project_id (str, optional): The project associated with the datastore.
+                Defaults to 'panoptes-survey'
+        """
+        super().__init__()
         self.ds_client = datastore.Client(project_id)
 
-        if date is None:
-            date = (current_time() - 1. * u.day).datetime
+        if start_date is None:
+            start_date = (current_time() - 1. * u.day).datetime
         else:
-            date = Time(date).datetime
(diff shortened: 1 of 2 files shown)


scripts/data/observations_to_datastore.py
```diff
@@ -0,0 +1,206 @@
...
+
+if __name__ == '__main__':
+
+    import argparse
+
+    parser = argparse.ArgumentParser(
+        description="Export observations information to datastore")
+    parser.add_argument('--date', default=None,
+                        help='Export start date, e.g. 2016-01-01, defaults to yesterday')
+    parser.add_argument('--auto-confirm', action='store_true', default=False,
```
jamessynge: --skip-confirmation?

scripts/data/observations_to_datastore.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class ObservatonsExporter(PanBase):
+    """ A simple class for exporting observations to Google Datastore
+
+    Observations are pulled from the corresponding mongo collection.
+    """
+
+    def __init__(self, date=None, project_id='panoptes-survey'):
+        super(ObservatonsExporter, self).__init__()
```
jamessynge: super().__init__() according to pep-3135.

https://www.python.org/dev/peps/pep-3135/

scripts/data/observations_to_datastore.py
```diff
@@ -0,0 +1,206 @@
...
+from google.cloud import datastore
+
+from pocs.base import PanBase
+from pocs.utils import current_time
+
+
+class ObservatonsExporter(PanBase):
+    """ A simple class for exporting observations to Google Datastore
+
+    Observations are pulled from the corresponding mongo collection.
```
jamessynge: Perhaps add a bit more theory of operation here, or more comments throughout? I can't immediately see how it works (well, I get the basic idea, but not enough).

wtgee: This was mostly moving my experimenting into something more permanent. I am going to start a team Discussion to see if the format is what we want. Also, I want to make a PR today that will put the records into Datastore in real-time, which should be much more obvious. I hesitated about even pushing this up.

scripts/data/observations_to_datastore.py
wtgee: My editor has built-in templates that pre-populate this and let me tab through the options to change and it maintains old style for compatibility. But I'll get this changed, thanks.

wtgee:
Closing for now as we might not want this as our solution.

PR #412, opened 2018-01-23
Small updates to simple helper script.

wtgee:
Currently using this on the Pi where I don't have a mongo installation

### DIFF
--- a/scripts/simple_sensors_capture.py
+++ b/scripts/simple_sensors_capture.py
@@ -1,22 +1,28 @@
 import time
+from bson import json_util
 
 from peas.sensors import ArduinoSerialMonitor
 
 
-def main(loop=True, delay=1., verbose=False):
+def main(loop=True, delay=1., filename=None, use_mongo=True, send_message=True, verbose=False):
     # Weather object
     monitor = ArduinoSerialMonitor(auto_detect=False)
 
     while True:
-        data = monitor.capture()
+        data = monitor.capture(use_mongo=use_mongo, send_message=send_message)
 
-        if verbose and len(data.keys()) > 0:
-            print(data)
+        if len(data.keys()) > 0:
+            if filename is not None:
+                with open(filename, 'a') as f:
+                    f.write(json_util.dumps(data) + '\n')
+
+            if verbose:
+                print(data)
 
         if not args.loop:
             break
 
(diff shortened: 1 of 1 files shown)


scripts/simple_sensors_capture.py
```diff
@@ -1,22 +1,28 @@
 import time
+from bson import json_util
```
jamessynge: Why not Python's json library?

scripts/simple_sensors_capture.py
```diff
@@ -1,22 +1,28 @@
...
 
     while True:
-        data = monitor.capture()
+        data = monitor.capture(use_mongo=use_mongo, send_message=send_message)
 
-        if verbose and len(data.keys()) > 0:
-            print(data)
+        if len(data.keys()) > 0:
+            if filename is not None:
+                with open(filename, 'a') as f:
```
jamessynge: Seems odd to open the file on each capture. How about open once and flush after each write or in a finally block so that it happens even if there is an exception?

scripts/simple_sensors_capture.py
wtgee: I've been using BSON as it handles some object types better and is "supposed" to be faster while still fully compatible. It's the underlying encoder for mongo. Not sure it's needed here so could change.

wtgee: Specifically, it handles `datetime` encoding without any effort.

jamessynge: Can we use bson without installing mongodb?

And does it still produce JSON (text), rather than BSON (binary).

wtgee: The `json_util.dumps` does produce text.

```
{'bar': 42, 'dt': datetime.now(), 'baz': 'bam'}
```
to
```
{"bar": 42, "dt": {"$date": 1516800132526}, "baz": "bam"}
```
Under the hood it is storing the binary representation of `42` as opposed to the string `'42'`, like json. This means the files can actually be larger.

scripts/simple_sensors_capture.py
wtgee: Good call, will change.

wtgee:
@jamessynge Will you take a look at how I implemented the file writing? Thanks.

Issue #268, opened 2017-12-29
Add config display to shells


jamessynge:
pocs_shell & peas_shell should have commands to display the config that is loaded.

jermainegug:
Would something like this work?
```
    def do_display_config(self, config):
        """ Display the config file for <pocs/peas> """
        if config is None:
            print_error("No config file for <pocs/peas> has been loaded.")
        else:
            print(config)
```

jermainegug:
Will do the same for pocs_shell and will open a PR.

wtgee:
Added in #417, thanks @jermainegug !

PR #417, opened 2018-01-25
Add display config to shells (#268)

jermainegug:
This pull request solves issue #268. Potentially could add options to display specific parts of the config file.

### DIFF
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -8,6 +8,7 @@ import sys
 from astropy.utils import console
 from pprint import pprint
 from threading import Timer
+from pprint import pprint
 
 from peas.sensors import ArduinoSerialMonitor
 from peas.weather import AAGCloudSensor
@@ -122,6 +123,13 @@ class PanSensorShell(cmd.Cmd):
                 log.setLevel(get_level[log.getEffectiveLevel()])
             except Exception as e:
                 print_error("Can't change log level for {}".format(sensor))
+
+    def do_display_config(self, config):
+        """ Display the config file for PEAS """
+        if self.config is None:
+            print_warning("No config file for PEAS has been loaded.")
+        else:
+            pprint(self.config)
 
 ##################################################################################################
 # Load Methods
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -6,7 +6,6 @@ import readline
 import sys
 
 from astropy.utils import console
-from pprint import pprint
 from threading import Timer
 from pprint import pprint
 
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -127,9 +127,9 @@ class PanSensorShell(cmd.Cmd):
     def do_display_config(self, config):
         """ Display the config file for PEAS """
         if self.config:
-            print_warning("No config file for PEAS.")
-        else:
             pprint(self.config)
+        else:
+            print_warning("No config file for PEAS.")
 
 ##################################################################################################
 # Load Methods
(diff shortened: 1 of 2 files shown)


wtgee:
@jamessynge you added this issue, anything else you had in mind?

bin/pocs_shell
```diff
@@ -7,6 +7,7 @@ import zmq
 from cmd import Cmd
 from subprocess import TimeoutExpired
 from pprint import pprint
+from pprint import pprint
```
wtgee: Looks like this is already there?

bin/pocs_shell
```diff
@@ -74,6 +74,13 @@ class PocsShell(Cmd):
 
         return self.pocs.is_safe()
 
+    def do_display_config(self, *arg):
+        """ Display the config file for POCS """
+        if self.pocs.config is None:
```
jamessynge: What if self.pocs is not set (i.e. setup_pocs has not been run)? Seems like it would throw an exception.

jamessynge:
Thanks very much for this change, I look forward to using it.

bin/pocs_shell
wtgee: @jermainegug Can you check on the above?

bin/peas_shell
```diff
@@ -123,6 +124,13 @@ class PanSensorShell(cmd.Cmd):
             except Exception as e:
                 print_error("Can't change log level for {}".format(sensor))
 
+    def do_display_config(self, config):
+        """ Display the config file for PEAS """
+        if self.config=={}:
```
wtgee: The way to check for this would be to just say `if self.config`:

bin/pocs_shell
```diff
@@ -74,6 +74,17 @@ class PocsShell(Cmd):
 
         return self.pocs.is_safe()
 
+    def do_display_config(self, *arg):
+        """ Display the config file for POCS """
+        try:
+            if self.pocs.config=={}:
```
wtgee: Same here where you just want to check `if self.pocs.config:`, but I think the point James was making was that if you have not run `setup_pocs` then `self.pocs` will be `None` and the `self.pocs.config` will generate an error. So you first need to check for `pocs` and then check for `config`, i.e. `if self.pocs and self.pocs.config:`

bin/pocs_shell
```diff
@@ -74,6 +74,17 @@ class PocsShell(Cmd):
 
         return self.pocs.is_safe()
 
+    def do_display_config(self, *arg):
+        """ Display the config file for POCS """
+        try:
+            if self.pocs and self.pocs.config:
+                print_warning("No config file for POCS.")
+            else:
```
jamessynge: I think the if and else cases have been swapped.

jermainegug: Seems like I did it for both. Fixing now, thank you.

jamessynge:
Thanks for the feature and fixes.

jermainegug:
I am not familiar with how the `codecov/project` check works, so not sure how to get this check to pass.

wtgee:
The codecov failures don't absolutely prevent us from merging although if they are way off you should pay attention to them. The reports are currently showing some odd things though depending on when it was branched. Don't worry about them for now.

bin/peas_shell
```diff
@@ -8,6 +8,7 @@ import sys
 from astropy.utils import console
 from pprint import pprint
 from threading import Timer
+from pprint import pprint
```
wtgee: This still looks like a duplicate of line 9.

PR #418, opened 2018-01-26
Eliminate circular dependencies in pocs package

jamessynge:
`pocs/__init__.py` imported pocs.core.POCS, which imported
PanBase from pocs, which meant that we had a circular
dependency which just barely loaded (i.e. small changes
could cause it to fail to load).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,7 @@ instead of `My File.py`.
 - Use root-relative imports (i.e. relative to the POCS directory). This means that rather
   than using a directory relative imports such as:
   ```python
-  from .. import PanBase
+  from ..base import PanBase
   from ..utils import current_time
   ```
   Import from the top-down instead:
--- a/bin/imiloa
+++ b/bin/imiloa
@@ -9,7 +9,7 @@ import time
 
 from astropy import units as u
 
-from pocs import POCS
+from pocs.core import POCS
 
 # Constants
 park_ha = 290 * u.deg
--- a/bin/pocs
+++ b/bin/pocs
@@ -6,7 +6,7 @@ import sys
 
 from warnings import warn
 
-from pocs import POCS
+from pocs.core import POCS
 
 
 def main():
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -14,8 +14,8 @@ from astropy.coordinates import ICRS
 from astropy.io import fits
 from astropy.utils import console
 
-from pocs import POCS
 from pocs import hardware
+from pocs.core import POCS
 from pocs.observatory import Observatory
 from pocs.scheduler.field import Field
 from pocs.scheduler.observation import Observation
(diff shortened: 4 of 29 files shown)


wtgee:
I would prefer to be able to do `from pocs import POCS`, which this change will prevent. This has presented a problem in a few situations however should be working. What changes were you doing that caused a failure?

jamessynge:
Why would you prefer `from pocs import POCS`? I think there were only 5 occurrences of that, so we wouldn't be saving anything.

I was working on resolving type annotation warnings from mypy, which apparently doesn't recognize that a module has a symbol in it that has been imported rather than locally assigned/defined. In addition, I was investigating why with certain changes `python setup.py develop` or `python setup.py install` would fail because `pocs.__version__` couldn't be imported due to circular import failures.

This is actually my third branch dealing with this issue, but the first I've turned into a PR. In one branch I attempted explain the situation in `pocs/__init__.py`, and also fix the mypy issue:

```python
# NOTES for editors of this file:
#
# 1) We have essentially circular dependencies here, which are rather delicate (i.e. it is easy to
#    create un importable modules). It would be best to avoid  directly importing these symbols
#    (e.g. NOT 'from pocs import POCS', but instead 'from pocs.core import POCS').
#
# 2) These imports are carefully ordered to ensure that we can successfully import POCS.
#
# 3) To support static type analysis by mypy, we explicitly add symbols that we could just
#    add with an 'from X import Y' statement.

from . import version
__version__ = version.__version__

from . import base
PanBase = base.PanBase

from . import core  # nopep8
POCS = core.POCS
```

wtgee:
Right, I've dealt with a it a bunch too. I prefer the direct import because I develop interactively quite frequently and that is also how I often debug so while it appears in the code only 5 times I end up typing it into the console multiple times per day. But that's actually been reduced now that we do dependency injection with the Observatory as I often just use it direclty. I also do think it is more clear without having to use the `.core` but if it is giving us problems we should change.

However, keep in mind we have not resolved to use the type hinting. As I mentioned in the Discussion I am not convinced making the code more obtuse and requiring a 3rd tool for it to be useful (thus further complicating development for beginners) is worth the fairly limited benefit I see to it at this point. I understand the benefits of type hinting I just don't see it as such a huge issue in our code that it merits the changes. What I do see is many hours spent trying to explain this and how to use it every time a new student is trying to help us with our code. So if you are making a lot of code changes for this we need to make that decision first.

jamessynge:
Oops, I searched for .py files, forgot about the shells and notebooks.

jamessynge:
I agree that the type annotations seem to be an excessive change. I think I'd prefer to simply implement some parts of the system in a language with strict typing.

Nonetheless, I'd prefer to get rid of the the circular imports issue, to improve the robustness of the code.

wtgee:
yes, sounds good. I just didn't want you putting a ton of work into it for me to reject it. :) I think for some parts it makes sense and if you are doing it to learn about it feel free.

And agreed about getting this change in.

jamessynge:
PTAL

jamessynge:
Thanks.

PR #421, opened 2018-01-28
Adjust Shutdown States

wtgee:
Change the routing on the shutdown states so that the `housekeeping` state is only entered when actually shutting down for the day. This means the unit can wait in the `parked` state if no immediate observations are found but if otherwise safe. If it's daytime after waiting then go to `housekeeping`. If it's bad weather wait another 30 minutes. Otherwise try to observe again.

* Add route from `parked` to `ready`
* Remove route from `housekeeping` to `ready`
* Return value for `power_down`

### DIFF
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -286,9 +286,6 @@ class POCS(PanStateMachine, PanBase):
 
         Returns:
             bool: Latest safety flag
-
-        Deleted Parameters:
-            event_data(transitions.EventData): carries information about the event if
         """
         if not self.connected:
             return False
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -290,6 +290,9 @@ class POCS(PanStateMachine, PanBase):
         Deleted Parameters:
             event_data(transitions.EventData): carries information about the event if
         """
+        if not self.connected:
+            return False
+
         is_safe_values = dict()
 
         # Check if night time
@@ -421,9 +424,13 @@ class POCS(PanStateMachine, PanBase):
         # If delay is greater than 10 seconds check for messages during wait
         if delay >= 10.0:
             while delay >= 10.0:
+                self.check_messages()
+                # If we shutdown leave loop
+                if self.connected is False:
+                    return
+
                 time.sleep(10.0)
                 delay -= 10.0
-                self.check_messages()
 
         if delay > 0.0:
             time.sleep(delay)
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -282,10 +282,12 @@ class POCS(PanStateMachine, PanBase):
             This condition is called by the state machine during each transition
 
         Args:
-            called from the state machine.
+            no_warning (bool, optional): If a warning message should show in logs,
+                defaults to False.
 
         Returns:
             bool: Latest safety flag
+
         """
         if not self.connected:
             return False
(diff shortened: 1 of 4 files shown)


pocs/core.py
```diff
@@ -263,6 +263,8 @@ def power_down(self):
             self._connected = False
             self.logger.info("Power down complete")
 
+        return self.connected is False
```
jamessynge: Why return this value. It looks like it can not ever return False.

pocs/tests/test_pocs.py
```diff
@@ -343,3 +343,27 @@ def start_pocs():
...
+    pocs.next_state = 'parking'
+    assert pocs.goto_next_state()
+    assert pocs.observatory.current_observation is None
+    assert pocs.observatory.mount.is_parked
+    assert pocs.goto_next_state()
+    # Should be safe and still have valid observations so next state should
+    # be ready
+    assert pocs.goto_next_state()
+    assert pocs.state == 'ready'
+    assert pocs.power_down()
```
jamessynge: If you remove the `POCS.power_down()` change, perhaps change this to:
```python
    pocs.power_down()
    assert not pocs.connected
```

wtgee:
@jamessynge Adjusted the PR message for explanation but also changed a bit of the logic and added a better test, PTAL.

pocs/state/states/default/parked.py
```diff
@@ -11,19 +11,35 @@ def on_enter(event_data):
...
+            else:  # This branch will only happen if there is an error causing a shutdown
                 pocs.say("Things look okay for now. I'm going to try again.")
                 pocs.next_state = 'ready'
-        else:
+        else:  # Normal end of night
             pocs.say("Cleaning up for the night!")
             pocs.next_state = 'housekeeping'
     else:
+        pocs.say("No observations found.")
+        # TODO Should check if we are close to morning and if so do some morning
```
jamessynge: Oh no, I've infected you with adding TODOs! ;-)

I'd love a similar behavior at the start of the evening (i.e. evening flats, or darks while the dome is closed).

pocs/state/states/default/parked.py
```diff
@@ -11,19 +11,35 @@ def on_enter(event_data):
...
-            pocs.say("No observations found. Going to stay parked for an hour then try again.")
-            pocs.sleep(delay=3600)  # 1 hour = 3600 seconds
+            pocs.say("Going to stay parked for half an hour then will try again.")
+
+            while True:
+                pocs.sleep(delay=1800)  # 30 minutes = 1800 seconds
 
-            pocs.reset_observing_run()
-            pocs.next_state = 'ready'
+                # We might have shutdown in long wait
```
jamessynge: "we might have shutdown during that sleep."

pocs/tests/test_pocs.py
```diff
@@ -343,3 +343,80 @@ def start_pocs():
 
     pocs_process.join()
     assert pocs_process.is_alive() is False
+
+
+def test_pocs_park_to_ready_with_obs(pocs):
```
jamessynge: I'm worried that some of these states are getting too complex. Let's have a further discussion about how to address that (e.g. multiple state machines, either nested or parallel).

wtgee: Yes, this made the `parked` state a bit obtuse with all the `if`-branching.

PR #423, opened 2018-01-28
Remove explicit call to backend

wtgee:
Having it specified just in this file is not the appropriate solution and generates large warnings whenever an image utility is used within a jupyter notebook.

Edit: A matplotlibrc file has been added and can be used via `MPLBACKEND=$POCS/.matplotlibrc`

### DIFF
--- a/.matplotlib/matplotlibrc
+++ b/.matplotlib/matplotlibrc
@@ -1,0 +1,1 @@
+backend: Agg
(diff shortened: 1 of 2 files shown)


jamessynge:
The build hung and was killed, so I restarted it.

wtgee:
Actually, I'm now remembering that travis will hang/fail in a non-obvious without something like this. I'm going to see if I can add it in via travis.

PR #425, opened 2018-01-28
Fix text for run_pocs command

jamessynge:


wtgee:
Is there more coming for this? The title doesn't really match what it does.

jamessynge:
The title was wrong by one critical letter: teXt, not teSt. Sigh.

PR #426, opened 2018-01-29
Adding default env vars file

wtgee:
Removing old unused file

I'm actually not sure we want the MPLBACKEND in there as it will change global settings. Makes sense for a unit on a computer but might annoy some future developer if not careful.


scripts/install/default-env-vars.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+export PANDIR=/var/panoptes
+export PANLOG=${PANDIR}/logs
+export POCS=${PANDIR}/POCS
+export PAWS=${PANDIR}/PAWS
+export PIAA=${PANDIR}/PIAA
+export MPLBACKEND=${POCS}/.matplotlib
```
jamessynge: I'm unfamiliar with this file. I don't see a .matplotlib in my clone.

scripts/setup-env.sh
```diff
@@ -1,28 +0,0 @@
-#!/usr/bin/env bash
```
jamessynge: I think we want a file that includes these actions. Ideally a file like that used by Anaconda (?) where one executes wget to fetch the file and passes it to bash.

scripts/install/default-env-vars.sh
wtgee: Sorry, meant to make a comment that this should go after #423, where I added that.

scripts/setup-env.sh
wtgee: Yes, I think so. We still need to clean up the install process to make it simpler. I still want a docker image that would work well actually. But this is just  small step as I realized a) we don't have those env variables collected anywhere and b) that other script was outdated and should be removed.

PR #430, opened 2018-01-31
Fix default envs

wtgee:
So this was the fix I actually wanted, where the env var points to the file location rather than specifying a backend.


scripts/install/default-env-vars.sh
```diff
@@ -5,4 +5,4 @@ export PANLOG=${PANDIR}/logs
 export POCS=${PANDIR}/POCS
 export PAWS=${PANDIR}/PAWS
 export PIAA=${PANDIR}/PIAA
-export MPLBACKEND=${POCS}/.matplotlib
\ No newline at end of file
+export MATPLOTLIBRC=${POCS}/.matplotlib
```
jamessynge: There is no such file in POCS. Do you have one?

Do you want to rename `POCS/examples/notebooks/matplotlibrc` to `POCS/matplotlibrc` (not `POCS/.matplotlib`)?

wtgee: The env var can point to a file or a directory and that directory should now exist: https://github.com/panoptes/POCS/tree/develop/.matplotlib

jamessynge: My apologies, I was looking for a file in the github UI, and not a directory, which are grouped separately from files. Sigh.

jamessynge: We should perhaps set this env var in conftest.py (in a future PR). I'd like to have fewer rather than more env vars.

wtgee:
As per #431, I'm happy to remove the env var for MATPLOTLIBRC as I don't think it's an ideal fix for reasons mentioned.

Also as per #431, this file can either just be a convenient reference or could be integrated into some kind of install script.

wtgee:
I just realized that the PANOPTES_CLOUD_KEY is also not listed here, which would be yet another one. :) It could be discoverable if we make some assumptions, i.e. That `$PANDIR/.key` is location and the file is `panoptes-survey-XXXX.json`.

jamessynge:
If we assume this is to be called from the user's profile directory (or from /etc/profile), then I think using the approach of checking for pre-existing values is a good idea as it allows the user to specify a subset of values first (e.g. just PANDIR), then calling the script you've got here.

I'd also agree with removing MATPLOTLIBRC from this. I'm not sure about PANOPTES_CLOUD_KEY. I can just as easily see using a field in pocs_local.yaml.

wtgee:
I've added the checks and remove matplotlibrc env var. I'll actually deal with matplotlibrc (probably by just moving the file into $POCS) in a separate PR.

scripts/install/default-env-vars.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 
-export PANDIR=/var/panoptes
-export PANLOG=${PANDIR}/logs
-export POCS=${PANDIR}/POCS
-export PAWS=${PANDIR}/PAWS
-export PIAA=${PANDIR}/PIAA
```
jamessynge: Do we not want/need PIAA?

jamessynge:
Thanks for the changes.

scripts/install/default-env-vars.sh
wtgee: No, PIAA is going to change soon and I'm not sure we need it (at least for now).

PR #433, opened 2018-02-02
Disable some tests when hardware is present

jamessynge:
Add option --without-hardware, complementing --with-hardware.
This allows the user to specify that tests are to be skipped
if they fail with hardware being present or if the user doesn't
want the hardware to be interacted with (e.g. when running
tests on a real unit).

Move parts of POCS/pocs/tests/conftest.py to POCS/conftest.py
in order to enable parsing of command line options without
having to specify pocs/tests on the commad line.

Mark 4 tests in test_observatory as disabled if hardware is present.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,97 @@
+# This is in the root POCS directory so that pytest will recognize the
+# options added below without having to also specify pocs/test, or a
+# one of the tests in that directory, on the command line; i.e. pytest
+# doesn't load pocs/tests/conftest.py until after it has searched for
+# tests.
+
+import os
+import pytest
+
+from pocs import hardware
+
+
+def pytest_addoption(parser):
+    hw_names = ",".join(hardware.get_all_names()) + ' (or all for all hardware)'
+    group = parser.getgroup("PANOPTES pytest options")
+    group.addoption(
+        "--with-hardware",
+        nargs='+',
+        default=[],
+        help=("A comma separated list of hardware to test. " + "List items can include: " +
+              hw_names))
+    group.addoption(
+        "--without-hardware",
+        nargs='+',
+        default=[],
+        help=("A comma separated list of hardware to NOT test. " + "List items can include: " +
+              hw_names))
+    group.addoption(
+        "--solve",
+        action="store_true",
+        default=False,
+        help="If tests that require solving should be run")
+    group.addoption(
+        "--test-cloud-storage",
+        action="store_true",
+        default=False,
+        dest="test_cloud_storage",
+        help="Tests cloud strorage functions." +
+        "Requires $PANOPTES_CLOUD_KEY to be set to path of valid json service key")
+
+
+def pytest_collection_modifyitems(config, items):
+    """Modify tests to skip or not based on cli options.
+
+    Certain tests should only be run when the appropriate hardware is attached.
+    Other tests fail if real hardware is attached (e.g. they expect there is no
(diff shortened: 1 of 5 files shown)


conftest.py
```diff
@@ -0,0 +1,97 @@
+# This is in the root POCS directory so that pytest will recognize the
+# options added below without having to also specify pocs/test, or a
+# one of the tests in that directory, on the command line; i.e. pytest
+# doesn't load pocs/tests/conftest.py until after it has searched for
```
wtgee: I don't understand the need for this as the `pytest_collection_modifyitems` is altering the tests after collection so it seems like this wouldn't apply anyway. How are you invoking pytest such that this is needed? And is this change somehow related to the `without-hardware` option (it doesn't seem to be) or just bundled in this PR at the same time?

pocs/tests/test_observatory.py
```diff
@@ -36,8 +36,11 @@ def images_dir(tmpdir_factory):
 
 
 def test_error_exit(config):
+    # TODO Describe why this is expected to fail, and how it is different
+    # from the other tests, esp. test_bad_mount_port.
+    # pytest.set_trace()
     with pytest.raises(SystemExit):
-        Observatory(ignore_local_config=True, simulator=['none'])
+        Observatory(ignore_local_config=True, config=config, simulator=['none'])
```
wtgee: Is this `config` needed here?

conftest.py
jamessynge: I discovered the invoke issue while trying to deal with the problem of running pytest on a unit with real hardware.

How to invoke:
```bash
$ cd $POCS
$ pytest --with-hardware=all
```

Yes, this means I bundled the move of conftest.py with the addition of the new option. Let me know if you want me to split.

pocs/tests/test_observatory.py
jamessynge: If config is not required here, then why does this test depend on the config fixture?

Really, I was confused during debugging to find that the config gets loaded from the file(s) even though the test depends on the config fixture.

conftest.py
jamessynge: Let me clarify: the `pytest_addoption` function is the part needed in $POCS/conftest.py, not `pytest_collection_modifyitems`. But, since the latter makes use of the former, it makes sense to keep them together.

jamessynge:
FYI, I'd like to find a way to provide the --without-hardware=all option by default on PAN006, but not on my laptop.

jamessynge:
I added a pytest.ini file to $POCS on PAN006 with --without-hardware=all

conftest.py
wtgee: I invoke `pytest` that way already so not sure why the conftest.py change would make things different. Just trying to understand what it does.

pocs/tests/test_observatory.py
wtgee: It can probably removed from the test. I think in normal situations we would want the config fixture to be used but this is testing that the default config file cannot load properly because there is no associated hardware.

wtgee: Or at least that's the idea. With hardware attached this PR is needed.

wtgee:
> I added a pytest.ini file to $POCS on PAN006 with --without-hardware=all

Yes, I was going to ask how we could do this. Should probably get this documented for units.

jamessynge:
Before merging I'm going to add to .gitignore so that POCS/pytest.ini is ignored.

PR #434, opened 2018-02-03
Upload images for night

wtgee:
* Upload the observation directory during `housekeeping` step
* Move upload function to utils and adjust cli script

Note that I haven't actually tested this yet. I was previously using the `upload_image_dir.py` script without problems.

Edited: I've added an option to the config file about whether or not image uploading should even be tried. This is under a `panoptes_network` heading and currently just has `image_storage`. In the future I see this being expanded, e.g. `network_scheduler`.

### DIFF
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -221,6 +221,7 @@ class Observatory(PanBase):
                 )
 
                 img_utils.clean_observation_dir(dir_name)
+                img_utils.upload_observation_dir(dir_name)
 
             self.logger.debug('Cleanup finished')
 
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -226,6 +226,7 @@ class Observatory(PanBase):
                 except KeyError:
                     self.logger.warning("pan_id not set in config, can't upload images.")
                 else:
+                    self.logger.debug("Uploading directory to google cloud storage")
                     img_utils.upload_observation_dir(pan_id, dir_name)
 
             self.logger.debug('Cleanup finished')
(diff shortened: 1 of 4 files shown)


pocs/utils/images/__init__.py
```diff
@@ -272,3 +273,33 @@ def _glob(s):
...
+        This requires that the command line utility `gsutil` be installed
+        and that authentication has properly been set up.
+
+    Args:
+        dir_name (str): Full path to observation directory.
+    """
+    assert os.path.exists(dir_name)
+    gsutil = shutil.which('gsutil')
+
+    img_path = '{}/*.fz'.format(dir_name)
```
jamessynge: It might be wise to check this using `glob()`.

pocs/utils/images/__init__.py
```diff
@@ -272,3 +273,33 @@ def _glob(s):
...
+    field_dir = dir_name.split('fields')[-1]
+    remote_path = '{}/{}/'.format(pan_id, field_dir).replace('//', '/')
```
jamessynge: I noticed when running the previous script I noticed that there were some double slashes in the linux path. For example:
```python
CommandException: No URLs matched: /var/panoptes/images/fields//HatP30/6575fc/20180127T023054/*.fz
CommandException: 1 file/object could not be transferred.
```
It might be good to use os.join.path rather than string manipulation.

pocs/observatory.py
```diff
@@ -222,6 +233,10 @@ def cleanup_observations(self):
 
                 img_utils.clean_observation_dir(dir_name)
 
+                if upload_images:
```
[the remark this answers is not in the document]
wtgee: Added `is True` (which is "more correct" than `== True`)

PR #438, opened 2018-02-04
Fixes for pan_id

wtgee:


pocs/utils/images/__init__.py
```diff
@@ -300,7 +300,7 @@ def _print(msg):
 
     img_path = os.path.join(dir_name, '*.fz')
```
jamessynge: Maybe use `os.path.normpath()` to tidy the path (i.e. to remove `//` or `/a/../`).

scripts/upload_image_dir.py
```diff
@@ -28,9 +28,9 @@ def main(date, auto_confirm=False, verbose=False):
     db = PanMongo()
     config = load_config()
     try:
-        pan_id = config['PAN_ID']
+        pan_id = config['pan_id']
```
jamessynge: I wondered why you'd chosen the all caps, but added it to my pocs_local.yaml. I figured you had a reason.

scripts/upload_image_dir.py
```diff
@@ -28,9 +28,9 @@ def main(date, auto_confirm=False, verbose=False):
...
     except KeyError:
-        warn("Can't upload without a valid PAN_ID")
+        warn("Can't upload without a valid pan_id in the config")
```
jamessynge: What makes it valid besides being present? Perhaps we should require a specific format? I.e. regex match `PAN\d\d\d`?

jamessynge:
Note, when I search for PAN_ID, I find it in quite a few places. Do they not also need fixing?

scripts/upload_image_dir.py
wtgee: You asked me about it before and I switched it to lowercase at that time. (Originally I had it as an env var and had copied it in as such).

wtgee:
> Note, when I search for PAN_ID, I find it in quite a few places. Do they not also need fixing?

I see one in `storage.py`, which I thought I had fixed. Where else are you seeing them?

jamessynge:
I think you've now found all the cases that I see locally.

Issue #131, opened 2017-11-25
test_messaging.py hangs (sometimes)


jamessynge:
Tests have been passing in general, but on this run test_messaging.py hung overnight.

> $ pytest
> ========================================================== test session starts ===========================================================
> platform linux -- Python 3.6.1, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
> rootdir: /home/james/git/panoptes/POCS, inifile:
> Downloading http://maia.usno.navy.mil/ser7/finals2000A.all
> |====================================================================================================| 3.1M/3.1M (100.00%)         0s
> collected 264 items 
> 
> pocs/tests/test_base_scheduler.py ...............
> pocs/tests/test_camera.py ........s..ssssss..................ssssssssssssssssssssssssss
> pocs/tests/test_codestyle.py F
> pocs/tests/test_config.py .............
> pocs/tests/test_constraints.py ..............
> pocs/tests/test_database.py ...
> pocs/tests/test_dispatch_scheduler.py ........
> pocs/tests/test_dome_simulator.py ...F
> pocs/tests/test_field.py ....
> pocs/tests/test_focuser.py .......sssssss..
> pocs/tests/test_images.py ..........
> pocs/tests/test_ioptron.py .
> pocs/tests/test_messaging.py ...

This is the tail of panoptes.log, log records from messaging.py:

> MainProcess(6322) MainThread 20171125032826 UTC    DEBUG         messaging.py:  80:create_subscriber         Creating subscriber. Port: 54321   Channel: 
> MainProcess(6322) MainThread 20171125032826 UTC    DEBUG         messaging.py:  57:create_publisher          Creating publisher. Binding to port 12345 
> Process-2(7093) MainThread 20171125032826 UTC    DEBUG         messaging.py:  80:create_subscriber         Creating subscriber. Port: 12345     Channel: 
> Process-2(7093) MainThread 20171125032826 UTC    DEBUG         messaging.py:  57:create_publisher          Creating publisher. Binding to port 54321 
> MainProcess(6322) MainThread 20171125032828 UTC    DEBUG         messaging.py:  80:create_subscriber         Creating subscriber. Port: 54321   Channel: 
> MainProcess(6322) MainThread 20171125032828 UTC    DEBUG         messaging.py:  57:create_publisher          Creating publisher. Binding to port 12345 
[...]

PR #439, opened 2018-02-04
Reduce flakiness of test_messaging.py

jamessynge:
Use two multiprocessing.Event objects to coordinate the usage of the
messaging forwarder. In particular, once the forwarder has created
the sockets we can create the publisher and subscriber. And when
shutting down, we should wait until the sockets have been closed
else we may find that we can't create the ports for the next test.
This means we can remove the sleep() calls which slow the tests
down without providing confidence.

Note that this reduces the flakiness on my laptop from perhaps
3% to around 0.2%, which is an improvement, but there is still
room to go.

Partial fix for #131.

### DIFF
--- a/pocs/tests/test_messaging.py
+++ b/pocs/tests/test_messaging.py
@@ -1,21 +1,41 @@
+import multiprocessing
 import pytest
 import time
 
 from datetime import datetime
-from multiprocessing import Process
 from pocs.utils.messaging import PanMessaging
 
 
 @pytest.fixture(scope='function')
 def forwarder():
+    ready = multiprocessing.Event()
+    done = multiprocessing.Event()
+
     def start_forwarder():
-        PanMessaging.create_forwarder(12345, 54321)
+        PanMessaging.create_forwarder(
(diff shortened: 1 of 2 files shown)


pocs/tests/test_messaging.py
```diff
@@ -1,21 +1,41 @@
...
+    # Wait a moment for the forwarder to start using those sockets.
+    time.sleep(0.05)
+
     yield messaging
+
+    # Stop the forwarder. Since we use the same ports in multiple
+    # tests, we wait for the process to shutdown.
     messaging.terminate()
+    for _ in range(100):
+        # We can't be sure that the sub-process will succeed in
```
wtgee: Why can't we be sure of this?

jamessynge:
This PR is being replaced by #440 which works around the Travis CI limitation regarding shared memory.

pocs/tests/test_messaging.py
jamessynge: Because that is what the docs for multiprocessing.Process.terminate say.

PR #440, opened 2018-02-04
Reduce flakiness of test_messaging.py

jamessynge:
Use two Event objects to coordinate the usage of the messaging
forwarder. In particular, once the forwarder has created the sockets
we can create the publisher and subscriber. And when shutting down,
we should wait until the sockets have been closed else we may find
that we can't create the ports for the next test. This means we can
remove the sleep() calls which slow the tests down without providing
confidence.

Unlike PR #439, this uses an instance of multiprocessing.SyncManager,
which means that the Event objects are accessed via proxies
communicating with a server process. This avoids the problem that
[TravisCI doesn't directly support shared memory](https://github.com/travis-ci/travis-cookbooks/issues/155) without funny
workarounds. This avoids the issue entirely.

Note that this reduces the flakiness on my laptop from perhaps
3% to around 0.2%, which is an improvement, but there is still
room to go.

Partial fix for #131.

### DIFF
--- a/pocs/tests/test_messaging.py
+++ b/pocs/tests/test_messaging.py
@@ -6,10 +6,14 @@ from datetime import datetime
 from pocs.utils.messaging import PanMessaging
 
 
+@pytest.fixture(scope='module')
+def mp_manager():
+    return multiprocessing.Manager()
+
 @pytest.fixture(scope='function')
-def forwarder():
-    ready = multiprocessing.Event()
-    done = multiprocessing.Event()
+def forwarder(mp_manager):
+    ready = mp_manager.Event()
+    done = mp_manager.Event()
 
     def start_forwarder():
         PanMessaging.create_forwarder(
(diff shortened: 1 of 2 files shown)


pocs/tests/test_messaging.py
```diff
@@ -1,21 +1,46 @@
+import multiprocessing
 import pytest
 import time
 
 from datetime import datetime
-from multiprocessing import Process
 from pocs.utils.messaging import PanMessaging
 
 
+@pytest.fixture(scope='module')
```
wtgee: Is this what fixes the issue on travis? Just curious.

jamessynge: multiprocessing.Event uses shared memory, while multiprocessing.Manager().Event uses proxies communicating with a shared process.

The scope=module is there just to increase the speed of the tests.

jamessynge:
Sadly, while this worked fine on the first Travis build, it failed on the second. Need to figure out why.

pocs/tests/test_messaging.py
wtgee: Got, thanks.

jamessynge:
Woohoo, latest version works both locally and in Travis.

PR #448, opened 2018-02-06
Make the saved JPG image larger

huntsmanscope:
We should also fix any WCS issues. I have noticed that the values on the axes do not print for images that have a WCS.

### DIFF
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -101,19 +101,11 @@ def _make_pretty_from_fits(fname, **kwargs):
     norm = ImageNormalize(interval=PercentileInterval(percent_value), stretch=LogStretch())
     wcs = WCS(fname)
 
-    nxpix = header.get('NAXIS1')
-    nypix = header.get('NAXIS2')
-
-    dpi = 160
-
-    fnx_size = nxpix / dpi
-    fny_size = nypix / dpi
-
-    plt.figure(figsize=(fnx_size, fny_size), dpi=dpi)
+    plt.figure(figsize=(10, 8), dpi=75)
 
     if wcs.is_celestial:
         ax = plt.subplot(projection=wcs)
-        ax.coords.grid(True, color='white', ls='-', alpha=0.3)
+        ax.coords.grid(True, color='white', ls='-', alpha=0.1)
 
         ra_axis = ax.coords[0]
         ra_axis.set_axislabel('Right Ascension')
@@ -124,7 +116,7 @@ def _make_pretty_from_fits(fname, **kwargs):
         dec_axis.set_major_formatter('dd:mm')
     else:
         ax = plt.subplot()
-        ax.grid(True, color='white', ls='-', alpha=0.3)
+        ax.grid(True, color='white', ls='-', alpha=0.1)
 
         ax.set_xlabel('X / pixels')
         ax.set_ylabel('Y / pixels')
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -97,9 +97,9 @@ def _make_pretty_from_fits(fname, **kwargs):
     title = '{} {}'.format(title, date_time)
     norm = ImageNormalize(interval=PercentileInterval(percent_value), stretch=LogStretch())
     wcs = WCS(fname)
-    
-    plt.figure(figsize=(10,8))
-    
+
+    plt.figure(figsize=(10, 8))
+
     if wcs.is_celestial:
         ax = plt.subplot(projection=wcs)
         ax.coords.grid(True, color='white', ls='-', alpha=0.3)
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -19,6 +19,11 @@ from pocs.utils import error
 from pocs.utils.images import fits as fits_utils
 from pocs.utils.images import focus as focus_utils
 
+palette = copy(plt.cm.inferno)
+palette.set_over('w', 1.0)
+palette.set_under('k', 1.0)
+palette.set_bad('g', 1.0)
+
 
 def crop_data(data, box_width=200, center=None, verbose=False):
     """ Return a cropped portion of the image
@@ -98,11 +103,6 @@ def _make_pretty_from_fits(fname=None, figsize=(10, 8), dpi=150, **kwargs):
     date_time = header.get('DATE-OBS', current_time(pretty=True)).replace('T', ' ', 1)
 
     percent_value = kwargs.get('normalize_clip_percent', 99.9)
-
-    palette = copy(plt.cm.inferno)
-    palette.set_over('w', 1.0)
-    palette.set_under('k', 1.0)
-    palette.set_bad('g', 1.0)
 
     title = '{} (Exposure time: {} s, Filter: {}) {}'.format(title, exp_time, filter, date_time)
     norm = ImageNormalize(interval=PercentileInterval(percent_value), stretch=LogStretch())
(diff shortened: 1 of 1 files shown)


AnthonyHorton:
Need to fix the `codestyle` errors before we merge this.

jamessynge:
Can you provide some more info about what led to this change?

Regarding how to make the size more generic, do you really mean how can you make the size be a specific number of pixels in size? FWIW, figsize is in inches but that doesn't tell you how many pixels wide the figure (saved jpg) will be. You also need the dpi parameter.

jermainegug:
While observing with the huntsman when we got the jpg file for the image they were too low quality to see any structure. 

Just wanted something that isn't hard coded, maybe like a ratio of the pixels in the fits file. 
I plan to add the dpi parameter but wanted to see if I could get a good figure size first.

huntsmanscope:
I was thinking something like this would be nice.
```
nxpix = header.get('NAXIS1')
nypix = header.get('NAXIS2')
    
fnx_size = nxpix / dpi
fny_size = nypix / dpi
    
plt.figure(figsize=(fnx_size, fny_size), dpi=dpi)
```
So at least the size of the image can come from the pixels rather than a hard coded value.

huntsmanscope:
I think there is something wrong with the solved fits images because it detects that they have celestial coordinates but it doesn't write the values on the grid for the axes. I tried it with an example file from online and it worked so must be the way it is writing it out in the header.

huntsmanscope:
m83_solved

wtgee:
I've never gotten the WCS grids labelling to work properly with PANOPTES images. I haven't spent a ton of time with it, but I believe it has to do with the min/max limits and the tick label spacing. By altering those you can get them to appear but it is finicky.

wtgee:
Looking at this image it would almost be nice to have the exposure time (and maybe filter) in the title.

AnthonyHorton:
Looking good, but yes, need to sort out those tick labels for the WCS images.

Setting the output image resolution to equal the FIT image resolution (as above) is not a good approach. You still won't get 1:1 reproduction of the image because of the borders, you wouldn't necessarily want that anyway (the FLI cameras produce 50 Megapixel images, we won't want 50 Megapixel JPEGS!), and the text label can end up way to big or way to small relative to the size of the plot (as in the example above).

Better, set the image size (in inches) to a value that gives sensible sized text, markers, etc., then set dpi to give a suitable resolution for on screen display (1-2 kpixels in each axis).  Make it possible to override the defaults through optional keyword parameters, but for what we want these JPEGS for fixed default values is probably best.

AnthonyHorton:
A couple of other possible enhancements here:

- Call `pocs.utils.focus.mask_saturated()` on the data to make saturation more obvious in the JPEGS (they'll appear black)
- Customise the colormap to make invalid data (e.g. masked saturated pixels) have a conspicuous colour to make it even more obvious. Example code, from `pocs.focuser.focuser.py`:

'''
palette = copy(plt.cm.inferno)
palette.set_over('w', 1.0)
palette.set_under('k', 1.0)
palette.set_bad('g', 1.0)
'''

Then plotting with `cmap=palette` will result in masked pixels being green.

huntsmanscope:
I can't find out how to fix the tick labels for the WCS images. I have just changed it back so that we are using inches.

I don't know how to do an override with a keyword parameter and I am not sure how to implement the two enhancements you have mentioned.

pocs/utils/images/__init__.py
```diff
@@ -101,19 +101,11 @@ def _make_pretty_from_fits(fname, **kwargs):
...
-    nxpix = header.get('NAXIS1')
-    nypix = header.get('NAXIS2')
-
-    dpi = 160
-
-    fnx_size = nxpix / dpi
-    fny_size = nypix / dpi
-
-    plt.figure(figsize=(fnx_size, fny_size), dpi=dpi)
+    plt.figure(figsize=(10, 8), dpi=75)
```
AnthonyHorton: `dpi` a bit low? This will give a 750 x 600 pixel image.  I'd be tempted to 100-150 dpi.

AnthonyHorton:
By overriding with keyword parameters I simply meant adding `size` and `dpi` arguments to the function, and give them default values (e.g. (10, 8) and 150).  Then the user has the option of using those default values, or changing them by passing something like `size=(12,8)` in the function call.

AnthonyHorton:
As for my suggested enhancements, my comment includes pretty much everything you need to do.  Look in `pocs/utils/focus.py` & you'll find the `mask_saturated()` function.  In this file you just need to `import pocs.utils.focus as focus_utils`, then do `data = mask_saturated(data)` to turn the array `data` into a masked array with saturated pixels masked.  `matplotlib` understands masked arrays and marks the masked pixels as 'bad', which by default will be plotted with the same colour as the low value limit of your colourmap (black for inferno).

If you do the second thing I suggested you can change the colour used for 'bad' pixels to something that distinguishes them for pixels that are OK but just have low values, e.g. green which isn't part of the normal inferno colourmap at all.

Just need `from copy import copy` then those 4 lines I gave you up at the top of the file, then in your `plt.imshow()` you use `cmap=palette` instead of `cmap='inferno'` in order to use your customised colourmap.

pocs/utils/images/__init__.py
```diff
@@ -101,7 +101,10 @@ def _make_pretty_from_fits(fname, **kwargs):
     norm = ImageNormalize(interval=PercentileInterval(percent_value), stretch=LogStretch())
     wcs = WCS(fname)
 
-    plt.figure(figsize=(10, 8), dpi=150)
+    figsize = (10, 8)
+    dpi = 150
```
AnthonyHorton: This edit changes nothing if you're still setting the values right here in the code. Put them in the function definition, e.g. replace `def _make_pretty_from_fits(fname, **kwargs):` with `def _make_pretty_from_fits(fname, figsize=(10, 8), dpi=150, **kwargs):` and take these two lines out.

huntsmanscope: That's what I was confused, but now I know what you mean.

jermainegug:
It can't find `pocs.utils.focus`.

AnthonyHorton:
Scroll up to my previous comment. Should be `pocs.utils.images.focus`.

wtgee:
@jermainegug If you are using Atom you can get an autopep8 plugin that will fix basic errors every time you save. Tests are failing again because of one missing space. Would be nice to not have to go through the commit/test/review cycle for little things like that and you don't want to spend your time thinking about it ideally.

autopep8: https://atom.io/packages/python-autopep8

Better yet, get something like https://atom.io/packages/ide-python which will do code checking, command completion, autope8, etc.

jermainegug:
@wtgee Yeah I know... but I am using the huntsman computer since mine decided to not turn on the first day we got here (think the hard drive has decided to die but that's okay cause I have everything backed up!), so I don't want to change branches on the huntsman computer just in case we lose something.

huntsmanscope:
Ok, all of @AnthonyHorton enhancements have been added and it the code has passed the checks. Just need to find out how we can get the WCS on our axis, @wtgee I have found nothing online for similar problems.

wtgee:
>  @wtgee I have found nothing online for similar problems.

I never did either which is why I never solved the problem. Might try looking in the actual github issues for astropy. They have a lot of issues so have to do some valid search terms.

wtgee:
>  @wtgee I have found nothing online for similar problems.

I never did either which is why I never solved the problem. Might try looking in the actual github issues for astropy. They have a lot of issues so have to do some valid search terms.

huntsmanscope:
Ok, maybe create another issue and merge these changes so that we are getting decent jpg files for our observing tonight?

AnthonyHorton:
I'm OK with merging this before we fix the WCS tick mark labelling so we can use it tonight, but should move the colormap customisation code (`palette` stuff) out of the function and into the preamble (just below the imports), as in my comment above (now hidden as 'outdated').

You're setting up the same colourmap, from scratch, every time the function is called.

PR #449, opened 2018-02-06
Fixing async calls for TheSkyX

wtgee:
Note: This still needs some work to make all the calls consistent but
can be used for testing now.

### DIFF
--- a/pocs/mount/bisque.py
+++ b/pocs/mount/bisque.py
@@ -30,9 +30,9 @@ class Mount(AbstractMount):
         self.template_dir = template_dir
 
 
-##################################################################################################
+##########################################################################
 # Methods
-##################################################################################################
+##########################################################################
 
     def connect(self):
         """ Connects to the mount via the serial port (`self._port`)
@@ -92,6 +92,7 @@ class Mount(AbstractMount):
     def _update_status(self):
         """ """
         status = self.query('get_status')
+        self.logger.debug("Status: {}".format(status))
 
         try:
             self._at_mount_park = status['parked']
@@ -152,9 +153,9 @@ class Mount(AbstractMount):
         self.logger.info("Mount park position set: {}".format(self._park_coordinates))
 
 
-##################################################################################################
+##########################################################################
 # Movement methods
-##################################################################################################
+##########################################################################
 
     def slew_to_target(self):
         """ Slews to the current _target_coordinates
@@ -183,6 +184,11 @@ class Mount(AbstractMount):
                     'dec': mount_coords[1],
                 })
                 success = response['success']
+                if success:
+                    self.status()
+                    while self.is_slewing:
+                        self.status()
+                        time.sleep(2)
 
             except Exception as e:
                 self.logger.warning(
@@ -192,6 +198,8 @@ class Mount(AbstractMount):
                 if not self.query('start_tracking')['success']:
                     self.logger.warning("Tracking not turned on for target")
                     self._is_tracking = True
+                else:
+                    self.logger.debug("Now tracking at target")
 
         return success
 
@@ -230,10 +238,7 @@ class Mount(AbstractMount):
         self.logger.debug('Parking mount')
         response = self.query('park')
 
-        if response['success']:
-            while not self.is_parked:
-                self.status()
-                time.sleep(2)
+        self._is_parked = response['success']
 
         return self.is_parked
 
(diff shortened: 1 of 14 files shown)


wtgee:
This was tested on the Huntsman system and worked well. TheSkyX simulator doesn't support all options so makes it hard to test otherwise.

pocs/mount/bisque.py
```diff
@@ -183,6 +184,11 @@ def slew_to_target(self):
                     'dec': mount_coords[1],
                 })
                 success = response['success']
+                if success:
+                    self.status()
+                    while self.is_slewing:
```
jamessynge: I'm confused by this code. self.status() returns a dictionary, which is ignored. Why bother making the status calls?

pocs/mount/bisque.py
```diff
@@ -192,6 +198,8 @@ def slew_to_target(self):
                 if not self.query('start_tracking')['success']:
                     self.logger.warning("Tracking not turned on for target")
                     self._is_tracking = True
+                else:
+                    self.logger.debug("Now tracking at target")
```
jamessynge: Perhaps add the target coords to the log message?

resources/bisque/guider/connect.js
```diff
@@ -0,0 +1,7 @@
+/* Java Script */
```
jamessynge: Given that these files have the `.js` extension, why to they all start with `/* Java Script */`?

resources/bisque/guider/regulate_temperature.js
```diff
@@ -0,0 +1,13 @@
+/* Java Script */
+var msg, success;
+ccdsoftCamera.Autoguider = 1;
+ccdsoftCamera.Connect();
```
jamessynge: I'm guessing that this Connect call can fail, but that isn't checked.

resources/bisque/guider/connect.js
```diff
@@ -0,0 +1,7 @@
...
+ccdsoftCamera.Autoguider = 1;
+ccdsoftCamera.Connect();
```
jamessynge: Why not check the result of the connect call?

resources/bisque/guider/disconnect.js
```diff
@@ -0,0 +1,7 @@
+/* Java Script */
+ccdsoftCamera.Autoguider = 1;
```
jamessynge: Why set this parameter here?

resources/bisque/guider/get_state.js
```diff
@@ -0,0 +1,46 @@
+/* Java Script */
+var msg, success;
+if (ccdsoftCamera.Connect()) {
+    msg = "DFError: Not connected";
```
jamessynge: What does "DFError" mean?

resources/bisque/guider/regulate_temperature.js
```diff
@@ -0,0 +1,13 @@
+/* Java Script */
+var msg, success;
+ccdsoftCamera.Autoguider = 1;
```
jamessynge: Again, why is `ccdsoftCamera.Autoguider = 1;` here?

resources/bisque/guider/reset.js
```diff
@@ -0,0 +1,9 @@
+/* Java Script */
+ccdsoftCamera.Autoguider = 1;
+ccdsoftCamera.Asynchronous = false;
+ccdsoftCamera.Disconnect();
```
jamessynge: Check status of these calls?

pocs/mount/bisque.py
wtgee: `status` calls `_update_status`, which sets a number of attributes based on that status (`is_parked`, `is_slewing`, etc).

Which actually goes back to the question of whether or not we use `status` for any decision logic.

resources/bisque/guider/connect.js
wtgee: Because TheSkyX has implemented this in a terrible, terrible fashion (besides the fact that he constantly refers to them as Java scripts).

This is the "special key" that lets TheSkyX know it is valid. 

http://www.bisque.com/scriptTheSkyX/scriptOverSocket.html

wtgee:
Just a note that most of the javascript files come from the original Dragonfly project that Huntsman was based on and I haven't changed them a lot, except where as necessary. As part of dealing with this async issue I was looking more closely at them and plan to do a major overhaul of many of them. They could definitely be vastly improved.

PR #451, opened 2018-02-06
Use master, dilated saturated pixel mask in autofocus

AnthonyHorton:
This is a fix for a problem seen with Huntsman, autofocus failing to select the best focus position when a saturated star is present in the focus images.  The modified autofocus algorithm masks a region surrounding any pixels which saturate in any of the images in the focus sweep.

### DIFF
--- a/pocs/focuser/focuser.py
+++ b/pocs/focuser/focuser.py
@@ -381,7 +381,7 @@ class AbstractFocuser(PanBase):
         n_positions = len(focus_positions)
         thumbnails = np.zeros((n_positions, thumbnail_size, thumbnail_size), dtype=thumbnail.dtype)
         masks = np.empty((n_positions, thumbnail_size, thumbnail_size), dtype=np.bool)
-        metric = np.empty(n_positions)
+        metric = np.empty((n_positions))
 
         for i, position in enumerate(focus_positions):
             # Move focus, updating focus_positions with actual encoder position after move.
--- a/pocs/focuser/focuser.py
+++ b/pocs/focuser/focuser.py
@@ -2,6 +2,7 @@ import matplotlib.colors as colours
 import matplotlib.pyplot as plt
 
 from scipy.interpolate import UnivariateSpline
+from scipy.ndimage import binary_dilation
 
 import numpy as np
 
@@ -324,7 +325,10 @@ class AbstractFocuser(PanBase):
                    plots,
                    start_event,
                    finished_event,
-                   smooth=0.4, *args, **kwargs):
+                   smooth=0.4,
+                   dilations=10,
+                   *args,
+                   **kwargs):
         # If passed a start_event wait until Event is set before proceeding
         # (e.g. wait for coarse focus to finish before starting fine focus).
         if start_event:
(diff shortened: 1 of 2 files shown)


pocs/focuser/focuser.py
```diff
@@ -387,13 +392,17 @@ def _autofocus(self,
...
             if dark_thumb is not None:
                 thumbnail = thumbnail - dark_thumb
-            # Calculate focus metric
-            metric[i] = focus_utils.focus_metric(
-                thumbnail, merit_function, **merit_function_kwargs)
-            self.logger.debug("Focus metric at position {}: {}".format(position, metric[i]))
+            thumbnails[i] = thumbnail
+
+        master_mask = masks.any(axis=0)
+        master_mask = binary_dilation(master_mask, iterations=dilations)
```
wtgee: Any chance this will ever mask out too much?

wtgee: Any chance this will ever mask out too much?

AnthonyHorton: It might mask slightly more than absolutely necessary under some circumstances. The size of the masked regions around saturated pixels is controlled by the number of dilations, which I've made a (secret, undocumented, inaccessible) adjustable parameter.

For any sane value of `dilations` it's never going to mask so much that it interferes with the focusing though.  The only way it would mask a significant proportion of the image is if there was a *lot* of saturated stars (i.e. exposure time is too long), which would cause the focusing to fail regardless.

I should probably made the number of dilations (and the smoothing parameter for the spline fit) into documentated, accessible parameters that can be adjusted via the config file, though, like the rest of the autofocus algorithm parameters.

wtgee: Great, thanks for explanation. If you can't make the dilations accessible immediately but want to test this tonight that is okay.

AnthonyHorton:
OK, should now be able to tinker with `autofocus_mask_dilations` and `autofocus_spline_smoothing` via the config file, the same as all the other autofocus parameters. I also put them in the docstrings.

Happy to get this merged now so we can give it some thorough testing tonight.  Still need to implement a timeout here so that autofocus will end even if a camera stops responding, but I'll do that in a different branch as that's an unrelated issue.

pocs/focuser/focuser.py
```diff
@@ -149,6 +149,8 @@ def autofocus(self,
                   take_dark=None,
                   merit_function=None,
                   merit_function_kwargs=None,
+                  mask_dilations=None,
+                  spline_smoothing=None,
```
wtgee: You mention the default value in the docstring but then it is hard-coded below. any reason not to put here? Same for `mask_dilations`

pocs/focuser/focuser.py
```diff
@@ -324,7 +346,10 @@ def _autofocus(self,
                    plots,
                    start_event,
                    finished_event,
-                   smooth=0.4, *args, **kwargs):
+                   smooth,
```
wtgee: The order is is backwards from everywhere else (with smooth first then dilations). I notice also that this is just a positional parameter then the thread passes kwargs. I'm assuming things go in correctly but seems a little odd just since they are not keyword args here. Probably doesn't need to change.

pocs/focuser/focuser.py
AnthonyHorton: Same with the other autofocus parameters, for the most part.

It's a somewhat ugly hack that resulted from me attempting to provide lots of different ways to set/override these values: in the config file (which feeds into the `Focuser ` constructor, which stores them as member variables), via the corresponding `Camera` object's `autofocus()` methods (which is what the `Observatory` calls), and via the `Focuser` object's `autofocus()` method.  The hard coded defaults only get used if there are no values set from either the config file or the call to `autofocus()`.

pocs/focuser/focuser.py
AnthonyHorton: Re-ordered for consistency, but yes, it shouldn't matter as I always call it with keyword arguments (for precisely this reason!).

Issue #178, opened 2017-12-08
Don't use fixed ports for messaging


wtgee:
We use zeromq to create a pub/sub forwarder between our various components and this is configured with a fixed port from the configuration file. Occasionally this is creating issues, especially when killing and then restarting tests, as the cleanup doesn't properly run and the code doesn't handle a port conflict well.

See #131

PR #453, opened 2018-02-07
Add run_messaging_hub, a forwarder for all port pairs

jamessynge:
We use pairs of ports to forward messages from publishers to
subscribers. scripts/run_messaging_hub.py will forward messages between
multiple pairs of ports, including those defined in
conf_files/pocs.yaml.

This file can replace scripts/start_messenger.py which only forwards
messages between a single fixed pair of ports.

### DIFF
--- a/scripts/run_messaging_hub.py
+++ b/scripts/run_messaging_hub.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+
+import argparse
+# import multiprocessing
+import sys
+import threading
+import time
+
+from pocs.utils.config import load_config
+from pocs.utils.logger import get_root_logger
+from pocs.utils.messaging import PanMessaging
+
+
+def run_forwarder(logger, sub_port, pub_port):
+    msg = 'Starting {} -> {} forwarder'.format(sub_port, pub_port)
+    print(msg)
+    logger.info(msg)
+    try:
+        PanMessaging.create_forwarder(sub_port, pub_port)
+    except Exception:
+        pass
+    msg = 'Forwarder for {} -> {} has stopped'.format(sub_port, pub_port)
+    print(msg)
+    logger.info(msg)
+
+
+def run_forwarders(sub_and_pub_pairs):
+    logger = get_root_logger()
+    logger.info('Starting forwarders')
+    threads = []
+    for sub, pub in sub_and_pub_pairs:
+        name = 'fwd_{}_to_{}'.format(sub, pub)
+        t = threading.Thread(target=run_forwarder, name=name, args=(logger, sub, pub), daemon=True)
+        logger.info('Starting thread {}', name)
+        t.start()
+        threads.append(t)
+        time.sleep(0.05)
+    time.sleep(0.2)
+    if not any([t.is_alive() for t in threads]):
+        msg = 'Failed to start any forwarder!'
+        logger.error(msg)
+        print(msg)
+    else:
+        print()
+        print('Hit Ctrl-c to stop')
+    try:
+        # Keep running until they've all died.
+        while threads:
+            for t in threads:
+                t.join(timeout=100)
+                if t.is_alive():
+                    continue
+                logger.info('Thread {} has stopped', t.name)
+                threads.remove(t)
+                break
+        # If we get here, then the forwarders died for some reason.
+        sys.exit(1)
+    except KeyboardInterrupt:
+        sys.exit(0)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Record sensor data from an Arduino and send it relay commands.')
+    parser.add_argument(
+        '--pair',
+        dest='pairs',
+        nargs=2,
+        action='append',
+        type=int,
+        help="Pair of ports to be forwarded: subscriber (input) port and publisher (output) port.")
+    parser.add_argument(
+        '--port',
+        dest='ports',
+        action='append',
+        type=int,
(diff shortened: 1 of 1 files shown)


jamessynge:
This is related to #97 and #178.

scripts/run_messaging_hub.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+
+import argparse
+# import multiprocessing
```
wtgee: Just a note to remove if not needed.

scripts/run_messaging_hub.py
```diff
@@ -0,0 +1,130 @@
...
+
+def run_forwarder(logger, sub_port, pub_port):
+    msg = 'Starting {} -> {} forwarder'.format(sub_port, pub_port)
+    print(msg)
+    logger.info(msg)
+    try:
+        PanMessaging.create_forwarder(sub_port, pub_port)
+    except Exception:
+        pass
+    msg = 'Forwarder for {} -> {} has stopped'.format(sub_port, pub_port)
```
wtgee: Maybe we can catch a `KeyboardInterrupt` and show an `info` message but on other `Exceptions` show a warning?

scripts/run_messaging_hub.py
```diff
@@ -0,0 +1,130 @@
...
+                break
+        # If we get here, then the forwarders died for some reason.
+        sys.exit(1)
+    except KeyboardInterrupt:
+        sys.exit(0)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Record sensor data from an Arduino and send it relay commands.')
```
wtgee: Change description here.

scripts/run_messaging_hub.py
```diff
@@ -0,0 +1,130 @@
...
+        help="Pair of ports to be forwarded: subscriber (input) port and publisher (output) port.")
+    parser.add_argument(
+        '--port',
+        dest='ports',
+        action='append',
+        type=int,
+        help='First port of a pair to be forwarded. The other is the next integer.')
+    parser.add_argument(
+        '--from_config',
+        action='store_true',
```
wtgee: Do you want `default=True` on this?

scripts/run_messaging_hub.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def arg_error(msg):
+        print(msg, file=sys.stderr)
+        parser.print_help()
+        sys.exit(1)
+
+    all_ports = []
+
+    def validate_port(port):
+        if not (0 < port and port < 65536):
```
wtgee: Should probably skip all the protected ports below 1024

scripts/run_messaging_hub.py
```diff
@@ -0,0 +1,130 @@
...
+            arg_error(
+                'Port {} is unsupported; must be between 0 and 65536, exclusive.'.format(port))
+        if port in all_ports:
+            arg_error('Port {} specified more than once.'.format(port))
+        all_ports.append(port)
```
wtgee: You want to add to `all_ports` here during validation? You also add to it down below in `add_pair` right after calling `validate_port`

wtgee: Seems like this should maybe return a bool (or the `port` if valid and `None` if not)

scripts/run_messaging_hub.py
wtgee: I see you catch Ctrl-C below so above comment might not be valid.

jamessynge: Right, the interrupt is handled on the main thread only IIUC.

scripts/run_messaging_hub.py
jamessynge: Is something very wrong with that? I don't see it.

scripts/run_messaging_hub.py
jamessynge: Thanks.

scripts/run_messaging_hub.py
jamessynge: No, because I'm allowing the caller to decide which... except I've pretty much settled on running it as:
```bash
$ run_messaging_hub.py --from_config
```
The other args may end up rarely used. If that is the case, perhaps I will change to --skip_config instead of --from_config.

scripts/run_messaging_hub.py
jamessynge: Glad you agree... that is what I originally had.

scripts/run_messaging_hub.py
jamessynge: Forgot to remove the append later. I just wanted a very simple mechanism that would not have a lot of branches hiding the basics. And, yes, I agree that the name `validate_port` doesn't imply that it is adding to `all_ports`. I'll change the name to `validate_unique_port` to clarify that it works with `all_ports`.

scripts/run_messaging_hub.py
wtgee: I don't think the messenger hub (this script) is necessarily related to just the arduino and sensor data. It handles the POCS `say` messages and shutdown commands, etc. It's more generic than just the Arduinos. Or is that not true?

In my head the description for this script would be something like "Sets up a publish and subscribe messaging hub for various communication between various components of the system" (don't feel tied to that text at all).

jamessynge: Sorry, I'm an idiot. I got confused between this script and another that is related to the sensors, from which I started when creating this new file. I'll send a fix.

Issue #361, opened 2018-01-16
Handle meridian flipping better


wtgee:
On PAN001 last night it tracked Wasp-35 for 55/60 images before it stopped. It was about 30° past the meridian, which is how far it is set to go. However, the scheduler should not have selected the target unless it can get through the full 60 images. Furthermore, the image timed out while trying to do a tracking correction whereas it should have seen that the mount status was no longer `Tracking` and moved to `scheduling` instead.

The tracking correction failed, which sends the mount to park. If the mount fails in the night when conditions are still good it will try to start up again. Feasibly it could do this without going all the way to park position first although that is probably safest.

So:
- [ ] Figure out what scheduler didn't know target would hit limits past meridian.
- [ ] Figure out what mount status says when stopped because of meridian.
- [ ] Decide if error in night with safe conditions should go home and try again or park and try again. (Or not try again).

wtgee:
I've set PAN001 to record the mount status tonight so can see what it reports when stopped at meridian. I suspect it just reads 'Stopped', which we could easily handle.

PR #454, opened 2018-02-08
Clearing Observations

wtgee:
* Don't clear the fields file and list when clearing the loaded observations
* Clean up tests and make more consistent

### DIFF
--- a/pocs/scheduler/scheduler.py
+++ b/pocs/scheduler/scheduler.py
@@ -175,8 +175,6 @@ class BaseScheduler(PanBase):
         """Reset the list of available observations"""
         # Clear out existing list and observations
         self.current_observation = None
-        self._fields_file = None
-        self._fields_list = None
         self._observations = dict()
 
     def get_observation(self, time=None, show_all=False):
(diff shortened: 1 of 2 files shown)


pocs/scheduler/scheduler.py
```diff
@@ -175,8 +175,6 @@ def clear_available_observations(self):
         """Reset the list of available observations"""
         # Clear out existing list and observations
         self.current_observation = None
-        self._fields_file = None
```
jamessynge: Oh, that looks like a problem. Why have you not seen the problem I reported? Are you restarting PAN001 most days?

wtgee: The `clear_available_observations` is triggered when no observations are found, which will happen if it is looking for a new target close to sunrise. PAN001 hasn't been operating most days since I started it because of bad weather (I've only had two nights in last week) and the other times it was in the middle of an observation all the way until Unsafe (because of daytime), so it never got to the NoObservations from scheduling. (This last part is vaguely related to the first checkbox in #361, where it is scheduling a sequence even though the tail end gets interrupted by either meridian or sunrise.)

PR #455, opened 2018-02-08
Don't clear current db

### LINKED ISSUE #443: get_current db issue
Since #414 was merged there is a frequent problem that when the weather record is being updated it is no longer available in the collection, causing weather safety to fail and parking the unit. It seems like the `insert_current` is taking a non-negligible amount of time to update. 

However, on PAN001 I am getting a failure after during the weather check after every move past the pointing state although the checks before that work fine, so it doesn't seem to necessarily be a timing issue. 

Will do more debugging.

wtgee:
This was causing some problems when new connections to the db were
made. This reverts us to previous behavior, which was in use for a long
time.

Closes #443


wtgee:
Right after this PR I realized that my autopep8 has reverted to 79 chars for some reason. I spent about 15 minutes trying to fix it and couldn't. Not sure what is going on. Sorry about that. I'm now realizing this is the source of some of the formatting complaints for last few weeks.

jamessynge:
That explains it!

PR #458, opened 2018-02-08
Handle SBIG readout errors more gracefully

AnthonyHorton:
During Huntsman commissioning we had issues with occasional `CE_RX_TIMEOUT` errors from the SBIG camera driver during camera readouts.  This resulting in exposures that never completed properly, causing issues like the autofocus function never completing.

These small changes are an attempt to handle errors during image readout more gracefully, ensuring that the `CC_END_READOUT` command gets sent and that the `exposure_event` gets set.


pocs/camera/fli.py
```diff
@@ -291,9 +291,8 @@ def _readout(self, filename, width, height, header, exposure_event):
                 warn(message)
                 break
 
-        self._exposure_lock.release()
-
         fits_utils.write_fits(image_data, header, filename, self.logger, exposure_event)
+        self._exposure_lock.release()
```
wtgee: So by putting this after, will this just write out whatever data you have up to this point?

pocs/camera/sbigudrv.py
```diff
@@ -347,33 +349,50 @@ def _readout(self, handle, centiseconds, filename, readout_mode_code,
...
+            self._set_handle(handle)
+            self._send_command('CC_END_EXPOSURE', params=end_exposure_params)
+            self._send_command('CC_START_READOUT', params=start_readout_params)
+            for i in range(height):
+                try:
+                    self._send_command('CC_READOUT_LINE',
+                                       params=readout_line_params,
+                                       results=as_ctypes(image_data[i]))
+                except RuntimeError as err:
+                    message = 'Readout error on {}: expected {} rows, got{}!'.format(handle,
```
wtgee: Does `handle` show the camera id or name? I can't remember but it seems like it tell us which camera it was os that we had to track it down.

AnthonyHorton: Neither, unfortunately. It's the internal identifier that the driver defines and uses to identify which camera its talking to.  I could (and probably should) modify these to give the camera UID instead, which is information that the driver can relatively easily retrieve.

pocs/camera/fli.py
AnthonyHorton: Yes, I want the call the `write_fits` to happen even if the readout encounters errors occur because it's the writing of the FITS file to disc (or at least the attempt to) that results in the `exposure_event` being set.  Can't set the event before writing the FITS file because that would trigger `process_observation()` to try to process a file that doesn't exist yet.

PR #461, opened 2018-02-10
Determine if PanMongoDB works just once

jamessynge:
Refactor db fixture to use db_type fixture.

Modify db_type fixture to determine if mongo is available just once.

Add log messages at the start and end of each test. This makes it easier
to debug tests because you can tell which log messages go with which test
(mostly).

Remove temp_file from pocs/tests/conftest.py as it is already in
the root conftest.py.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -8,6 +8,7 @@ import os
 import pytest
 
 from pocs import hardware
+from pocs.utils.logger import get_root_logger
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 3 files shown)


jamessynge:
"mostly" because I think that some fixtures might run earlier or later than these.

And I expect travis to fail because I found that I left an extra comma in the code.

jamessynge:
I'll remove your changes to the timeouts, too.

wtgee:
Great, thanks. Looks good

PR #462, opened 2018-02-10
Add install of Pandas

jamessynge:


jamessynge:
Please provide more details about the install-dependencies.sh failure. I just ran it locally, where it ran to completion.

PR #463, opened 2018-02-10
Add PanMemoryDB, an in memory variant for testing

jamessynge:
Rename test_panmongo.py to test_database.py.
Add clearing of the testing database to the db_type fixture.
Add AbstractPanDB, base class for the concrete DB types,
with all the method documentation.
Modify PanDB() into a constructor for the actual DB type.


jamessynge:
I'm going to send some smaller PRs for portions of this which are less 'controversial' (require less justification).

pocs/utils/database.py
```diff
@@ -111,295 +229,278 @@ def __init__(self,
...
     def _make_id(self):
         return str(uuid4())
 
-    def _warn(self, *args, **kwargs):
-        if hasattr(self, 'logger'):
-            self.logger.warning(*args, **kwargs)
-        else:
-            warn(*args)
+    @classmethod
+    def reset_for_test(cls, db_name):
```
wtgee: Just marking here for comments about making more generic.

pocs/utils/database.py
```diff
@@ -111,295 +229,278 @@ def __init__(self,
...
+
+    We serialize the objects in order to test the same code path used
+    when storing in an external database.
+    """
+
+    active_dbs = weakref.WeakValueDictionary()
+
+    @classmethod
+    def get_or_create(cls, db_name=None, **kwargs):
+        """Returns the named db, creating if needed.the
```
wtgee: Fix up end of sentence.

pocs/utils/database.py
```diff
@@ -111,295 +229,278 @@ def __init__(self,
...
+
+        This method exists because PanDB gets called multiple times for
+        the same database name, With mongo or a file store where the storage
```
wtgee: s/W/w/

pocs/utils/database.py
```diff
@@ -111,295 +229,278 @@ def __init__(self,
...
+        self.lock = threading.Lock()
+
+        logger = self.logger or get_root_logger()
+        logger.info('Create new PanMemoryDB for {}', self.db_name)
+
+    def _make_id(self):
+        return str(uuid4())
+
+    def insert_current(self, collection, obj, store_permanently=True):
+        self._warn('insert_current {!r} {!r}'.format(collection, obj))
```
wtgee: Do you want to keep this in here in the long run or was this mostly for testing?

pocs/utils/database.py
```diff
@@ -111,295 +229,278 @@ def __init__(self,
...
+            self._warn("Problem inserting object into current collection: {}, {!r}".format(e, obj))
+            return None
+        with self.lock:
+            self.current[collection] = obj
+            if store_permanently:
+                self.collections.setdefault(collection, {})[obj_id] = obj
+        return obj_id
+
+    def insert(self, collection, obj):
+        self._warn('insert {!r} {!r}'.format(collection, obj))
```
wtgee: Likewise for this and a few below. Not sure if you want them in there long-term or not.

pocs/utils/database.py
```diff
@@ -111,295 +229,278 @@ def __init__(self,
...
+
+    def clear_current(self, type):
+        try:
+            del self.current['type']
+        except KeyError:
+            pass
+
+    @classmethod
+    def reset_for_test(self, db_name):
+        # For some reason we're not seeing all the references disappear
```
wtgee: This comment is true just for the PanMemoryDB?

scripts/plot_weather.py
```diff
@@ -147,7 +147,7 @@ def get_table_data(self, data_file):
 
             print('  Retrieving data from Mongo database')
             db = PanDB()
-            entries = [x for x in db.db.weather.find(
+            entries = [x for x in db.weather.find(
```
wtgee: Thanks, wasn't happy with the ugly specific hack.

pocs/base.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, *args, **kwargs):
             db_type = self.config['db']['type']
             db_name = self.config['db']['name']
 
-            _db = PanDB(db_type=db_type, db_name=db_name, logger=self.logger).db
+            _db = PanDB(db_type=db_type, db_name=db_name, logger=self.logger)
```
wtgee: Much nicer, thanks.

pocs/tests/conftest.py
```diff
@@ -38,9 +40,6 @@ def config_with_simulated_dome(config):
     return config
 
 
-_can_connect_to_mongo = None
-
-
 def can_connect_to_mongo():
```
wtgee: Any chance this should be in the root conftest file?

pocs/tests/conftest.py
```diff
@@ -59,11 +58,12 @@ def can_connect_to_mongo():
     return _can_connect_to_mongo
 
 
-@pytest.fixture(scope='function', params=['mongo', 'file'])
+@pytest.fixture(scope='function', params=['mongo', 'file', 'memory'])
 def db_type(request):
     # If testing mongo, make sure we can connect, otherwise skip.
     if request.param == 'mongo' and not can_connect_to_mongo():
         pytest.skip("Can't connect to {} DB, skipping".format(request.param))
+    PanDB.reset_for_test(request.param, 'panoptes_testing')
```
wtgee: See comments below about `reset_for_test`

pocs/tests/test_database.py
```diff
@@ -31,14 +34,33 @@ def test_simple_insert(db):
...
+    with pytest.raises(InvalidCollection):
         db.insert_current('foobar', {'test': 'insert'})
 
-    with pytest.raises(AssertionError):
+    with pytest.raises(InvalidCollection):
         db.insert('foobar', {'test': 'insert'})
 
 
-def test_bad_object(db):
+def test_log_bad_object(db, caplog):
```
wtgee: I know I saw `caplog` fixture somewhere but did it get pulled out of this PR or am I just missing it?

pocs/tests/test_pocs.py
```diff
@@ -192,13 +191,32 @@ def test_is_weather_safe_no_simulator(pocs):
     assert pocs.is_weather_safe() is False
 
 
+def wait_for_message(sub, type=None, attr=None, value=None):
```
wtgee: I think I'm guilty of this, but `type` is a built-in function and probably shouldn't be used as a variable name. (See also comments about `id` below).

pocs/tests/test_pocs.py
```diff
@@ -192,13 +191,32 @@ def test_is_weather_safe_no_simulator(pocs):
...
+
 def test_run_wait_until_safe(observatory):
     os.environ['POCSTIME'] = '2016-09-09 08:00:00'
 
     # Make sure DB is clear for current weather
     observatory.db.clear_current('weather')
 
     def start_pocs():
+        observatory.logger.info('start_pocs ENTER')
+        # Why is this not ALL hardware (i.e. why change from the default
```
wtgee: It's removing `weather` from the list so we can test for bad weather before setting to good, otherwise the weather simulator would always report good. (which is what the whole test is for)

pocs/tests/test_pocs.py
```diff
@@ -220,32 +238,26 @@ def start_pocs():
...
         assert pocs.is_weather_safe() is True
         pocs.power_down()
+        observatory.logger.info('start_pocs EXIT')
 
     pub = PanMessaging.create_publisher(6500)
     sub = PanMessaging.create_subscriber(6511)
 
-    pocs_process = Process(target=start_pocs)
-    pocs_process.start()
+    pocs_thread = threading.Thread(target=start_pocs)
```
wtgee: Do you want to strip this out in lieu of #468?

pocs/tests/test_pocs.py
```diff
@@ -220,32 +238,26 @@ def start_pocs():
...
+    time.sleep(2)
+    # Insert a dummy weather record to break wait
+    observatory.db.insert_current('weather', {'safe': True})
 
-        if msg_type == 'STATUS':
-            current_state = msg_obj.get('state', {})
-            if current_state == 'pointing':
-                pub.send_message('POCS-CMD', 'shutdown')
-                break
+    wait_for_message(sub, type='STATUS', attr='state', value='pointing')
```
wtgee: Note that we could probably change this from `pointing` to `scheduling` or `ready` to speed up the test slightly. Once it is out of `sleeping` the test is a success and the scheduling can potentially take non-trivial amounts of time.

pocs/utils/database.py
```diff
@@ -8,6 +10,108 @@
...
+
+    def _warn(self, *args, **kwargs):
+        if self.logger:
+            self.logger.warning(*args, **kwargs)
+        else:
+            warn(*args)
+
+    def validate_collection(self, collection):
+        if collection not in self.collection_names:
+            msg = 'Collection type {!r} not available'.format(collection)
```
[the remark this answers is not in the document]
jamessynge: Necessary? No. I prefer it because it puts quotes around it if it is a string; for example, an empty string would appear as "", which is a lot easier to read in the whole message, which then would be `Collection type "" not available`, as opposed to `Collection type  not available`.

pocs/utils/database.py
```diff
@@ -8,6 +10,108 @@
...
+            self._warn(msg)
+            # Can't import pocs.utils.error earlier
+            from pocs.utils.error import InvalidCollection
```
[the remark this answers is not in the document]
jamessynge: It should be, but that forces a circular import problem: PanBase uses PanDB, PanDB uses error, error uses PanBase.

I suggest that we consider removing the PanBase base class from error.py.

pocs/utils/database.py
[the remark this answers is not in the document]
wtgee: As per our private conversation, let's maybe look at wrapping this up into StrFormatLogRecord. Fine to leave in this PR for now.

pocs/utils/database.py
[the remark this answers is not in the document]
wtgee: I think that is probably a good suggestion and a cleaner way to do things as we can get hit with that circular problem elsewhere otherwise.

PR #468, opened 2018-02-10
Use threads instead of processes in test_pocs.py

jamessynge:
Run pocs in a thread instead of in a separate process to make it easier
to share resources between the test thread and pocs. This was motivated
by PanMemoryDB, but I think it will help in general.

Add a Timeout class to pocs.utils and use in test_pocs to force the
ending of long running tests.

Shutdown pocs even if message not received; use a try-finally block
to ensure we send the shutdown message.

### DIFF
--- a/pocs/tests/test_pocs.py
+++ b/pocs/tests/test_pocs.py
@@ -1,15 +1,35 @@
 import os
 import pytest
 import time
-
-from multiprocessing import Process
+import threading
 
 from astropy import units as u
 
 from pocs import hardware
 from pocs.core import POCS
 from pocs.observatory import Observatory
+from pocs.utils import Timeout
 from pocs.utils.messaging import PanMessaging
+
+
+def wait_for_running(sub, max_duration=90):
+    """Given a message subscriber, wait for a RUNNING message."""
+    timeout = Timeout(max_duration)
+    while not timeout.expired():
+        msg_type, msg_obj = sub.receive_message()
+        if msg_obj and 'RUNNING' == msg_obj.get('message'):
+            return True
+    return False
+
+
+def wait_for_pointing(sub, max_duration=90):
+    """Given a message subscriber, wait for the state to be `pointing`."""
+    timeout = Timeout(max_duration)
+    while not timeout.expired():
+        msg_type, msg_obj = sub.receive_message()
+        if msg_type == 'STATUS' and msg_obj and 'pointing' == msg_obj.get('state'):
+            return True
+    return False
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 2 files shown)


pocs/tests/test_pocs.py
```diff
@@ -1,17 +1,37 @@
...
+    """Given a message subscriber, wait for a RUNNING message."""
+    timeout = Timeout(max_duration)
+    while not timeout.expired():
+        msg_type, msg_obj = sub.receive_message()
+        if msg_obj and 'RUNNING' == msg_obj.get('message'):
+            return True
+    return False
+
+
+def wait_for_pointing(sub, max_duration=90):
```
wtgee: The MemoryDB PR has a note about perhaps changing which state we check for so we don't run through scheduling.

pocs/utils/__init__.py
```diff
@@ -48,6 +48,42 @@ def flatten_time(t):
...
+    def expired(self):
+        """Return a boolean, telling if the timeout has expired."""
+        return self.time_left() <= 0
+
+    def time_left(self):
+        """Return how many seconds are left until the timeout expires."""
+        if self.is_non_blocking:
+            return 0
+        else:
+            delta = self.target_time - time.monotonic()
```
wtgee: Am I missing something about where these come from? Should this class inherit from `serialutil.Timeout`?

pocs/utils/__init__.py
```diff
@@ -48,6 +48,42 @@ def flatten_time(t):
     return t.isot.replace('-', '').replace(':', '').split('.')[0]
 
 
+# This is a streamlined variant of PySerial's serialutil.Timeout.
+class Timeout(object):
+    """Simple timer object for tracking whether a time duration has elapsed.
+
+    Attribute `is_non_blocking` is true IFF the duration is zero.
```
wtgee: I don't understand where non-blocking case. Where would you be using a timeout=0?

pocs/utils/__init__.py
jamessynge: I decided to just copy their code and eliminate things we don't need. time.monotonic() is like time.time(), but guarantees that it won't go backwards.

pocs/tests/test_pocs.py
jamessynge: Yup, I've incorporated that in this branch, pushing in a few minutes.

pocs/utils/__init__.py
jamessynge: Yes, timeout==0 means that it will always be considered to have expired.

jamessynge:
This change is really about simplifying #463 which had gotten too large, reflecting the various bugs I found along the while debugging use of the memory db.

pocs/utils/__init__.py
wtgee: Ok, I didn't realize `target_time` first gets set in the `restart` method so didn't see where it was coming from. Makes sense now.

wtgee: Although maybe just `start` is a better name?

jamessynge: restart because __init__ calls it, so that it is started at creation. If too confusing, I could remove restart, so you have to create a new instance each time, which is definitely the common way to use it.

PR #475, opened 2018-02-12
Add scripts for starting POCS automatically - WIP

jamessynge:
These scripts are designed for running:

* ZeroMQ message forwarders
* PAWS Web Server
* PEAS Shell
* POCS Shell

They are each run in a separate terminal within
tmux, a terminal multiplexer.

These were originally developed for PAN006, and this PR
is for making them more generally applicable.

This is a WIP because I've not yet installed them on
PAN006 and tested them.

### DIFF
--- a/scripts/startup/README.md
+++ b/scripts/startup/README.md
@@ -1,0 +1,47 @@
+These scripts support starting the PANOPTES software automatically
+when the Linux computer that runs a PANOPTES starts or restarts.
+
+First we need to have the PANOPTES environement variables set in all
+the shells that are involved in running PANOPTES. The simplest
+solution is to set them globally on the system. For example,
+on Ubuntu:
+
+```bash
+sudo cp $POCS/scripts/startup/set_panoptes_env_vars.sh \
+        /etc/profile.d/
+sudo chmod +x /etc/profile.d/set_panoptes_env_vars.sh
+```
+
+This will then be executed when any user logs in. The exact means
+of doing this varies on different versions of Unix/Linux.
+
+Next, we need to arrange for `su_panoptes.sh` to be executed when
+the operating system launches. First copy it to /etc:
+
+```bash
+sudo cp $POCS/scripts/startup/su_panoptes.sh \
+        /etc/
+sudo chmod +x /etc/su_panoptes.sh
+```
+
+Next, edit `/etc/rc.local`, adding a command to run `su_panoptes.sh`.
+For example, adding it to the default Ubuntu `rc.local`:
+
+```bash
+#!/bin/sh -e
+#
+# rc.local
+#
+# This script is executed at the end of each multiuser runlevel.
+# Make sure that the script will "exit 0" on success or any other
+# value on error.
+#
+# In order to enable or disable this script just change the execution
+# bits.
+#
+# By default this script does nothing.
+
+/etc/su_panoptes.sh >> /tmp/su_panoptes.log 2>&1
+
+exit 0
+```
--- a/scripts/startup/set_panoptes_env_vars.sh
+++ b/scripts/startup/set_panoptes_env_vars.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+# This file is to be copied into /etc/profile.d/ and SOURCED by
+# /etc/rc.local. We have the shebang line at the top to help shellcheck
+# recognize the type of script.
+
+export PANUSER=panoptes          # User that runs PANOPTES software.
+export PANDIR=/var/panoptes      # Main directory.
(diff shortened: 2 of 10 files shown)


scripts/startup/README.md
```diff
@@ -0,0 +1,47 @@
...
+sudo cp $POCS/scripts/startup/set_panoptes_env_vars.sh \
+        /etc/profile.d/
+sudo chmod +x /etc/profile.d/set_panoptes_env_vars.sh
+```
+
+This will then be executed when any user logs in. The exact means
+of doing this varies on different versions of Unix/Linux.
+
+Next, we need to arrange for `su_panoptes.sh` to be executed when
+the operating system launches. First copy it to /etc:
```
wtgee: Can we symlink this in case it changes?

scripts/startup/set_panoptes_env_vars.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
wtgee: Why not the same as `scripts/install/default-env-vars.sh` (which has your `-z` additions)?

scripts/startup/start_peas.sh
```diff
@@ -0,0 +1,28 @@
...
+sleep 5s
+tmux send-keys -t "${WINDOW}" "load_weather" C-m
+sleep 5s
+tmux send-keys -t "${WINDOW}" "start" C-m
+sleep 20s
+tmux send-keys -t "${WINDOW}" "status" C-m
+sleep 10s
+tmux send-keys -t "${WINDOW}" "last_reading environment" C-m
+sleep 10s
+tmux send-keys -t "${WINDOW}" "last_reading weather" C-m
```
wtgee: Note that the first weather reading usually pulls across `Undefined` for all the values which I suppose could cause alarm when shown here.

scripts/startup/README.md
jamessynge: That is a security hole, as it allows the file to be changed without entering the password. For that reason, I prefer to copy it. The files that run as $PANUSER are not copied.

scripts/startup/set_panoptes_env_vars.sh
jamessynge: Because I'm an idiot? I should have suggested that you document the file as intended to be modified to suit local conditions and then copied into /etc/profile.d/.

As it is, `default-env-vars.sh` is well suited to running from $HOME/.profile at a point AFTER the user has provided any non-standard values.

scripts/startup/start_peas.sh
jamessynge: How often do we get readings? 30 seconds? A minute? As it stands, it runs about 40 seconds after starting.

scripts/startup/README.md
wtgee: Right, got it.

scripts/startup/start_peas.sh
wtgee: The first reading takes longer and comes back with the undefined values. I find it's usually about 2 minutes before any valid weather reading. Could just leave as is and we can watch it to see if it looks okay.

scripts/startup/start_log_viewer.sh
```diff
@@ -0,0 +1,17 @@
...
+LOGFILE="${2}"
+echo "Running $(basename "${0}") at $(date), WINDOW=${WINDOW}, LOGFILE=${LOGFILE}"
+
+# Wait for bash to be ready (not necessary, but makes
+# the window look tidier when you attach later).
+sleep 1s
+
+tmux send-keys -t "${WINDOW}" "date" C-m
+tmux send-keys -t "${WINDOW}" "cd \"${PANLOG}\"" C-m
+tmux send-keys -t "${WINDOW}" "less --follow-name \"${LOGFILE}\"" C-m
```
wtgee: I'll probably later submit a patch to this to add `grc` in front of the `less` (with appropriate checks if it exists) as I think its nice to have colorized log files. However that requires having a proper `grc` config file since we don't show standard log messages.

scripts/startup/start_peas.sh
jamessynge: Sure, will leave as is.

wtgee:
One thing I like about byobu is the default status bar is much more helpful and user friendly. Will this display a status bar with names of windows?

wtgee:
Linking #97.

jamessynge:
Yes, this will show the names of the windows, along with their numbers. I use ^B# (where # is the number of the window) to switch to a window.

jamessynge:
Submitting now, even as a WIP, so that it is easy to install on PAN006 for testing.

Issue #478, opened 2018-02-13
Camera simulator


wtgee:
Currently the camera simulator generates some random data but then changes the filename so that the `solved.fits` is used during processing. This is having some strange side-effects in various places. Additionally, trying to just `run` pocs with the simulators leads to a solve-field error which means it hangs after the `pointing` state. 

I would like to be able to do a normal run with POCS (for instance, with a unit set up) and have it run through a full observation cycle, even showing the fake data in PAWS.

More generally, the simulator code needs to be fixed up a bit and made consistent.

AnthonyHorton:
The camera field is only supposed to generate random data if its been asked for a dark frame. Otherwise it should just copy the data from `unsolved.fits`, generate a set of headers for it, & save the result to a FITS file in the normal location.

wtgee:
> The camera simulator is only supposed to generate random data if its been asked for a dark frame. Otherwise it should just copy the data from unsolved.fits, generate a set of headers for it, & save the result to a FITS file in the normal location.

Good to know. I was experiencing some (intermittent) issues and @jamessynge reported the same thing, so this is mostly meant as a way to explore that. While I was working on some of the DB abstraction I noticed failures in the logs that were not causing test failures.

This may be a non-issue. I'll look into what I was experiencing. 

For the record, here is what James reported:

> I'm debugging tests with my new PanMemoryDB, and in test_pocs.py am down to one failure. If I run that by itself and study the log file, I notice that the camera (simulator) doesn't seem to work right (or at least not reliably.

> So, first, what should happen with "taking" an exposure in that setting?

> And do you have any idea what the cause of the file errors below could be? If they are expected because of the simulator being incomplete, perhaps we can fix that so that it is less difficult to debug.

> Note that on my laptop, there is no /var/panoptes/images/fields directory.

```
D0209 23:58:35.321 simulator.py:79      Taking 5 second exposure on Cam00: /var/panoptes/POCS/pocs/tests/data/solved.fits
D0209 23:58:35.323 pointing.py:59       Waiting for images: 0.0 seconds
D0209 23:58:35.324 simulator.py:79      Getting mount simulator status
D0209 23:58:39.436 pointing.py:59       Waiting for images: 3.0 seconds
D0209 23:58:39.437 simulator.py:79      Getting mount simulator status
E0209 23:58:40.346 fits.py:338          Error writing image to /var/panoptes/POCS/pocs/tests/data/solved.fits!
[...]

Issue #191, opened 2017-12-14
Local test execution is too slow.


jamessynge:
We want folks to be able to run the tests easily and often. Unfortunately a number of them are quite slow, including the new astrohaven test.

jamessynge:
Here are the 20 slowest tests from a recent run of `pytest --durations=20`
```
================== slowest 20 test durations ==================
60.89s call     pocs/tests/test_pocs.py::test_run
40.44s call     pocs/tests/test_pocs.py::test_run_wait_until_safe
31.67s call     pocs/tests/test_pocs.py::test_run_no_targets_and_exit
15.12s call     pocs/tests/test_dispatch_scheduler.py::test_set_observation_then_reset
14.87s call     pocs/tests/test_observatory.py::test_status
11.26s call     pocs/tests/test_observatory.py::test_get_observation
11.21s call     pocs/tests/test_dispatch_scheduler.py::test_observed_list
11.20s call     pocs/tests/test_dispatch_scheduler.py::test_continue_observation
10.12s call     pocs/tests/test_pocs.py::test_run_power_down_interrupt
10.07s call     pocs/tests/test_pocs.py::test_run_interrupt_with_reschedule_of_target
9.14s call     pocs/tests/test_pocs.py::test_unsafe_park
7.83s call     pocs/tests/test_dispatch_scheduler.py::test_new_observation_seq_time
7.03s call     pocs/tests/test_camera.py::test_observation[simulator]
5.01s call     pocs/tests/test_camera.py::test_exposure_collision[simulator]
5.01s call     pocs/tests/test_camera.py::test_exposure[simulator]
4.81s call     pocs/tests/test_observatory.py::test_autofocus_coarse
4.35s call     pocs/tests/test_pocs.py::test_power_down_while_running
3.93s call     pocs/tests/test_camera.py::test_autofocus_coarse[simulator]
3.79s call     pocs/tests/test_dispatch_scheduler.py::test_get_observation
3.79s call     pocs/tests/test_dispatch_scheduler.py::test_reset_observation
```

jamessynge:
The situation is getting worse. We need to figure out how to streamline the process. It used to be under 5 minutes to run the tests, but I think test_pocs.py now takes longer than that.

As a start, we need to figure out where the time is going.

wtgee:
Yes, the addition of PanFileDB made the tests much longe, which was mentioned in #414 as the reason I was hesitant to merge it.

[...]

PR #479, opened 2018-02-13
Fix camera simulator

AnthonyHorton:
Fix for bugs discussed in https://github.com/panoptes/POCS/issues/478

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -184,14 +184,16 @@ class AbstractCamera(PanBase):
         Args:
             observation (~pocs.scheduler.observation.Observation): Object
                 describing the observation
-            headers (dict): Header data to be saved along with the file.
+            headers (dict, optional): Header data to be saved along with the file.
+            filename (str, optional): pass a filename for the output FITS file to
+                overrride the default file naming system
             **kwargs (dict): Optional keyword arguments (`exp_time`, dark)
 
         Returns:
             threading.Event: An event to be set when the image is done processing
         """
         # To be used for marking when exposure is complete (see `process_exposure`)
-        camera_event = Event()
+        observation_event = Event()
 
         exp_time, file_path, image_id, metadata = self._setup_observation(observation,
                                                                           headers,
@@ -206,16 +208,16 @@ class AbstractCamera(PanBase):
             observation.exposure_list[image_id] = file_path
 
         # Process the exposure once readout is complete
-        t = Thread(target=self.process_exposure, args=(metadata, camera_event, exposure_event))
+        t = Thread(target=self.process_exposure, args=(metadata, observation_event, exposure_event))
         t.name = '{}Thread'.format(self.name)
         t.start()
 
-        return camera_event
+        return observation_event
 
     def take_exposure(self, *args, **kwargs):
         raise NotImplementedError
 
-    def process_exposure(self, info, signal_event, exposure_event=None):
+    def process_exposure(self, info, observation_event, exposure_event=None):
         """
         Processes the exposure.
 
@@ -224,7 +226,7 @@ class AbstractCamera(PanBase):
 
         Args:
             info (dict): Header metadata saved for the image
-            signal_event (threading.Event): An event that is set signifying that the
+            observation_event (threading.Event): An event that is set signifying that the
                 camera is done with this exposure
             exposure_event (threading.Event, optional): An event that should be set
                 when the exposure is complete, triggering the processing.
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -268,7 +268,7 @@ class AbstractCamera(PanBase):
         })
 
         # Mark the event as done
-        signal_event.set()
+        observation_event.set()
 
     def autofocus(self,
                   seconds=None,
--- a/pocs/camera/simulator.py
+++ b/pocs/camera/simulator.py
@@ -26,36 +26,23 @@ class Camera(AbstractCamera):
         The simulator merely markes the `connected` property.
         """
         # Create a random serial number
-        self._serial_number = 'SC{:4d}'.format(random.randint(0, 9999))
+        self._serial_number = 'SC{:04d}'.format(random.randint(0, 9999))
 
         self._connected = True
         self.logger.debug('{} connected'.format(self.name))
 
     def take_observation(self, observation, headers=None, filename=None, *args, **kwargs):
-        camera_event = Event()
 
-        exp_time, file_path, image_id, metadata = self._setup_observation(observation,
-                                                                          headers,
-                                                                          filename,
-                                                                          *args,
-                                                                          **kwargs)
-
+        exp_time = kwargs.get('exp_time', observation.exp_time.value)
         if exp_time > 5:
-            exp_time = 5
+            kwargs['exp_time'] = 5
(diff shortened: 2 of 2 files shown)


wtgee:
I'm going to pull this down and run it to see if I can find the same messages in the logs that I was experiencing the other day. Will approve after that. Thanks for getting in!

wtgee:
Wow, coverage up by 2.1%. Yes, I think this was causing errors in the code but not failing the tests, so we must now be getting further on in tests. 👍

wtgee:
Ok, I think I see the issue and it has to do with a bad naming choice on my part. In the `_fake_exposure` method you are grabbing the `unsolved.fits ([here](https://github.com/panoptes/POCS/blob/develop/pocs/camera/simulator.py#L100)), but that it is a fits file that should really be named `unsolvable.fits` as it is the one we use to test what happens when solving fails. So when I run pocs with all simulators (via pocs_shell or jupyter) it will fail on the solve in the `pointing` state and so shut itself down. 

There is a valid question to be asked of whether or not we should shut down if solving fails or not, but I think for here and now we should change the file to `solved.fits` that it grabs data from as then solving will continue. @anthonyhorton, is there any reason you would want those fake exposures to not be solving?

AnthonyHorton:
@wtgee No, I don't have any reasons to want the solves to fail. I'll change it so it copies the data from `solved.fits` instead.

wtgee:
There's still actually a couple of things going on:

* The headers are not copied but are generated from the current observation yet saved with the data from `solved.fits`. This is then suggesting an RA/Dec to solve-field that doesn't match the data, which is causing it to timeout without solving (which takes ~30 seconds and may be contributing partially to #191).
* Simulator images are saved in `$PANDIR/images/fields` under the camera simulator name but are (sometimes?) not properly cleaned up.

AnthonyHorton:
Ah, I see. Yes, the simulator doesn't copy headers from `solved.fits` because I wanted the simulator to use as much of the real camera code as possible, including header generation. I could just overwrite the mount coordinates values with the ones from

wtgee:
I tried to just override the mount coordinates in the header but didn't seem to help. I'll do a little more exploring and then report back.

Also not entirely sure what to do about cleanup. I think it's better to have them going to the same location but it's a lot of management to clean up. Not terrible during tests but it means anyone using a simulator (e.g. manually) has to know to clean them up. Could be handled as part of `Observatory.cleanup_observations_dir` but that requires running that too. I'll give it a little more thought.

wtgee:
Ok, one problem which might explain the intermittent behavior is the name of the camera simulator. You are assigning a random int [here](/var/panoptes/images/fields/Wasp2/SC9593/20180213T221404//pointing00.fits) but the format string doesn't have leading zeros, so if number is less than 1000 it ends up with a space in the name, which breaks things. Changing formatting to `{:04d}` solves this issue.

wtgee:
I ended up overwriting the coords in `_fake_exposure` with:

```
solved_header = fits.getheader(file_path)
header['RA-MNT'] = solved_header['RA-MNT']
header['HA-MNT'] = solved_header['HA-MNT']
header['DEC-MNT'] = solved_header['DEC-MNT']
``` 

And then adding a copy of `_process_fits` to the simulator but remove the call to `update_headers`. Then things appear to be working ok.

wtgee:
Here's a patch if you want.

[patch.txt](https://github.com/panoptes/POCS/files/1722011/patch.txt)

wtgee:
One other change, the `simulator.take_observation` needs an `is_primary` for adding to exposure list:

```
        # Add most recent exposure to list
        if self.is_primary:
            observation.exposure_list[image_id] = file_path
```

AnthonyHorton:
Ah, yes, the mount coordinates are not normally added until `process_exposure()` calls `_process_fits()`, which calls `update_headers()`, so any values added in `_fake_exposure` will get overwritten.

As an alternative to overriding `process_fits()` to remove the `update_headers()` call I could instead call `update_headers()` then overwrite the mount coordinates.  Trying to exercise as much of the real camera code as possible with the simulator to keep the test coverage up.

AnthonyHorton:
One thing confusing me, when & how does the field solving get triggered?

AnthonyHorton:
OK, I _think_ this should work now. With the addition of the `is_primary` check to `take_observation()` it became almost identical to the base class `take_observation` method. The only difference is the exposure time trimming so now I just do that then call `super().take_observation()`.

I've also added an override for `_process_fits()` that calls `super()._process_fits()` then overwrites with mount coordinates header entries with those from `solved.fits`.

I also took the opportunity to update a docstring and improve the logic/consistency of a couple of variable names in `camera.py`.

pocs/camera/simulator.py
```diff
@@ -107,3 +93,14 @@ def _fake_exposure(self, filename, header, exposure_event):
                                           dtype=fake_data.dtype)
 
         fits_utils.write_fits(fake_data, header, filename, self.logger, exposure_event)
+
+    def _process_fits(sefl, file_path, info):
```
wtgee: `sefl` -> `self`

pocs/camera/camera.py
```diff
@@ -224,7 +226,7 @@ def process_exposure(self, info, signal_event, exposure_event=None):
 
         Args:
             info (dict): Header metadata saved for the image
-            signal_event (threading.Event): An event that is set signifying that the
+            observation_event (threading.Event): An event that is set signifying that the
```
wtgee: Line 271 needs to get changed to match. Won't let me comment on it here.

pocs/camera/simulator.py
```diff
@@ -107,3 +93,14 @@ def _fake_exposure(self, filename, header, exposure_event):
...
+        super()._process_fits(file_path, info)
+        self.logger.debug('Overriding mount coordinates for camera simulator')
+        solved_path = "{}/pocs/tests/data/{}".format(os.getenv('POCS'), 'solved.fits')
+        solved_header = fits.getheader(solved_path)
+        with fits.open(file_path, 'update') as f:
+            hdu = f[0]
+            hdu.header.set('RA-MNT', solved_header['RA-MNT'], 'Degrees')
+            hdu.header.set('HA-MNT', solved_header['HA-MNT'], 'Degrees')
+            hdu.header.set('DEC-MNT', solved_header['DEC-MNT'], 'Degrees')
```
wtgee: Need to `return file_path`

pocs/camera/simulator.py
```diff
@@ -26,37 +26,23 @@ def connect(self):
...
     def take_observation(self, observation, headers=None, filename=None, *args, **kwargs):
-        camera_event = Event()
 
-        exp_time, file_path, image_id, metadata = self._setup_observation(observation,
-                                                                          headers,
-                                                                          filename,
-                                                                          *args,
-                                                                          **kwargs)
+        exp_time = kwargs.get('exp_time', observation.exp_time.value)
+        if exp_time > 5:
```
wtgee: Can you change this to `2` while you are at it? I just tested and works fine.

AnthonyHorton:
+2.62%!

wtgee:
Awesome. Looks good for merging to me. Anything else you wanted in there?

AnthonyHorton:
No, pretty happy with it now.

Issue #450, opened 2018-02-06
catch camera timeout


lspitler:
For SBIGs it seems to happen on readout.

wtgee:
In general we perhaps need something in the `camera.process_exposure()` method that times out events after a while (probably set up in the `take_exposure` methods).

Note that for Huntsman specifically we have an omission in the `observing` state where we are incrementing the timer but never acting on it. See huntsman observing [here](https://github.com/AstroHuntsman/huntsman-pocs/blob/develop/states/huntsman/observing.py#L23) compared to the breakout in the panoptes observing [here](https://github.com/panoptes/POCS/blob/develop/pocs/state/states/default/observing.py#L28).

wtgee:
Also just a note that we should obviously fix the Huntsman observing state but we experience them most often in the autofocus routines.

AnthonyHorton:
Some attempts to handle timeouts more gracefully in https://github.com/panoptes/POCS/pull/458

AnthonyHorton:
Could use some fresh eyes on this.  Based on the logs it looks like execution is getting stuck in this section of code in `pocs/camera/sbigudrv.py`:

```
        # Readout data
        with self._command_lock:
            self._set_handle(handle)
            self._send_command('CC_END_EXPOSURE', params=end_exposure_params)
            self._send_command('CC_START_READOUT', params=start_readout_params)
            for i in range(height):
                try:
                    self._send_command('CC_READOUT_LINE',
                                       params=readout_line_params,
                                       results=as_ctypes(image_data[i]))
                except RuntimeError as err:
                    message = 'Readout error on {}: expected {} rows, got{}!'.format(self._ccd_info[handle]['serial number'],
                                                                                     height,
                                                                                     i)
                    self.logger.error(message)
                    self.logger.error(err)
                    warn(message)
                    break

            try:
                self._send_command('CC_END_READOUT', params=end_readout_params)
            except RuntimeError as err:
                message = "Error ending readout on {}: {}".format(self._ccd_info[handle]['serial number'],
                                                                  err)
                self.logger.error(message)
            else:
                self.logger.debug('Readout on {} complete'.format(self._ccd_info[handle]['serial number']))
            finally:
                fits_utils.write_fits(image_data, header, filename, self.logger, exposure_event)
```

[...]

PR #481, opened 2018-02-13
SBIG camera debug messages bug fix

AnthonyHorton:
Fixes a bug introduced in fb82559afd8b34f5cfb3b64b771de3a6f0545f71, & adds an additional debug message in an attempt to help work out the cause of the issues discussed in https://github.com/panoptes/POCS/issues/450


AnthonyHorton:
This bug was fixed on the fly on the Huntsman control computer, but remains upstream. Further fixes will are expected once we work out the causes of the issues in #450, but I thought it would be good to just get this small fix merged in the meantime.

PR #485, opened 2018-02-15
Improve logging and speed of startup scripts

jamessynge:
Create a directory for each tmux_launch execution, used
by all of the startup scripts it launches.

Run the scripts for initializing each of the terminals
in a separate process and don't wait for them to complete.
This will speed the process of starting up the system.

Add -ex to all of the bash shebang lines so that the log
files show what was being executed, and so that bash
errors are detected.

### DIFF
--- a/scripts/startup/README.md
+++ b/scripts/startup/README.md
@@ -74,8 +74,8 @@ you add that rule at the end of the file, save the file and exit the
 editor. You can confirm that it was added correctly by viewing the
 file with the command `crontab -l`. Next, reboot the computer (e.g.
 with the command `sudo shutdown -r now`), log back in, remove that
-test rule, and see if /tmp/at-reboot-worked was created. If it was,
-then `@reboot` works. Otherwise, skip the @reboot rule in the
+test rule, and see if `/tmp/at-reboot-worked` was created. If it was,
+then `@reboot` works. Otherwise, skip the `@reboot` rule in the
 steps that follow, and follow the instructions in 
 __Execution by `root`__ after editing your crontab.
 
--- a/scripts/startup/README.md
+++ b/scripts/startup/README.md
@@ -3,6 +3,17 @@ when the Linux computer that runs a PANOPTES starts or restarts.
 We assume here that you are logged in as user $PANUSER, and that user
 $PANUSER has permissions to execute `sudo`.
 
+> Note: We use `tmux`, a [terminal
+> multiplexer](https://en.wikipedia.org/wiki/Terminal_multiplexer),
+> to run all of the software. This means that when the computer boots,
+> it can create several virtual terminals, each running a separate
+> program. Later, when you want to see what those programs are doing,
+> you can login and attach to those terminals. With `tmux`, the command
+> to do that is:
+> ```bash
+> tmux attach-session -t panoptes
+> ```
+
 # PANOPTES Environement Variables
 
 First we need to have the PANOPTES environement variables set in all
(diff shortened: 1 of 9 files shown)


jamessynge:
It should have a several second head start on PAWS.

scripts/startup/README.md
```diff
@@ -15,8 +19,55 @@ sudo chmod +x /etc/profile.d/set_panoptes_env_vars.sh
...
+`crontab -e`, which will start an editor on your crontab file.
+
+```bash
+POCS=/var/panoptes/POCS
+PANLOG=/var/panoptes/logs
+
+# m h  dom mon dow   command
+@reboot              /bin/bash --login $POCS/scripts/startup/tmux_launch.sh
+*/5 *   *   *   *    /bin/bash --login python $POCS/scripts/plot_weather.py >> $PANLOG/plot_weather.log 2>&1
+11  12  *   *   *    /bin/bash --login python $POCS/pocs/utils/data.py >> $PANLOG/update_data.log 2>&1
```
wtgee: The other wiki page about this only has it running once per week (i.e. `dow 0`). Did you mean for it to run every day? Doesn't hurt I suppose but not totally necessary.

scripts/startup/README.md
```diff
@@ -15,8 +19,55 @@ sudo chmod +x /etc/profile.d/set_panoptes_env_vars.sh
...
+```
+
+If your values for POCS and PANLOG don't match those show, please
```
wtgee: show -> shown (? I'm assuming that's what you meant, but maybe "shown above" would be a little more clear).

scripts/startup/README.md
```diff
@@ -15,8 +19,55 @@ sudo chmod +x /etc/profile.d/set_panoptes_env_vars.sh
...
+edit them appropriately. Use the fully evaluated values (e.g. the
+value produced by `echo $POCS` and by `echo $PANLOG`), as `cron`
+does not expand variables in those lines.
+
+Save the edited file and exit the editor.
```
wtgee: Do we need a `vim` difficulty warning? I'm not sure what default editor on Ubuntu is right now.

scripts/startup/README.md
```diff
@@ -15,8 +19,55 @@ sudo chmod +x /etc/profile.d/set_panoptes_env_vars.sh
...
+
+Notice that each of these commands starts with `/bin/bash --login`. We
+do this because `cron` does not run the commands in a shell with all
+of the normal environment variables set, so we force that to happen.
+
+## Execution by `root`
+
+If necessary, you can modify the Linux boot scripts to start the
```
wtgee: Explain when "if necessary" applies, i.e. "if your crontab does not support the reboot directive". Of course then we will just get questions about which ones do and do not. I'm guessing it's not as simple as a version number but rather which implementation?

wtgee: We should figure out if works on ubuntu 16.04 (and maybe upcoming 18.04) and highlight that.

scripts/startup/tmux_launch.sh
```diff
@@ -34,14 +46,20 @@ echo "PANDIR: ${PANDIR}"
...
+
 # $- expands to the current option flags as specified upon invocation, by the
 # set built-in command, or those set by the shell itself (such as the -i).
-echo '$-:' "$-"
+echo 'Shell options ($-):' "$-"
+
+# These two environment variables don't have to be set, but let's record
+# their values as they may impact things.
+echo "PIAA: ${PIAA}"
+echo "MATPLOTLIBRC: ${MATPLOTLIBRC}"
```
wtgee: We move the file into the base of $POCS and are no longer setting this so is probably best to remove.

scripts/startup/tmux_launch.sh
```diff
@@ -1,7 +1,15 @@
 #!/bin/bash -ex
 
-# This script is designed to be run from /etc/rc.local (or similar)
-# as user $PANUSER with the PANOPTES environment variables set.
+# Start the PANOPTES software in a detached tmux (terminal multiplexer)
+# session; this enables the PANOPTES unit's owner/administrator to
+# attach to that session later, see the output and take control of all
+# the shells.
```
wtgee: Might be worth included an example of how to attach to session.

jamessynge: Done, here and in README.md.

scripts/startup/tmux_launch.sh
jamessynge: Done.

scripts/startup/README.md
```diff
@@ -31,15 +42,45 @@ reboots. We document these ways to achieve this:
 * `@reboot` rule in crontab of user $PANUSER
 * Execution by root during boot
 
+When you are ready to have the telescope run automatically (i.e. have
```
wtgee: Probably shouldn't actually use the word "telescope". Maybe "PANOTPES unit"?

scripts/startup/README.md
```diff
@@ -31,15 +42,45 @@ reboots. We document these ways to achieve this:
...
+```
+
+Add the rule to your ($PANUSER) crontab by executing the command
+`crontab -e`, which will start an editor on your crontab file. After
+you add that rule at the end of the file, save the file and exit the
+editor. You can confirm that it was added correctly by viewing the
+file with the command `crontab -l`. Next, reboot the computer (e.g.
+with the command `sudo shutdown -r now`), log back in, remove that
+test rule, and see if /tmp/at-reboot-worked was created. If it was,
+then `@reboot` works. Otherwise, skip the @reboot rule in the
```
wtgee: `@reboot` in back-ticks

scripts/startup/README.md
```diff
@@ -31,15 +42,45 @@ reboots. We document these ways to achieve this:
...
+@reboot touch /tmp/at-reboot-worked
+```
+
+Add the rule to your ($PANUSER) crontab by executing the command
+`crontab -e`, which will start an editor on your crontab file. After
+you add that rule at the end of the file, save the file and exit the
+editor. You can confirm that it was added correctly by viewing the
+file with the command `crontab -l`. Next, reboot the computer (e.g.
+with the command `sudo shutdown -r now`), log back in, remove that
+test rule, and see if /tmp/at-reboot-worked was created. If it was,
```
wtgee: File name in back-ticks

PR #488, opened 2018-02-16
Default to rereading the fields file

### LINKED ISSUE #344: Fix reread_fields_file
* Make an option in config file
* Make as part of base scheduler, not specific to dispatch

wtgee:
It's annoying and inconvenient to have to shut down POCS just to read new targets. This changes the default to always reread the target file before scheduling from the `scheduling` state.

Closes #344


pocs/scheduler/dispatch.py
```diff
@@ -30,6 +30,8 @@ def get_observation(self, time=None, show_all=False, reread_fields_file=False):
                 defaults to time called
             show_all (bool, optional): Return all valid observations along with
                 merit value, defaults to False to only get top value
+            reread_fields_file (bool, optional): If the fields file should be reread
```
jamessynge: And the reason for not making the change here in the params list (i.e. default =True).

wtgee:
@jamessynge I changed this to a config option, which fixed the tests (coverage went down 2.5% the other way). Let me know what you think of this setup. Thanks.

PR #492, opened 2018-02-19
Ensure that bash is used by tmux

jamessynge:
When cron runs an @reboot rule, only a minimal environment is provided,
and SHELL=/bin/sh. This means that without taking steps, tmux windows
will have sh as their shell, while our startup scripts assume that
the windows are running bash.

### DIFF
--- a/scripts/startup/start_panoptes_in_tmux.sh
+++ b/scripts/startup/start_panoptes_in_tmux.sh
@@ -1,10 +1,23 @@
 #!/bin/bash -ex
 
+if [[ -z "$BASH_VERSION" ]] ; then
+  echo "This script must be run by bash."
+  exit 1
+fi
+
 # This script is designed to be run inside tmux, launched by
 # tmux_launch.sh. Make sure that is so.
 echo "Running ${BASH_SOURCE[0]} at $(date)"
 if [[ -z "${STARTUP_LOG_DIR_SLASH}" ]] ; then
   echo "This script must be run by tmux_launch.sh"
+  exit 1
+fi
+
+# And we need the shell to have the the PANOPTES environment
+# setup.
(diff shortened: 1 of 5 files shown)


scripts/startup/start_panoptes_in_tmux.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash -ex
 
+if [[ -z "$BASH_VERSION" ]] ; then
```
wtgee: If you have the bash shebang then why is this necessary?

jamessynge: Here I probably don't, but I was struggling to diagnose the failures. If I debugged this correctly, it came down to the `tmux new-window` command using `$SHELL` and `bash --login` not setting `$SHELL`.

`start_panoptes_in_tmux.sh` is executed by bash because of the shebang, but the windows (i.e. virtual terminals) that it creates using `tmux new-window` don't have a command specified, so just run the default shell. Those windows are then sent commands that are designed to work in a bash shell with POCS, PANDIR, etc. all set, and the conda panoptes-env activated. Without the change in tmux_launch.sh, the default shell ends up being /bin/sh, which doesn't support the commands that are sent to the window.

jamessynge:
I suspect that the shebang line in `tmux_launch.sh` is ignored because I'm using this crontab line to start it:
```bash
@reboot              /bin/bash --login $POCS/scripts/startup/tmux_launch.sh >> $PANLOG/tmux_launch.cron-reboot.log 2>&1
```
I started the command with `/bin/bash --login` to get the environment fully setup. The only argument is a file, and the bash `-c` option wasn't provided, so bash runs that file as a shell source file and doesn't treat that argument as a command line. Therefore, it does not try to interpret the first non-option argument as an executable file.

I've updated the docs & comments to provide some more detail about this.

PR #499, opened 2018-02-26
Global formatting with yapf

jamessynge:


pocs/camera/camera.py
```diff
@@ -242,9 +241,8 @@ def process_exposure(self, info, observation_event, exposure_event=None):
 
         try:
             self.logger.debug("Extracting pretty image")
-            img_utils.make_pretty_image(file_path,
-                                        title=info['field_name'],
-                                        primary=info['is_primary'])
+            img_utils.make_pretty_image(
```
wtgee: These kind of changes are the only ones I'm not super keen on, but I think that's a throw-back to my Perl days. I like to have the params lined up under the first param if possible. Not a big deal though.

pocs/camera/camera.py
```diff
@@ -521,15 +510,13 @@ def command(self, cmd):
                     stdout=subprocess.PIPE,
                     stderr=subprocess.STDOUT,
                     universal_newlines=True,
-                    shell=False
-                )
+                    shell=False)
```
wtgee: I'm also not the biggest fan of putting the closing parentheses on last line as opposed to aligned with opening of call.

jamessynge: This is from pep8:

When trailing commas are redundant, they are often helpful when a version control system is used, when a list of values, arguments or imported items is expected to be extended over time. The pattern is to put each value (etc.) on a line by itself, always adding a trailing comma, and add the close parenthesis/bracket/brace on the next line. However it does not make sense to have a trailing comma on the same line as the closing delimiter (except in the above case of singleton tuples).

```
Yes:

FILES = [
    'setup.cfg',
    'tox.ini',
    ]
initialize(FILES,
           error=True,
           )
```

jamessynge:
@wtgee Where do we stand on this issue? A bunch of files were previously committed with the wrong formatting due to an editor misconfiguration, IIRC, while this change is more sweeping (and clearly obsolete by now).

jamessynge:
This change is obviously never going to be applied as is. But, I'd still like to get to some consistent formatting. I'd like the opinion of @wtgee and @AnthonyHorton on doing this incrementally (i.e. not all at once). For example, I could ask which files folks are working on, and then format the other files.

Issue #500, opened 2018-02-26
Poor handling of serial I/O failure with Astrohaven dome 


jamessynge:
pocs_shell output:

```
POCS > run_pocs
Starting messaging
Command publisher started on port 6500
Command subscriber started on port 6501
Message subscriber started on port 6511
Message publisher started on port 6510
Starting POCS - Press Ctrl-c to interrupt
POCS stopped.
POCS > open_dome
Problem opening the dome: read failed: device reports readiness to read but returned no data (device disconnected or multiple access on port?)
POCS > exit
Shutting down POCS instance, please wait
Traceback (most recent call last):
  File "/home/panoptes/anaconda3/envs/panoptes-env/lib/python3.6/site-packages/serial/serialposix.py", line 501, in read
    'device reports readiness to read but returned no data '
serial.serialutil.SerialException: device reports readiness to read but returned no data (device disconnected or multiple access on port?)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bin/pocs_shell", line 899, in <module>
    PocsShell().cmdloop()
  File "/home/panoptes/anaconda3/envs/panoptes-env/lib/python3.6/cmd.py", line 138, in cmdloop
    stop = self.onecmd(line)
  File "/home/panoptes/anaconda3/envs/panoptes-env/lib/python3.6/cmd.py", line 217, in onecmd
    return func(arg)
  File "bin/pocs_shell", line 256, in do_exit
    self.do_power_down()
  File "bin/pocs_shell", line 332, in do_power_down
    self.pocs.power_down()
  File "/var/panoptes/POCS/pocs/core.py", line 227, in power_down
    if not self.observatory.close_dome():
  File "/var/panoptes/POCS/pocs/observatory.py", line 535, in close_dome
    if not self.dome.is_closed:
  File "/var/panoptes/POCS/pocs/dome/astrohaven.py", line 86, in is_closed
    v = self._read_latest_state()
  File "/var/panoptes/POCS/pocs/dome/astrohaven.py", line 128, in _read_latest_state
    data = self.serial.read_bytes(size=1)
  File "/var/panoptes/POCS/pocs/utils/rs232.py", line 189, in read_bytes
    return self.ser.read(size=size)
  File "/home/panoptes/anaconda3/envs/panoptes-env/lib/python3.6/site-packages/serial/serialposix.py", line 509, in read
    raise SerialException('read failed: {}'.format(e))
[...]

PR #501, opened 2018-02-28
A script for resetting a USB device

jamessynge:
May help with #500.

### DIFF
--- a/scripts/reset_usb_device.py
+++ b/scripts/reset_usb_device.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python
+
+# This is copied from: https://unix.stackexchange.com/a/412223
+# It has not yet been adapted in any particular way for PANOPTES
+# (i.e. no connection with our config files, no logging).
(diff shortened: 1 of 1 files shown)


jamessynge:
I did some basic tests, but not more. If you prefer, we can hold off on this and work (in our copious free time) to add more focused code of our own. Maybe we should add this to a list of tools that folks can consider using (e.g. a troubleshooting guide).

wtgee:
Pinging on this. I have used it on PAN001 and it worked well to reset a device. I'm okay with putting it into the repo. Maybe we should add `$POCS/scripts/utils/`? (for even more subfolders!)

jamessynge:
Thanks for testing. I don't yet see a need for a sub-folder for utils as we only have 14 entries in scripts... though I would consider adding a sub-folder for cron scripts (besides tmux_launch.sh which is already in the startup sub-folder).

@wtgee are you OK with committing this as is?

PR #502, opened 2018-03-04
Docker setup

### LINKED ISSUE #27: Docker Setup
Use Docker to ease the installs.

wtgee:
Adding initial Dockerfile and updated instructions.

We could design the container better so it is not such a monolithic service but this gets at least the initial file there. Also needs to be reduced in size.

Closes #27


jamessynge:
Are you seeking a review on this still? If so, I suspect @mimming would be your best bet.

wtgee:
> Are you seeking a review on this still? If so, I suspect @mimming would be your best bet.

I might change this still so am leaving it open. I'm not sure that starting up the jupyterlab is the best default option.

README.md
```diff
@@ -90,6 +92,59 @@ See below for more details.
...
+`gcloud` is a command line utility that lets you interact with many of the Google
+cloud services. We will primarily use this to authenticate your account but this
+is also used, for example, to upload images your PANOPTES unit takes.
+
+See the gcloud [installation page](https://cloud.google.com/sdk/docs/#install_the_latest_cloud_tools_version_cloudsdk_current_version)
+for easy install instructions.
+
+#### Let Docker use gcloud
+
+Docker needs to be able to use your `gcloud` login to pull the PANOPTES images. There
```
jamessynge: Do you think that the "advanced-authentication" is still necessary? The page now recommends just using this command:
```
gcloud auth configure-docker
```
After that (assuming I'd not previously installed docker-credential-gcr, which it doesn't appear that I did according to `gcloud components list`), running this command worked:
```
docker pull gcr.io/panoptes-survey/pocs:latest
```

README.md
```diff
@@ -90,6 +92,59 @@ See below for more details.
...
+ docker pull gcr.io/panoptes-survey/pocs:latest
+```
+
+#### Start the POCS image
+
+```
+docker run -it -p 9000:9000 --name pocs gcr.io/panoptes-survey/pocs
+```
+
+The POCS image will automatically start [Jupyter Lab](https://jupyter.org/) running
```
jamessynge: This appears to be aspirational, but not currently true. The command is just `/bin/bash`.

README.md
wtgee: This could have changed, I'll check it out.

README.md
wtgee: Actually it worked perfectly fine but I removed it. If we leave the CMD as `/bin/bash` then you can launch whatever you want when you start the instance so it's a bit more generic.

I'll fix the README, thanks.

jamessynge:
Looks like ffmpeg needs to be added to the deps used when creating the image.
And really we should probably look at breaking this up into multiple layers.

jamessynge:
And now I see that the same thing (ffmpeg being missing) happens in travis-ci.

wtgee:
@jamessynge I think this could be reasonably merged. Could feasibly remove the changes to the README and just leave the Dockerfile.  Thoughts?

jamessynge:
How about creating a docker directory, and putting the Dockerfile and relevant portions of the README into that directory. Then I think this would be good to go.

jamessynge:
I want to build on this, so rather than wait for your updates, I'm going to merge as is, then will make changes I suggested.

Issue #506, opened 2018-03-24
Startup scripts fail if no internet available


jamessynge:
The startup scripts I created run `pocs/utils/data.py` to update the IERS Bulletin A coordinate files. Specifically, we run `astroplan.download_IERS_A()`. This throws an exception if there is no internet, which the startup scripts don't tolerate, and thus POCS isn't started properly.

I think the simple fix is to remove the update of the IERS Bulletin A coordinate files from the primary startup script, and to modify the instructions for creating a crontab entry for updating these files to also include running them at boot.

PR #507, opened 2018-03-24
Make startup scripts & downloading of files more robust

jamessynge:
Fix issue #506 by making the downloading code ignore failures to
download files.


pocs/utils/data.py
```diff
@@ -1,46 +1,145 @@
...
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument(
+        '--folder',
+        help=('Destination folder for astrometry indices. Default: {}'.format(
+            _default_data_folder())),
+        default=_default_data_folder())
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        '--keep_going',
```
wtgee: We don't have a good convention on argument names but I think I prefer the dash, which gets converted to an underscore for the parameter. So `--keep-going` becomes `keep_going` as variable.

For some reason underscores in command line parameters seems off to me. Thoughts?

pocs/utils/data.py
```diff
@@ -1,46 +1,145 @@
...
+        action='store_true',
+        help='Ignore download failures and keep going to the other downloads (default)')
+    group.add_argument(
+        '--nokeep_going', action='store_true', help='Fail immediately if any download fails')
```
wtgee: Somewhat related to lack of convention mentioned above, I would separate the "no", i.e. `--no-keep-going`.

pocs/utils/data.py
```diff
@@ -1,46 +1,145 @@
...
+    group.add_argument(
+        '--wide_field', action='store_true', help='Do download wide field indices (default)')
+    group.add_argument(
+        '--nowide_field', action='store_true', help='Skip downloading wide field indices')
 
     args = parser.parse_args()
 
     if args.folder and not os.path.exists(args.folder):
-        print("{} does not exist.".format(args.folder))
+        print("Warning, data folder {} does not exist.".format(args.folder))
```
wtgee: Warn or just auto-create? We create directories in a number of different places so might just make sense here.

scripts/plot_weather.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
wtgee: Lots of boilerplate in these bash scripts. Any way we can consolidate?

scripts/update_indices.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
wtgee: See comment about boilerplate above.

wtgee: I think the filename here isn't clear. Maybe something explicit like "download_support_files.sh"?

pocs/utils/data.py
```diff
@@ -1,46 +1,145 @@
...
-            url = "http://data.astrometry.net/{}".format(fn)
+    Astrometry.net provides indices used to 'plate solve', i.e. to determine
+    which stars are in an arbitrary image of the night sky.
+    """
+
+    def __init__(self, data_folder=None, wide_field=True, narrow_field=False, keep_going=True):
+        """
+        Args:
+            data_folder: Path to directory into which to copy the astrometry.net indices.
+            wide_field: If true, downloads wide field astrometry.net indices.
```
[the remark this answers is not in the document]
jamessynge: Fixed.

pocs/utils/data.py
```diff
@@ -1,46 +1,142 @@
...
+    # --no_narrow_field is the default, so the the args list below ignores args.no_narrow_field.
+    dl = Downloader(
+        data_folder=args.folder,
+        keep_going=args.keep_going or not args.no_keep_going,
+        narrow_field=args.narrow_field,
+        wide_field=args.wide_field or not args.no_wide_field)
+    return dl.download_all_files()
+
+
+if __name__ == '__main__':
```
wtgee: I want to say that the argparse statements should be in this `__name__` and that those arguments should then be passed to a `main` but I can't find anything to support why that would be a preferred way.

scripts/download_support_files.sh
```diff
@@ -0,0 +1,52 @@
...
+echo "PANUSER: ${PANUSER}"
+echo "PANDIR: ${PANDIR}"
+echo "PANLOG: ${PANLOG}"
+echo "POCS: ${POCS}"
+echo "PAWS: ${PAWS}"
+echo "PIAA: ${PIAA}"
+
+set -x
+
+python "${POCS}/pocs/utils/data.py"
```
wtgee: This whole wrapper script seems to be to guarantee that the environment is correct for downloading but it doesn't check the `python` version. Should probably `source activate panoptes-env` if we are still recommending that route otherwise would need to make sure python3

scripts/plot_weather.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "PANUSER: ${PANUSER}"
+echo "PANDIR: ${PANDIR}"
+echo "PANLOG: ${PANLOG}"
+echo "POCS: ${POCS}"
+echo "PAWS: ${PAWS}"
+echo "PIAA: ${PIAA}"
+
+set -x
+
+"${POCS}/scripts/plot_weather.py"
```
wtgee: Same as above although we should make them consistent. This one appears to be executable where the `data.py` above is not?

pocs/utils/data.py
jamessynge: I don't see a need for that... except that I didn't pass in `sys.argv[1:]` so that it could be passed to `argparse.parse()`. However, the examples I found didn't bother with that. And it wouldn't make sense to call the method `main` if the parsing had already occurred Anyway, we can fix that later if it makes sense.

scripts/download_support_files.sh
jamessynge: The `source activate panoptes-env` should be in $HOME/.profile, and that is what the instructions call for. After all, this is a script designed to be run from cron on a PANOPTES scope, not necessarily for a dev box.

Nonetheless, I suspect you're right that I should make such a change, along with other changes to reduce the boilerplate. Since this PR has been going for a while, I'd like to handle those changes in another PR. I filed #519 to track this.

scripts/plot_weather.sh
jamessynge: Yup, the POCS/scripts are executable, but I don't think we have many in POCS/pocs that are executable.

scripts/download_support_files.sh
wtgee: Sounds good, thanks. This PR is still approved so I'll leave to you to merge when ready.

PR #510, opened 2018-04-06
Change `clobber` keyword

wtgee:
https://github.com/astropy/astropy/pull/6203

### DIFF
--- a/pocs/tests/test_config.py
+++ b/pocs/tests/test_config.py
@@ -16,12 +16,12 @@ def test_load_simulator(config):
     assert 'night' in config['simulator']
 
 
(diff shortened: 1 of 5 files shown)


wtgee:
> FWIW, it seems like most or all uses of clobber in POCS could be changed to overwrite.

Good call, I got them all now.

> Looks like we also need to change PIAA.

I'll have a PR for that shortly.

> Do we need to update requirements to select the correct (updated) version of astropy?

I went ahead and did this. I've been running 3.x for a bit now and haven't seen any issues. Can revert if think it's going to be a problem or want in separate PR.

PR #512, opened 2018-04-16
Add support for Optec FocusLynx based focus controllers

AnthonyHorton:
Includes support for the  Starlight Instruments Focus Boss II controller and Feather Touch Focuser, as used by Huntsman.

This is an initial implementation incorporating all the standard Focuser features required for current POCS operations, but not the advanced features such as automated temperature compensation, support for 2 focus motors on one controller or control over Ethernet or WiFi. It also lacks, for now, the device node scan that most of the current Camera & Focuser classes in POCS have. Instead the (possibly non-deterministic) serial port device node needs to be manually specified in the config file.

I'm putting the PR in now because there will Huntsman observing this week and I want to test this then. I'll write up Issues for the above enhancements.

PR #511, opened 2018-04-12
A few changes from last Huntsman observing run

wtgee:
* Make sure observation is still in valid list.
* Link pretty image from FITS to latest
* Weather plot fixes

PR #515, opened 2018-05-01
Updating dependencies

wtgee:
* Updating astropy dependency #510 
* Locking pycodestyle==2.3.1 until fix. Affects #511, #512, #514

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 astroplan
-astropy >= 2.0.0
+astropy >= 3.0.0
 ccdproc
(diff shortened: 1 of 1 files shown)


AnthonyHorton:
An observation, the requirement for astropy >= 3.0.0 also formalises the requirement for Python 3 because from 3.0.0 onward astropy is only available for Python 3. I know that POCS has always been Python 3 code but with this requirement attempts to install it in a Python 2 environment will fail due to unresolvable dependencies. This is a good thing! :-)

Issue #505, opened 2018-03-12
pocs_shell loop stops at end of night


jamessynge:
At dawn this morning POCS went through the end of night cleanup, but then printed this:

```
I0311 10:21:21.852 messaging.py:134     PANCHAT Weather is good and it is dark. Something must have gone wrong. Stopping loop.
```

The previous Dark Check reported False (see below), so I don't know why it would be reporting that it is dark.

Another problem is that it has apparently stopped the `run_pocs` loop (the logging stopped shortly after that), but pocs_shell didn't not prompt for another command.

```
D0311 10:12:27.985 core.py:336          Dark Check: True
D0311 10:12:27.985 core.py:351          Checking weather safety
D0311 10:12:27.986 core.py:371          Weather Safety: True [48 sec old - 2018-03-11 10:11:40.474000]
D0311 10:12:30.002 machine.py:253       Before calling track from pointing state
D0311 10:12:34.619 machine.py:267       After calling track. Now in tracking state
D0311 10:12:34.622 machine.py:309       Source: tracking	 Dest: observing
D0311 10:12:34.622 machine.py:171       Transition method: observe
D0311 10:12:34.622 machine.py:209       Checking safety for observe
D0311 10:12:34.730 core.py:336          Dark Check: True
D0311 10:12:34.730 core.py:351          Checking weather safety
D0311 10:12:34.732 core.py:371          Weather Safety: True [54 sec old - 2018-03-11 10:11:40.474000]
D0311 10:12:36.747 machine.py:253       Before calling observe from tracking state
I0311 10:12:41.651 messaging.py:134     PANCHAT I'm finding exoplanets!
D0311 10:12:41.652 observatory.py:425   Getting headers for : KIC 8462852: 120.0 s exposures in blocks of 10, minimum 60, priority 200
D0311 10:12:42.266 observatory.py:263   Exposing for camera: Cam00
D0311 10:12:42.266 camera.py:437        image_id: Panoptes Unit PAN006 @ Wheaton College_6575fc_20180311T101242
D0311 10:12:42.266 canon_gphoto2.py:124 Taking 120.0 second exposure on Cam00: /var/panoptes/images/fields/Kic8462852/6575fc/20180311T101114//20180311T101242.cr2
D0311 10:12:42.272 observatory.py:263   Exposing for camera: Cam01
D0311 10:12:42.273 camera.py:437        image_id: Panoptes Unit PAN006 @ Wheaton College_7bab97_20180311T101242
[...]

PR #516, opened 2018-05-07
Sleeping Logic

wtgee:
* Should only try to stop the states if it is actually safe.

Addresses some of the issues in #505 (there are still a few paths to shutdown).

### DIFF
--- a/pocs/state/states/default/sleeping.py
+++ b/pocs/state/states/default/sleeping.py
@@ -2,7 +2,7 @@ def on_enter(event_data):
     """ """
     pocs = event_data.model
 
-    if pocs.should_retry is False:
+    if pocs.is_safe() and pocs.should_retry is False:
         pocs.say("Weather is good and it is dark. Something must have gone wrong. " +
                  "Stopping loop.")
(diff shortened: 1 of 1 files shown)


pocs/state/states/default/sleeping.py
```diff
@@ -2,7 +2,7 @@ def on_enter(event_data):
     """ """
     pocs = event_data.model
 
-    if pocs.should_retry is False:
+    if pocs.is_safe() and pocs.should_retry is False:
```
jamessynge: What happens if safety is fluctuating rapidly? Or can that not happen?

My concern is that we might find things are unsafe at one point during a single transition but later find that things have become safe. This seems to imply that the state machine should only be able to read the safety value once per transition attempt.

wtgee: Once the system is marked unsafe it won't mark safe again for the amount of time specified in `peas.yaml` under `weather.aag_cloud.safety_delay`, which defaults to 15 minutes (on PAN001 I switched it to 30 minutes to handle fluctuating weather).

PR #530, opened 2018-07-06
Mount Calibration - WIP

wtgee:
Adding a method to calibrate the mount when it becomes lost

Todo:
- [ ] Use the `say` method
- [x] Show image in PAWS

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -92,6 +92,20 @@ class PocsShell(Cmd):
         i = DriftShell()
         i.cmdloop()
 
+    def do_calibrate_mount(self, *arg):
+        """Calibrates the mount.
+
+        This should only be run during initial setup of the unit
+        or when the mount has become "lost" and no longer knows where
+        it is pointing.
+
+        """
+        assert self.pocs and self.pocs.is_initialized(), \
+            "POCS has not been setup."
+
+        print_info("Starting mount calibration")
+        self.pocs.observatory.calibrate_mount()
+
     def do_start_messaging(self, *arg):
         """Starts the messaging system for the POCS ecosystem.
 
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -100,9 +100,6 @@ class PocsShell(Cmd):
         it is pointing.
 
         """
-        assert self.pocs and self.pocs.is_initialized(), \
-            "POCS has not been setup."
-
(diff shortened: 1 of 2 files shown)


pocs/observatory.py
```diff
@@ -537,6 +543,96 @@ def close_dome(self):
...
+            that a slew to a given AltAz will at the very least
+            point somewhere above the horizon.
+
+        Performs the following steps:
+             1. Slews to home.
+             2. Slews to target (default Alt=40° and Az=80°).
+             3. Takes a 30 second image with the primary camera.
+             4. Plates solves image and obtains coords for center of image.
+             5. Displays current separation.
+             6. Sets mount to current coords.
```
jamessynge: Is there a point here where it waits for the user to do something with the mount alignment?

bin/pocs_shell
```diff
@@ -92,6 +92,21 @@ class PocsShell(Cmd):
         i = DriftShell()
         i.cmdloop()
 
+    def do_calibrate_mount(self, *arg):
```
jamessynge: *arg seems unncessary.

bin/pocs_shell
```diff
@@ -92,6 +92,21 @@ class PocsShell(Cmd):
...
+        """Calibrates the mount.
+
+        This should only be run during initial setup of the unit
+        or when the mount has become "lost" and no longer knows where
+        it is pointing.
+
```
jamessynge: Extra blank line.

pocs/observatory.py
```diff
@@ -493,6 +499,96 @@ def close_dome(self):
             self.logger.info('Closed dome')
         return self.dome.close()
 
+    def calibrate_mount(self, target=None, exp_time=30):
+        """Calibrates the mount to a target.
```
jamessynge: What does calibration achieve? Reading the steps I don't get a clue of what the end result is.

bin/pocs_shell
wtgee: All of the `do_` commands have that. We don't use it here but I just included it for consistency. Can remove.

pocs/observatory.py
wtgee: I'll fill out the docs a little more, thanks.

wtgee:
I'm going to convert this to a utility script rather than an integrated part of the Observatory. Closing for now.

PR #535, opened 2018-07-10
Merge changes from May Huntsman commissioning run

AnthonyHorton:
A new Huntsman commissioning run (and a bunch of frenzied coding) is imminent, so I'd like to merge the assorted minor changes from the previous run. These are a few bug fixes, plus some more substantial changes to the autofocus routine and the Bisque dome control.

### DIFF
--- a/conf_files/peas.yaml
+++ b/conf_files/peas.yaml
@@ -14,7 +14,8 @@ environment:
 weather:
     station: mongo
     aag_cloud:
-        serial_port: '/dev/ttyUSB1'
+      #        serial_port: '/dev/ttyUSB1'
+        serial_port: '/dev/tty.USA19H2P1.1'
         threshold_cloudy: -25
         threshold_very_cloudy: -15.
         threshold_windy: 50.
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -122,7 +122,6 @@ class POCS(PanStateMachine, PanBase):
     def should_retry(self):
         return self._obs_run_retries >= 0
 
-
 ##################################################################################################
 # Methods
 ##################################################################################################
@@ -181,13 +180,6 @@ class POCS(PanStateMachine, PanBase):
         if self.has_messaging is False:
             self.logger.info('Unit says: {}', msg)
         self.send_message(msg, channel='PANCHAT')
-
-        if 'slack_webhook_url' in self.config:
-            try:
-                url = self.config['slack_webhook_url']
-                response = requests.post(url, json={"text": msg})
-            except Exception as e:
-                self.logger.debug("Error posting to slack: {}".format(e))
 
     def send_message(self, msg, channel='POCS'):
         """ Send a message
@@ -404,7 +396,6 @@ class POCS(PanStateMachine, PanBase):
         free_space = get_free_space()
         return free_space.value >= required_space.to(u.gigabyte).value
 
-
 ##################################################################################################
 # Convenience Methods
 ##################################################################################################
@@ -449,7 +440,6 @@ class POCS(PanStateMachine, PanBase):
         """
         while not self.is_safe(no_warning=True):
             self.sleep(delay=self._safe_delay)
-
 
 ##################################################################################################
 # Class Methods
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -5,6 +5,7 @@ import time
 import warnings
 import multiprocessing
 import zmq
+import requests
 
 from astropy import units as u
 
@@ -180,6 +181,13 @@ class POCS(PanStateMachine, PanBase):
         if self.has_messaging is False:
             self.logger.info('Unit says: {}', msg)
         self.send_message(msg, channel='PANCHAT')
+
+        if 'slack_webhook_url' in self.config:
+            try:
+                url = self.config['slack_webhook_url']
+                response = requests.post(url, json={"text": msg})
+            except Exception as e:
+                self.logger.debug("Error posting to slack: {}".format(e))
 
     def send_message(self, msg, channel='POCS'):
         """ Send a message
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -5,7 +5,6 @@ import time
 import warnings
 import multiprocessing
 import zmq
-import requests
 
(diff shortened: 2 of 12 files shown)


wtgee:
Thanks @AnthonyHorton. A lot of these changes are in #511. I'll do an actual review in a bit as this PR might be better than that one since it is fresher.

conf_files/peas.yaml
```diff
@@ -14,7 +14,8 @@ environment:
 weather:
     station: mongo
     aag_cloud:
-        serial_port: '/dev/ttyUSB1'
+      #        serial_port: '/dev/ttyUSB1'
+        serial_port: '/dev/tty.USA19H2P1.1'
```
wtgee: We should probably just comment this out and force the use of `peas_local.yaml`. I've been thinking of doing that system wide to make it obvious to new installers what needs to be changed in the config. Should be okay for here, just mostly thinking out loud.

pocs/core.py
```diff
@@ -181,6 +182,13 @@ def say(self, msg):
             self.logger.info('Unit says: {}', msg)
         self.send_message(msg, channel='PANCHAT')
 
+        if 'slack_webhook_url' in self.config:
```
wtgee: Some of the slack and twitter stuff got officially pulled into POCS, so let's strip all this from here.

pocs/focuser/focuser.py
```diff
@@ -437,22 +437,38 @@ def _autofocus(self,
...
-                best_focus = focus_positions[imax]
-            else:
-                extrema = fit(stationary_points)
-                if len(extrema) > 0:
-                    best_focus = stationary_points[extrema.argmax()]
-                    fitted = True
+            # Fit data around the maximum value to determine best focus position.
+            # Initialise models
+            shift = models.Shift(offset=-focus_positions[imax])
+            poly = models.Polynomial1D(degree=4, c0=1, c1=0, c2=-1e-2, c3=0, c4=-1e-4,
```
wtgee: Any chance you'll want to change the degree of this, i.e. pull from config?

pocs/observatory.py
```diff
@@ -485,6 +486,9 @@ def autofocus_cameras(self, camera_list=None, coarse=False):
 
         autofocus_events = dict()
 
+        if coarse is None:
+            coarse = not self._coarse_focus_done
```
wtgee: Seems like either this variable name is wrong or the default logic is wrong. Doesn't this end up doing a coarse focus every time *but* the first time? Also see setter below.

pocs/observatory.py
```diff
@@ -507,6 +511,8 @@ def autofocus_cameras(self, camera_list=None, coarse=False):
                 else:
                     autofocus_events[cam_name] = autofocus_event
 
+        self._coarse_focus_done = True
```
wtgee: Along with above, seems like this should check if a coarse focus was done and if so set variable. It is currently always set to True as far as I can tell.

pocs/utils/theskyx.py
```diff
@@ -56,6 +56,7 @@ def read(self, timeout=5):
 
             try:
                 response = self.socket.recv(4096).decode()
+                self.logger.debug('Response from reading TSX socket: {}'.format(response))
```
wtgee: This should be removed. I was using it for debug and it generated way too much output.

wtgee:
@AnthonyHorton , if you see this in time could you actually split this into separate PRs? After the removal of the items commented on above there is essentially:

* camera focus (can include small exp_time fix and warnings import)
* dome parking
* color on pretty image from FITS.

Will keep it a little cleaner. I'm guilty of the same thing on #511 but @jamessynge was good and called me out on it. ;)

pocs/core.py
AnthonyHorton: OK, removed.

pocs/focuser/focuser.py
AnthonyHorton: It's possible, but unlikely I think. As well as changing the degree of the polynomial you'd need to change the number of points around the maximum used for the fitting, so it isn't an under or over constrained fitting problem.

pocs/utils/theskyx.py
AnthonyHorton: Done.

pocs/observatory.py
AnthonyHorton: I think it's doing something fairly sensible. `self._coarse_focus_done` is set to `False` in the constructor, so unless `coarse` has been set in the call to `autofocus_cameras` then `coarse` will get set to `True` here, and so it will do a coarse focus the first time `autofocus_cameras` is called.

Not clear to me that we actually want it to do this, though. Coarse focus is only needed in a pretty limited set of circumstances already, and will only be needed less with time.

wtgee: Sadly I think I wrote this code so I would hopefully understand it. Actually, I missed the `not` in the logic. Thanks.

pocs/observatory.py
AnthonyHorton: It is all a bit odd. It sets `self._coarse_focus_done` to `True` after any call to `autofocus_cameras`, regardless of whether a coarse focus was actually performed.

I wondering whether we should just remove all of this. We hardly every want to actually do a coarse focus, and I feel like any logic for deciding whether to do it or not belongs more in the state machine than here.

wtgee: That's okay with me so I leave the decision up to you. We could feasibly build a coarse-focus state that we could transition to at various points.

pocs/core.py
```diff
@@ -5,6 +5,7 @@
 import warnings
 import multiprocessing
 import zmq
+import requests
```
wtgee: Not needed, was part of the slack stuff. Will pull through with a merge from upstream/develop anyway.

AnthonyHorton: OK, removing for now.

pocs/observatory.py
AnthonyHorton: OK. Will remove it for now and see how we go during the next Huntsman run, and reinstate if needed.

PR #538, opened 2018-07-11
Link pretty image from FITS file

wtgee:
The CR2 files are linked as part of the download process. This links those generated directly from FITS files (e.g. SBIG, FLI).

Replacing parts of #511.

### DIFF
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -90,7 +90,20 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
     if fname.endswith('.cr2'):
         return _make_pretty_from_cr2(fname, timeout=timeout, **kwargs)
     elif fname.endswith('.fits'):
-        return _make_pretty_from_fits(fname, **kwargs)
+        pretty_path = _make_pretty_from_fits(fname, **kwargs)
+
+        # Symlink latest.jpg to the image; first remove the symlink if it already exists.
+        latest_path = '{}/images/latest.jpg'.format(os.getenv('PANDIR'))
+        try:
+            os.remove(latest_path)
+        except FileNotFoundError:
+            pass
+        try:
+            os.symlink(pretty_path, latest_path)
+        except Exception:
+            warn("Can't link latest image")
+
+        return pretty_path
 
 
 def _make_pretty_from_fits(


pocs/utils/images/__init__.py
```diff
@@ -90,7 +90,20 @@ def make_pretty_image(fname, timeout=15, **kwargs):  # pragma: no cover
     if fname.endswith('.cr2'):
         return _make_pretty_from_cr2(fname, timeout=timeout, **kwargs)
     elif fname.endswith('.fits'):
-        return _make_pretty_from_fits(fname, **kwargs)
+        pretty_path = _make_pretty_from_fits(fname, **kwargs)
+
+        # Symlink latest.jpg to the image; first remove the symlink if it already exists.
+        latest_path = '{}/images/latest.jpg'.format(os.getenv('PANDIR'))
```
jamessynge: What do you think about moving the linking for CR2 files from the cr2_to_jpg.sh script into this file?
Then we don't have two different approaches. And we don't have the primary==False check in just the CR2 case, and not the FITS case?

wtgee: I would almost prefer to go the other way and bury the details of the linking inside `_make_pretty_from_fits`, which is almost what I did in the first place. 

For the CR2 we use the `convert` shell command. We could split this out of the existing bash script but then would just be calling another bash script. Not a huge deal either way.

jamessynge:
I'm not sure I understand the problem with overwriting that you're
mentioning. Could you clarify?

wtgee:
> I'm not sure I understand the problem with overwriting that you're mentioning. Could you clarify?

The JPG is extracted directly from the CR2 (it's an embedded thumbnail). Originally we were keeping that copy alongside the FITS file but also resizing the JPG and adding the caption to the bottom. So for the CR2 it was not actually doing a symlink to the original file but instead generating a new JPG. The timelapse movies have been generated from the original JPGs, which means they are not resized and don't have a caption. 

So with the above change we no longer have two JPGs but instead resize and caption the original and then do a symlink. As such the timelapse will start to appear with a caption (which will have an unreadable time since it's 3 fps) and will be slightly different size. Not a big deal either way and probably for the better to have only one JPG anyway.

wtgee:
Note also that on PAN008 here in Sydney running a Raspberry Pi B 3 (not the +) it was taking too long for me to run the `convert` process (inside `$POCS/scripts/cr2_to_jpg.sh`) that resizes and captions. As such I am skipping that entirely and just writing the one jpeg alongside the fits and linking it to the latest in the images folder. 

I do miss the time stamp as it's handy to know the image change. Luckily (:confused: ?) my tracking isn't very good right now so it's easy to tell when the image updated.

jamessynge:
I've updated to this version of software. I find that the pretty image is no longer being produced.

PR #541, opened 2018-07-24
Evening flats - WIP

wtgee:
Incorporating the evening flats developed for Huntsman. This includes some clean-up of the methods. Have tested on PAN008 and will test more on PAN001 before ready.

docstring:

        This method will slew the mount to the given AltAz coordinates (which
        should be roughly opposite of the setting sun) and then begin the flat-field
        procedure. The first image starts with a simple 1 second exposure and
        after each image is taken the average counts are anazlyzed and the exposure
        time is adjusted to try to keep the counts close to `target_adu_percentage`
        of the `(max_counts + min_counts) - bias`.

        The next exposure time is calculated as:

        ```
            exp_time = int(previous_exp_time * (target_adu / counts) *
                       (2.0 ** (elapsed_time / 180.0)) + 0.5)
        ```

        Under- and over-exposed images are rejected. If image is saturated with
        a short exposure the method will wait 60 seconds before beginning next
        exposure.

        Optionally, the method can also take dark exposures of equal exposure
        time to each flat-field image.


joshwalawender:
What are we trying to get from these flats?  With PANOPTES's 10x15 degree field of view, these will definitely be plagued by gradients.  If all we need this for is to eliminate small scale flat field effects, then this data will work, but I would not trust it to correct any flat field effects larger than maybe a few hundred pixels.  The interaction of these gradients with the data analysis algorithm will have to be considered carefully.

wtgee:
> What are we trying to get from these flats? With PANOPTES's 10x15 degree field of view, these will definitely be plagued by gradients. If all we need this for is to eliminate small scale flat field effects, then this data will work, but I would not trust it to correct any flat field effects larger than maybe a few hundred pixels. The interaction of these gradients with the data analysis algorithm will have to be considered carefully.

Yeah, we will most definitely have a huge gradient. The only flats I have now are cloud flats and aren't worth analyzing but after I get some from PAN001 I can look at that.

I think for now the goal is to simply get the existing code in there so that we could be taking flats in order to analyze them. If we are finding in the future that it's not needed/useless we can change or remove, but for now I'd like to at least start getting some of the data.

wtgee:
See also #543 for taking bias frames.

conf_files/pocs.yaml
```diff
@@ -7,6 +7,7 @@ location:
     elevation: 3400.0 # Meters
     utc_offset: -10.00 # Hours
     horizon: 30 # Degrees
+    civil_horizon: 0 # Degrees
```
jamessynge: Could you add a comment explaining the purpose of the different horizon values?

conf_files/pocs.yaml
```diff
@@ -64,3 +65,8 @@ messaging:
 #         output_timestamp: False
 
 state_machine: simple_state_table
+# Default coordinates to take evening flats
+flat_field:
+    twilight:
+        alt: 70
+        az: 120
```
jamessynge: How are these selected? Seems like we should compute the az from the sun.

pocs/core.py
```diff
@@ -323,7 +323,7 @@ def is_dark(self):
 
         """
         # See if dark
-        is_dark = self.observatory.is_dark
+        is_dark = self.observatory.is_dark()
```
jamessynge: It would be nice if we had a naming convention allowing one to distinguish fields from methods.

pocs/core.py
```diff
@@ -431,13 +431,14 @@ def sleep(self, delay=2.5, with_status=True):
         if delay > 0.0:
             time.sleep(delay)
 
-    def wait_until_safe(self):
-        """ Waits until weather is safe.
+    def wait_until_dark(self, horizon='astro'):
+        """ Waits until it is dark.
```
jamessynge: This sounds like a large change in semantics. Which of "is_safe()" or "is_dark()" take weather into account?

pocs/observatory.py
```diff
@@ -70,16 +77,25 @@ def __init__(self, *args, **kwargs):
...
+    def is_dark(self, horizon='astro'):
+
+        try:
+            horizon_deg = self.config['location']['{}_horizon'.format(horizon)]
+        except KeyError:
+            # Look for legacy name. Note this will happen for either horizon
+            try:
+                horizon_deg = self.config['location']['twilight_horizon']
+            except KeyError:
+                self.logger.info("Can't find twilight_horizon, using -18°")
```
jamessynge: This isn't in sync with the yaml file, which has no astro_horizon. Did you mean to add one?

pocs/observatory.py
```diff
@@ -70,16 +77,25 @@ def __init__(self, *args, **kwargs):
 # Properties
 ##########################################################################
 
-    @property
-    def is_dark(self):
-        horizon = self.location.get('twilight_horizon', -18 * u.degree)
+    def is_dark(self, horizon='astro'):
```
jamessynge: Maybe horizon should be called horizon_name?

pocs/observatory.py
```diff
@@ -70,16 +77,25 @@ def __init__(self, *args, **kwargs):
...
+
+        try:
+            horizon_deg = self.config['location']['{}_horizon'.format(horizon)]
```
jamessynge: And horizon_deg could be called horizon_angle?

pocs/observatory.py
```diff
@@ -70,16 +77,25 @@ def __init__(self, *args, **kwargs):
...
+                horizon_deg = -18 * u.degree
 
         t0 = current_time()
-        is_dark = self.observer.is_night(t0, horizon=horizon)
+        is_dark = self.observer.is_night(t0, horizon=horizon_deg)
 
         if not is_dark:
             sun_pos = self.observer.altaz(t0, target=get_sun(t0)).alt
-            self.logger.debug("Sun {:.02f} > {}".format(sun_pos, horizon))
+            self.logger.debug("Sun {:.02f} > {} [{}]".format(
```
jamessynge: Perhaps swap the order of horizon and horizon_deg; e.g.:

Sun -10 > -18 ("astro" horizon)

pocs/observatory.py
```diff
@@ -378,8 +394,9 @@ def update_tracking(self):
                 # Adjust tracking for up to 30 seconds then fail if not done.
                 start_tracking_time = current_time()
                 while self.mount.is_tracking is False:
-                    if (current_time() - start_tracking_time).sec > 30:
-                        raise Exception("Trying to adjust Dec tracking for more than 30 seconds")
+                    if (current_time() - start_tracking_time).sec * 1000 > max_time:
```
jamessynge: max_time should have a unit in its name (e.g. _ms), or better yet should be a datetime.timedelta object.

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+                           max_num_exposures=10,
+                           take_darks=False,
+                           *args, **kwargs
+                           ):
+        """Take flat fields.
+
+        This method will slew the mount to the given AltAz coordinates (which
+        should be roughly opposite of the setting sun) and then begin the flat-field
+        procedure. The first image starts with a simple 1 second exposure and
+        after each image is taken the average counts are anazlyzed and the exposure
```
jamessynge: c/anazlyzed/analyzed/

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+
+        Args:
+            alt (float, optional): Altitude for flats
+            az (float, optional): Azimuth for flats
+            min_counts (int, optional): Minimum ADU count
+            max_counts (int, optional): Maximum ADU count
+            bias (int, optional): Default bias for the cameras
+            max_exptime (float, optional): Maximum exposure time before stopping
+            camera_list (list, optional): List of cameras to use for flat-fielding
+            target_adu_percentage (float, optional): Exposure time will be adjust so
```
jamessynge: c/adjust/adjusted/

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+                that counts are close to: target * (`min_counts` + `max_counts`). Default
```
jamessynge: c/Default/Defaults/

And below.

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+                           az=None,
+                           min_counts=1000,
+                           max_counts=12000,
+                           bias=2048,
+                           max_exptime=60.,
+                           camera_list=None,
+                           target_adu_percentage=0.5,
+                           max_num_exposures=10,
+                           take_darks=False,
+                           *args, **kwargs
```
jamessynge: *args and **kwargs don't appear to be used.

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+            for cam_name in camera_list:
+
+                camera = self.cameras[cam_name]
+                exp_time = exp_times[cam_name][-1].value
+
+                filename = "{}/flats/{}/{}/{}.{}".format(
+                    image_dir,
+                    camera.uid,
+                    flat_obs.seq_time,
+                    'flat_{:02d}'.format(flat_obs.current_exp),
```
jamessynge: Why not just put that into the filename format string?

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+
+            # Block until done exposing on all cameras
+            while not all([info['event'].is_set() for info in camera_events.values()]):
+                self.logger.debug('Waiting for flat-field image')
+                time.sleep(1)
+
+            # Check the counts for each image
+            is_saturated = False
+            for cam_name, info in camera_events.items():
+                img_file = info['filename'].replace('.cr2', '.fits')
```
jamessynge: That appears to be a canon specific path. I thought this came from Huntsman?

pocs/observatory.py
```diff
@@ -537,6 +555,235 @@ def close_dome(self):
...
+        flat_obs = self._create_flat_field_observation(alt=alt, az=az)
+
+        if camera_list is None:
+            camera_list = list(self.cameras.keys())
+
+        # Start with 1 second exposures for each camera
+        exp_times = {cam_name: [1. * u.second] for cam_name in camera_list}
+
+        # Loop until conditions are met for flat-fielding
+        while True:
```
jamessynge: Nothing to do yet, but I'm worried about this being a potentially very long running function. Seems like something worthy of an async work flow system (ala a state machine).

pocs/observatory.py
```diff
@@ -572,8 +819,8 @@ def _setup_location(self):
             pressure = config_site.get('pressure', 0.680) * u.bar
             elevation = config_site.get('elevation', 0 * u.meter)
             horizon = config_site.get('horizon', 30 * u.degree)
-            twilight_horizon = config_site.get(
-                'twilight_horizon', -18 * u.degree)
+            observe_horizon = config_site.get(
+                'observe_horizon', -18 * u.degree)
```
jamessynge: Is observe_horizon expected to be in the yaml file? I don't see it.

conf_files/pocs.yaml
wtgee: Yes, probably should. That change might need to come when we add the morning flats so are having to compute it anyway.

pocs/core.py
wtgee: Methods should have begin with a verb, fields should be nouns. Our predicates are a mish-mash right now between properties and methods. I think it makes more sense to make this a property instead of a method but was more consistent the other way. Can change in future (or now I suppose).

pocs/core.py
wtgee: It is a change in semantics to make it only a wait loop for the different horizon levels. Therefore it is not checking the weather safety at all, which is handled by other methods. The change in name was an attempt to make that more clear.

pocs/observatory.py
wtgee: Yes, meant to switch, thanks.

pocs/observatory.py
wtgee: Too much copying and pasting, not enough thinking. :)

pocs/observatory.py
wtgee: I adapted it first and tested it in PAN008 before submitting PR

pocs/observatory.py
wtgee: Correct, it currently blocks inside this loop for a while. There are a number of checks to get out of the loop although notably none for weather. Should be since at least at MLO we can have rapidly changing conditions right as sun goes down.

pocs/observatory.py
wtgee: I've updated config to add these.

wtgee:
Quick update: I had a chance to test this on some cloud flats last night and I think I got some of the kinks worked out. I need to do a little work to clear up the morning logic (and maybe generalize it to just take morning flats).

Hopefully some more testing over the weekend and then we can get it integrated.

wtgee:
FYI, will be testing this on PAN001 and PAN008 tonight and if it works well it will be ready to open up tomorrow.

wtgee:
Closing this as bloated. Overview issue moved to #732

Issue #546, opened 2018-08-13
Add support for distributed Cameras


AnthonyHorton:
Experience with the Huntsman Telescope has shown that running arrays of USB cameras from a single computer is unreliable. Consequently the Huntsman team have decided to follow the example of the Dragonfly Telescope and change to a distributed control system, with each camera being controlled by its own computer and a central control computer performing the observatory level tasks.

This requires additional features for POCS to support distributed operation. Fortunately the modular, object orientated structure of POCS provides an natural interface for the split, that between the `Observatory` object and the `Camera` object. What is required is a network layer between these objects.

AnthonyHorton:
There is, hopefully, significant overlap between the long term interests of PANOPTES and the short term needs of Huntsman, but arguably these are separate issues.

It sounds like what you're thinking of is a major rewrite of PANOPTES to turn it into robust, adaptive, and deeply distributed system, something I think we would all agree should be on the roadmap. On the other hand what I had in mind when opening this Issue was the absolute _minimum_ change required to enable cameras to be controlled by their own Raspberry Pi's (in our case Pi 3 B+'s). We should certainly try to make the latter a step along the path towards the former to avoid later duplication of effort, but that may not be practical.

It was a strong desire to limit the scope of the changes that lead me to start with Pyro in #547. It may not be the best solution long term, but it did offer the prospect of a workable solution with minimal changes to existing code. It makes it possible to address the Issue by adding additional optional functionality, rather than altering the existing architecture.

I see parallels here with the SBIG and FLI CCD cameras, and Birger Engineering and Optec focus controllers, which were added to POCS because Huntsman (and hopefully other projects, in the future) needed them.

jamessynge:
I'm happy to hear that you're assuming for the minimum change. The PR's
list of tasks done and in progress made it seem rather more significant
than you describe here. I'll take a look at it.

wtgee:
Agree with @AnthonyHorton, I don't think this introduces any major complications and puts some of the camera line in work with the long term road-map of decoupling the hardware from the `Observatory`, which was agreed on a long time ago as any first step toward moving to a more decentralized system.

@jamessynge I sent you the info on Pyro4 a few weeks back and mentioned that was the most likely option that was being looked at. It shouldn't be incompatible with indilib/indigo. And, I suppose I should point out, this *is* making an individual camera work with a Pi, and with far less effort than an immediate switch to something like indilib, which would not at all be practical right now.

[...]

PR #547, opened 2018-08-13
Add support for distributed Cameras - WIP

AnthonyHorton:
Addressing Issue #546 by adding scripts and wrappers to enable distributed control of camera arrays using the Pyro4 library.

Still very much a work in progress, just opening the PR now to solicit feedback and to get Travis CI and Codecov to keep an eye on me.

- [X] Pyro name server startup
- [X] Pyro camera server startup
- [X] Automatic discovery of remote cameras
- [X] Initialise remote cameras, get UIDs
- [X] Initiate exposures on remote cameras
- [X] Transfer files from remote cameras after exposures
- [x] Autofocus of remote cameras
- [x] `take_observation` using remote cameras
- [x] Access remote camera properties
- [x] Set remote camera properties 

Edit: Additional todo's to make this more usable

- [ ] Have `pocs.camera.create_cameras_from_config` handle camera server (re)start on the distributed controllers
- [ ] Have camera servers periodically check the name server and re-register is needed, in order to handle the case where the name server has been restarted while they are running
- [ ] Have `pocs.camera.create_cameras_from_config` handle name server (re)start. This is only helpful once the camera servers automatically re-register themselves

Between them these changes should reduce the complexity of POCS startup to be the same as it is without any distributed cameras, and also allow restarting the name server, POCS, or any camera server without having to restart everything else.


jamessynge:
Before proceeding with this, can we have an architectural review of this very large set of changes?

AnthonyHorton:
@jamessynge I would welcome that. Two points though:

1. Huntsman needs this _urgently_. Thus I'm inclined to proceed in the meantime anyway, in order to get something working well enough that Huntsman can start using it. If review comes up with a better approach and none of this stuff ever gets merged that's fine, but we need something now.
1. The changes aren't as large as they might look, or at least not as widespread. I have very deliberately ensured that I have not touched any of the existing code in either the `Observatory` or `Camera` classes. My approach has been to create a new `Camera` class to represent the client side interface to the distributed cameras, and create a wrapper for the existing `Camera` classes on the server side. No change to any of the existing interfaces, and the new code is limited to these two new classes (both in `pocs/camera/pyro.py`), plus two new scripts, one new config file, one new utility function, and some tests.

conf_files/pyro_camera.yaml
```diff
@@ -0,0 +1,99 @@
+# Config file for a distributed camera & focuser.
+#
+# Contains a default configuration with a simulated camera and focuser for testing
+# and example configurations for other camera and focusers (commented out).
+#
+# Do not edit this file, instead copy one of the examples to pyto_camera_local.yaml
```
jamessynge: Is this supposed to be pyro_camera_local (not pyto_...)?

conf_files/pyro_camera.yaml
```diff
@@ -0,0 +1,99 @@
...
+# before editing as required.
+#
+# Simulated camera and focusers:
+name: camera.simulator.001
+host: 127.0.0.1
```
jamessynge: localhost might be better, as this works with IPv6 as well as IPv4... assuming the underlying library can handle this.

conf_files/pyro_camera.yaml
```diff
@@ -0,0 +1,99 @@
...
+    autofocus_range:
+      - 50
+      - 250
+    autofocus_step:
+      - 5
+      - 25
+    autofocus_seconds: 1
+    autofocus_size: 500
+directories:
+    base: /var/panoptes
```
jamessynge: If feasible (i.e. easy-ish), it might be nice to allow for envvars here.

conf_files/pyro_camera.yaml
```diff
@@ -0,0 +1,99 @@
...
+    images: temp
+    data: data
+#
+# SBIG camera with Birger focuser:
```
jamessynge: Do these commented out entries serve a purpose? E.g. do they document something not shown above?

pocs/camera/camera.py
```diff
@@ -68,7 +68,7 @@ def __init__(self,
                     self.logger.critical("Couldn't import Focuser module {}!".format(module))
                     raise err
                 else:
-                    self.focuser = module.Focuser(**focuser, camera=self)
+                    self.focuser = module.Focuser(**focuser, camera=self, config=self.config)
```
jamessynge: Note that this construct can lead to run time errors due to adding conflicting entries into the focuser dict: if it contains camera or config, then an exception will be raised. Not saying that you shouldn't use it, but it can be useful to make a copy of the dict and then insert the overrides (e.g. camera= and config=), allowing you to warn of errors instead of dying or to simply provide a better error message.

pocs/camera/camera.py
```diff
@@ -319,8 +318,6 @@ def autofocus(self,
                 keyword arguments for the merit function.
             mask_dilations (int, optional): Number of iterations of dilation to perform on the
                 saturated pixel mask (determine size of masked regions), default 10
-            spline_smoothing (float, optional): smoothing parameter for the spline fitting to
```
jamessynge: FWIW, these several spline changes seem unrelated to adding the distributed camera support.

pocs/camera/fli.py
```diff
@@ -215,7 +216,8 @@ def take_exposure(self,
             seconds = seconds * u.second
 
         if not self._exposure_lock.acquire(blocking=False):
-            message = 'Attempt to start exposure on {} ({}) while exposure in progress! Waiting...'.format(self.name, self.uid)
+            message = 'Attempt to start exposure on {} ({})'.format(self.name, self.uid) + \
```
jamessynge: Are the remaining changes in this file just pep8 formatting?

pocs/camera/pyro.py
```diff
@@ -0,0 +1,435 @@
...
+from pocs.camera import AbstractCamera
+
+
+# Enable local display of remote tracebacks
+sys.excepthook = Pyro4.util.excepthook
+
+
+class Camera(AbstractCamera):
+    """
+    Class representing the client side interface to a distributed array of cameras
```
jamessynge: Not a single camera in the array? I'm curious why that approach vs. exposing a set of remote cameras.

pocs/focuser/birger.py
```diff
@@ -403,33 +403,43 @@ def _initialise(self):
...
 
     def _get_lens_info(self):
         response = self._send_command('id', response_length=1)
         self._lens_info = response[0].rstrip()
-        self.logger.debug("Got lens info '{}' for {} on {}".format(self.lens_info, self.name, self.port))
+        self.logger.debug("Got lens info '{}' for {} on {}".format(self._lens_info,
+                                                                   self.name,
+                                                                   self.port))
 
     def _initialise_aperture(self):
```
jamessynge: Since this is only used by Huntsman it doesn't matter, but I'd suggest American English spelling of words in the API shared with PANOPTES.

pocs/utils/pyro/__init__.py
```diff
@@ -0,0 +1,65 @@
+import netifaces
+
+
+def get_own_ip(verbose=False, logger=None):
```
jamessynge: Is this code being run on the remote camera?

pocs/utils/pyro/pyro_camera_server.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
jamessynge: Please add file level comments describing the purpose.
And should this be in POCS/scripts? That is a true question, not a statement at all.

pocs/utils/pyro/pyro_name_server.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
```
jamessynge: Ditto about comments.

conf_files/pyro_camera.yaml
AnthonyHorton: Yep!

conf_files/pyro_camera.yaml
AnthonyHorton: localhost should work. I'll try it.

conf_files/pyro_camera.yaml
AnthonyHorton: Yes, they're templates/examples for real hardware configurations. They don't differ a lot from the uncommented localhost + simulators config but they do illustrate typical values.

pocs/camera/camera.py
AnthonyHorton: Yes, leftover arguments and docstring lines that should have been removed back when the spline fitting was removed. Don't really belong in this PR but couldn't resist pruning them.

pocs/camera/fli.py
AnthonyHorton: Yes.

pocs/utils/pyro/__init__.py
AnthonyHorton: It is optionally used by both the camera servers (on the remote camera controllers) and the name server (can be anywhere, but probably on the main control computer), more frequently by the latter.

I don't anticipate it being used much at all in operational systems though, where it's likely that everything will have a static IP which can be put into config files, but it is handy during testing.

pocs/camera/pyro.py
AnthonyHorton: It made sense at the time? I have, to begin with, been trying to avoid changes to `Observatory`, as part of my general approach of making these changes as self contained as possible and leaving all existing interfaces unchanged, & a single camera array object facilitated that.

Treating an array of cameras as if it were one camera is becoming increasingly awkward as I work my way up to the higher level functions, though, so it may be time to rethink that. Would need to modify `Observatory._create_cameras` and add the functions for finding & (re)connecting to remote cameras but it wouldn't be a major change, and would make implementing `take_observation` somewhat tidier.

pocs/utils/pyro/pyro_camera_server.py
AnthonyHorton: Agree on the comments, unsure on file location. It is a script, but also an importable function, and could be used either way. It's in utils right now mainly because I'm trying to get coverage to include it (scripts aren't included, but could be with a change to coveragerc).

pocs/camera/pyro.py
AnthonyHorton: Thanks for raising this point. I've done a rewrite with single cameras and, at the cost of a small amount of additional code in `Observatory`, it's made things a _lot_ simpler. I'd previously been tying myself in knots trying to find a neat way to adapt `take_observation` to work with an array of cameras but with single cameras I didn't need to re-implement it at all, the base class version is fine.

pocs/utils/pyro/pyro_camera_server.py
AnthonyHorton: Comment added.

pocs/utils/pyro/pyro_name_server.py
AnthonyHorton: Comment added.

pocs/camera/camera.py
AnthonyHorton: OK, I see your point. The `focuser` `dict` _shouldn't_ ever contain `camera` or `config` items, but it would break things if it somehow did. Will combine into a `kwargs` dict before passing to the `Focuser` constructor.

AnthonyHorton:
Getting close... Have moved testing of the distributed cameras into the main `pocs/tests/tests_camera.py` file so that the distributed cameras are put through the same set of tests as the `simulator`, `sbig`, & `fli` cameras. Discovered a few bugs in the process, & fixed them. The important methods `take_exposure`, `autofocus` and `take_observation` now all seem to be working, and property access is in place. The only tests that are failing now are ones that test autofocus failure modes by poking around inside the `Camera`'s `focuser` attribute to break things, which doesn't work with the distributed cameras because I haven't directly exposed the focuser in the distributed camera interface.

For the most part I don't actually want to expose the focuser, the intent has always been that it should be controlled via the config files and the camera's `autofocus` method only, though providing a small amount of low level functionality may make sense (e.g. getting and setting the current focus position). Will have to think about what to include. Most if not all of the currently failing tests should probably be skipped, though. I don't want to add a bunch of extra interfaces just to enable the tests to break things, especially as these tests would just be re-testing code that's already being exercised during testing of the non-distributed camera simulator.

Once I've sorted that out I want to do a quick review of the tests to make sure they really are making sure that everything is working as intended, then do a bit of testing with real hardware.

AnthonyHorton:
Right, I think this is essentially feature complete, it's passing tests, and those tests have good coverage. It's ready for review.

A few known issues, that I'd appreciate any help with:

1. The test coverage report is misleading. It is not counting code that is executed via the scripts as being covered. Something wrong with the coverage configuration, I think. Need to work out what the problem is and fix it.
1. I'm doing something wrong with the logging on the camera server side. The camera server script creates a `CameraServer` object, which creates a logger, and a `Camera`, which in turn creates a `Focuser`. I've got the logger logging to `$PANLOG`/pyro-camera-server-*.log` and am attempting to pass that logger from the `CameraServer` to the `Camera` to the `Focuser`but instead of all three sharing that log the logger gets restarted twice and only logs stuff from the `Focuser`.

jamessynge:
Could you look into how to add coverage of the changes on observatory.py and focuser.py? The report indicates no coverage of new lines.

pocs/tests/test_camera.py
```diff
@@ -11,14 +12,18 @@
...
+import signal
 from ctypes.util import find_library
 
 import astropy.units as u
 import astropy.io.fits as fits
+import Pyro4
 
-params = [SimCamera, SBIGCamera, FLICamera]
-ids = ['simulator', 'sbig', 'fli']
+params = [SimCamera, PyroCamera, SBIGCamera, FLICamera]
```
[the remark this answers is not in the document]
wtgee: I'll echo this. We need to think about how to test against specific hardware types a little more carefully.

wtgee:
I'm not sure what is going on with our coverage but it's often misleading, not just on this PR, especially for the patches. I'll try to spend some time with it.

pocs/camera/fli.py
```diff
@@ -32,6 +32,7 @@ def __init__(self,
                  name='FLI Camera',
                  set_point=25 * u.Celsius,
                  filter_type='M',
+                 library_path=False,
```
wtgee: This FLI camera doesn't have a docstring so things like the `library_path` aren't documented anywhere.

pocs/camera/pyro.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+class Camera(AbstractCamera):
+    """
+    Class representing the client side interface to a distributed camera
+    """
+    def __init__(self,
+                 uri,
+                 name='Pyro Camera',
+                 model='pyro',
```
wtgee: Seems like we will also need to be able to specify the actual model behind the pyro proxy.

pocs/camera/pyro.py
```diff
@@ -0,0 +1,475 @@
...
+        dir_name = dir_name.rstrip('/') + '/'
+
+        # Make sure proxy is in async mode
+        Pyro4.async(self._proxy, async=True)
+
+        # Start the exposure
+        self.logger.debug('Taking {} second exposure on {}: {}'.format(
+            seconds, self.name, base_name))
+        # Remote method call to start the exposure
+        exposure_result = self._proxy.take_exposure(seconds=seconds,
```
wtgee: I get a bit worried about how much this (and the `autofocus`) method repeat what is in the main camera class. It seems like the only real change is to have the `._proxy` (and the setting of the async) above. Is that a mostly correct assessment?  Seems like it should be a bit easier to proxy some of these methods without duplicating code.

pocs/focuser/focuser.py
```diff
@@ -398,6 +360,22 @@ def _autofocus(self,
                                               self._camera.uid,
                                               start_time)
 
+        if take_dark:
+            file_path = "{}/{}.{}".format(file_path_root, "dark", self._camera.file_extension)
```
wtgee: Does this need a unique (timestamped?) name?

pocs/focuser/focuser.py
```diff
@@ -486,14 +464,17 @@ def _autofocus(self,
             fitted = True
 
             # Guard against fitting failures, force best focus to stay within sweep range
-            if best_focus < focus_positions[0]:
-                self.logger.warning("Fitting failure: best focus {} below sweep limit {}".format(best_focus,
-                                                                                                 focus_positions[0]))
+            min_focus, max_focus = focus_positions[0], focus_positions[-1]
+            if best_focus < min_focus:
+                m = "Fitting failure: best focus {} below sweep limit {}".format(best_focus,
```
wtgee: Should this be `msg` instead of `m`?

pocs/focuser/focuser.py
```diff
@@ -486,14 +464,17 @@ def _autofocus(self,
...
+                                                                                 min_focus)
+                self.logger.warning(msg)
                 best_focus = focus_positions[1]
 
-            if best_focus > focus_positions[-1]:
-                self.logger.warning("Fitting failure: best focus {} above sweep limit {}".format(best_focus,
-                                                                                                 focus_positions[-1]))
+            if best_focus > max_focus:
+                m = "Fitting failure: best focus {} above sweep limit {}".format(best_focus,
```
wtgee: Ditto

pocs/utils/pyro/pyro_camera_server.py
wtgee: I do the same thing for `pocs/utils/data.py` but want to move away from it. I think we just leave everything within `pocs/` as importable modules and just create a thin-wrapper in `scripts`.

conf_files/pyro_camera.yaml
AnthonyHorton: OK, localhost does work. Now using it in the test configuration. Of course localhost (or its IPv4 or IPv6 equivalents) is only relevant for tests, passing it to either the name server or the camera server prevents any connections between machines.

conf_files/pyro_camera.yaml
AnthonyHorton: More documentation makes sense, the question is where to put it. The `name` and `host` entries are the only two here that are specific the distributed camera configuration. The `camera` and `directories` sections, and the `focuser` sub-section, are exactly the same as for a regular `pocs_local.yaml` config. Currently the only documentation for these are in the docstrings of the corresponding classes.

pocs/utils/pyro/pyro_camera_server.py
AnthonyHorton: OK, that makes sense. I'll keep the functions that do all the work in `$POCS/pocs/utils/pyro` and have minimal scripts in `$POCS/scripts` that just parse the command line argument and call the function.

pocs/observatory.py
```diff
@@ -727,9 +732,39 @@ def _create_cameras(self, **kwargs):
 
                 self.cameras[cam_name] = cam
 
+        distributed_cameras = kwargs.get('distributed_cameras',
+                                         camera_info.get('distributed_cameras', False))
+        if not a_simulator and distributed_cameras:
```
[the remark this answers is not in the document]
AnthonyHorton: Done.

pocs/tests/test_camera.py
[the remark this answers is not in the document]
AnthonyHorton: `test_camera.py` will attempt to test a real SBIG and/or FLI camera (and associated focusers) if there is a `pocs_local.yaml` present that has camera configs for them. I used this feature extensively during development of the SBIG and FLI camera classes, but don't routinely run the test suite on real hardware. The implementation here is undeniably a bit of a hack, and I certainly agree we should develop a more systematic approach to testing with real hardware.

pocs/tests/test_camera.py
```diff
@@ -27,18 +32,50 @@ def images_dir(tmpdir_factory):
     return str(directory)
 
 
+def end_process(proc):
+    proc.send_signal(signal.SIGINT)
+    return_code = proc.wait()
+
+
+@pytest.fixture(scope='module')
+def name_server(request):
```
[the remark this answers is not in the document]
AnthonyHorton: Makes sense for these 4 tests, especially as the code they're testing is mostly in `pocs/utils/pyro`

pocs/camera/fli.py
AnthonyHorton: Huh. None of the camera classes have docstrings. How did I get away with that up to now?

pocs/camera/pyro.py
AnthonyHorton: The actual model behind the Pyro proxy is obtained from the proxy in `connect()`, and used to update the `model` property then. Consequently the value passed to `__init__()` is almost immediately discarded unless the attempt to connect to the remote camera fails.

pocs/camera/pyro.py
AnthonyHorton: Most methods are easy to proxy with minimal code, but `take_exposure` and `autofocus` do require a bit more.

The main reason for this is because I'm doing the transfer of files from the remote computer, etc.,  outside of Pyro. This is what the Pyro docs recommend, it's not intended for moving big binary blobs around. Most of the code in the `take_exposure` and `autofocus` methods of both the `Camera` and `CameraServer` classes is related to setting up these file transfers and cleaning up afterwards.

Most of the rest of the code in this file is converting the types of arguments/return values before/after passing them through Pyro. Pyro's default serialiser (serpent) handles Python builtin types automatically so for the sake of simplicity I make sure I'm only passing builtin types around.

pocs/focuser/focuser.py
AnthonyHorton: Maybe. It does have a unique path, though, which includes both the camera UID and a timestamp. Have been doing it this way since the autofocus functionality was added.

pocs/focuser/focuser.py
AnthonyHorton: Probably, but then I'd have to split the string up in order to keep the line lengths below 100 characters.

pocs/utils/pyro/__init__.py
```diff
@@ -63,3 +70,52 @@ def get_own_ip(verbose=False, logger=None):
     if logger:
         logger.debug(msg)
     return host
+
+
+def run_name_server(host=None, port=None, autoclean=0):
```
jamessynge: Doc strings please.

pocs/observatory.py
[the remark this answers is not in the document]
jamessynge: Thanks.

scripts/pyro_camera_server.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+"""
+Script to run a Pyro camera server. This should be run on the control computer for a distributed
```
jamessynge: This runs on the main computer, or on one of the camera computers? The name makes it sound like it is on each camera, but the comment makes it sound like the main computer.

scripts/pyro_camera_server.py
```diff
@@ -0,0 +1,18 @@
...
+camera. The configuration for the camera is read from the pyro_camera.yaml/pyro_camera_local.yaml
+config file. The camera servers should be started after the name server, but before POCS.
```
jamessynge: This strikes me as problematic. It would seem to be best to have a rendezvous that can allow for a uncoordinated starting, and especially restarting (e.g. one camera is wonky, so you power cycle it without requiring POCS to be shutdown).

scripts/pyro_camera_server.py
AnthonyHorton: The camera computers. That's what I meant by 'control computer for a distributed camera'.

AnthonyHorton: Reworded slightly for avoidance of doubt.

AnthonyHorton:
OK, I've made a few changes in response to comments and, probably more interestingly, I've made some changes to `travis.yml` and `.coveragerc` that appear to have been successful in getting coverage to properly measure coverage from subprocesses. This has made the coverage report much more accurate for the new code and, presumably, for the rest of it too.

While working on that I also discovered a very perplexing issue with the test suite. `test_locate_name_server` hangs indefinitely _if and only if_ it is run almost at the end of the ~25 minute 481 test full suite test run, which is exactly when it was run due to it residing in `tests/utils/test_pryo_utils.py`.  Running `tests/utils/test_pyro_utils.py` by itself, or together with the other two tests files directly relevant to this PR (`tests/test_camera.py` and `tests/test_observatory.py`) results in all tests passing. In desperation I tried renaming `tests/utils/test_pyro_utils.py` to `tests/test_aardvark.py` so that it would be run at the start of the POCS tests and that 'fixed' the problem! I have a strong suspicion that neither of you will be happy with renaming files after nocturnal African insectivores as a bug fixing technique but I'm at a loss as to what the root cause is. Nothing at all appears in the logs when `test_locate_name_server` hangs. _Something_ must be causing the state to change in the ~20 minute interval between `test_observatory.py` and `utils/test_pyro_utils.py` but I don't have any leads on what it might be.

wtgee:
I think Aardvark Testing™ is the way of the future! What better way to kill those bugs!

I'll try to run some tests today and see if I can find out anything.

wtgee:
I'm getting:

```
pocs/tests/test_camera.py::test_get_temp[pyro] SKIPPED                                                                                                                                                                                                                                                                 [ 16%]
pocs/tests/test_camera.py::test_set_set_point[pyro] SKIPPED                                                                                                                                                                                                                                                            [ 16%]
pocs/tests/test_camera.py::test_enable_cooling[pyro] SKIPPED                                                                                                                                                                                                                                                           [ 16%]
pocs/tests/test_camera.py::test_get_cooling_power[pyro] SKIPPED                                                                                                                                                                                                                                                        [ 16%]
pocs/tests/test_camera.py::test_disable_cooling[pyro] SKIPPED                                                                                                                                                                                                                                                          [ 17%]
pocs/tests/test_camera.py::test_exposure[pyro] The authenticity of host 'localhost (127.0.0.1)' can't be established.                                                                                                                                                                                                        
ECDSA key fingerprint is SHA256:/1m9Lp9ZnW/hIjUfsW0KN7348zK3gqhEmlzdrFZJQ5c.                                                                                                                                                                                                                                                 
Are you sure you want to continue connecting (yes/no)? FAILED
```

.coveragerc
```diff
@@ -1,11 +1,15 @@
 [run]
 branch = True
-concurrency = multiprocessing
+concurrency =
+    multiprocessing
```
wtgee: It would actually be great to split these and the travis changes into a separate PR

AnthonyHorton:
@wtgee Yes, for the file transfers `$PANUSER` needs to be able to do password-less ssh to the remote computer(s), which for the tests is `localhost`. Lines 36-39 of `.travis.yml` are what set this up for Travis:
```bash
ssh-keygen -b 2048 -t rsa -f ~/.ssh/id_rsa -N ''
touch ~/.ssh/authorized_keys
cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
ssh-keyscan -H localhost >> ~/.ssh/known_hosts
``` 
Looks like you just need the `ssh-keyscan` bit.

wtgee:
> Looks like you just need the ssh-keyscan bit.

Meaning everyone needs to run that on their local before they can do testing?

AnthonyHorton:
Yes. I should make a setup script for that.

.coveragerc
AnthonyHorton: Good point. Will do.

wtgee:
I needed

```bash
touch ~/.ssh/authorized_keys
cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
ssh-keyscan -H localhost >> ~/.ssh/known_hosts
```

pocs/utils/pyro/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    host = config.get('host', None)
+    if not host:
+        host = get_own_ip(verbose=True)
+    port = config.get('port', 0)
+
+    with Pyro4.Daemon(host=host, port=port) as daemon:
+        try:
+            name_server = Pyro4.locateNS()
+        except errors.NamingError as err:
+            warn('Failed to locate Pyro name server: {}'.format(err))
```
wtgee: Need to import `warn`

pocs/tests/conftest.py
```diff
@@ -41,3 +44,30 @@ def config_with_simulated_dome(config):
 @pytest.fixture
 def data_dir():
     return '{}/pocs/tests/data'.format(os.getenv('POCS'))
+
+
+def end_process(proc):
+    proc.send_signal(signal.SIGINT)
+    return_code = proc.wait()
```
wtgee: Did you want to return this `return_code`?

pocs/utils/pyro/__init__.py
AnthonyHorton: Oops. Need to watch out for that sort of thing when cutting & pasting functions between files!

pocs/tests/conftest.py
AnthonyHorton: Returning it would be a good idea from the point of view of making this a more generally useful function.  As it's currently used in these tests though I think any return value is just going to disappear down a black hole.  `end_process` is used as request finalizer for the `name_server` and `camera_server` fixtures, so only gets called when they go out of scope, at which point I don't think there's anything left to receive and act on the return value.

I should probably be checking the return value within `end_process` itself though, and emitting a warning if it's non-zero.  Might want to put a timeout on that wait, too, just in case.

wtgee:
FWIW, the `test_locate_name_server` seems to be hanging because the `name_server` is somehow already running. Could that somehow be related to the the timeout on the wait in `end_process`?

AnthonyHorton:
The `name_server` _should_ already be running. That's why I scoped the fixture as 'session'. I'm trying to get both `name_server` and `camera_server` to start up once, in that order (important), and run for the duration of the testing before being shut down cleanly by `end_process`.

When all the tests with a requirement for `name_server` or `camera_server` run within a reasonably period of time that seems to Work as Intended. When `test_locate_name_server` gets run 20 minutes later something seems to get confused though.  

What seems to be happening in that case is that `name_server` _is_ still running, but for mysterious reasons `test_locate_name_server` tries to start it again, which it _doesn't_ do if it runs earlier. The `pyro_name_server.py` script won't allow you to start a second instance though, and exits with an error code. The finalizer `end_process` then gets called to do cleanup, and gets stuck `wait`ing on a process that has already exited.

scripts/pyro_name_server.py
```diff
@@ -14,6 +14,6 @@
     parser.add_argument("--host", help="hostname or IP address to bind the server on")
     parser.add_argument("--port", help="port number to bind the server on")
     parser.add_argument("--autoclean", help="interval between registration autoclean (0=disabled)",
```
jamessynge: I'm unclear on what "registration autoclean" means.

scripts/pyro_camera_server.py
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 """
-Script to run a Pyro camera server. This should be run on the control computer for a distributed
-camera. The configuration for the camera is read from the pyro_camera.yaml/pyro_camera_local.yaml
-config file. The camera servers should be started after the name server, but before POCS.
+Script to run a Pyro camera server. This should be run on the camera control computer for a
+distributed camera. The configuration for the camera is read from the pyro_camera.yaml/
+pyro_camera_local.yaml config file. The camera servers should be started after the name server,
```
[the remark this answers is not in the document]
AnthonyHorton: Well, for now it involves a bunch of ssh sessions. My plan though is to remove the ordering requirement.

Need to get the `Observatory` to periodically check the name server and update its list of distributed cameras, rather than just doing this once as a result of the call of `create_cameras`. Similarly need to get the camera servers to check the name server and make sure they're registered, and also deregister on shutdown. With those modifications it would be possible to (re)start camera servers after starting POCS and, with a bit of care, allow the name server to be restarted while POCS is running too.

scripts/pyro_name_server.py
AnthonyHorton: Pyro objects register with the name server to allow their URI to be looked up by name, metadata matches, etc. The name server has an option to automatically remove the registrations of objects that it is unable to contact, useful for server crashes, network issues, etc. This is the 'autoclean' function, and to enable it you specify an interval between cleanings.

AnthonyHorton:
We've had some opportunity to do some testing with this code on real hardware recently, running up to 6 distributed cameras, with focusers, on Huntsman.  Testing showed that it is currently functional, but clunky to use.  Some helper scripts for (re)starting the system would, well, help, but the main issue is the already noted requirement to restart everything, from scratch, in the right order whenever there is a problem with an individual camera or focuser.

The structural changes in #612 provide a template for implementing dynamic camera addition/removal, which will go a long way towards making the system more robust and easier to work with. I hope to make some progress on that next week.

pocs/tests/test_aardvark.py
```diff
@@ -0,0 +1,45 @@
+import logging
```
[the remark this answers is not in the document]
AnthonyHorton: Done.

pocs/tests/conftest.py
```diff
@@ -41,3 +45,64 @@ def config_with_simulated_dome(config):
...
+    return ns_proc
+
+
+@pytest.fixture(scope='session')
+def camera_server(name_server, request):
+    cs_cmds = [os.path.expandvars('$POCS/scripts/pyro_camera_server.py'),
+               '--ignore_local']
+    cs_proc = subprocess.Popen(cs_cmds)
+    request.addfinalizer(lambda: end_process(cs_proc))
+    # Give camera server time to start up
```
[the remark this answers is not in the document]
AnthonyHorton: I don't think there's a way do this without rendering most of the tests in `test_aardvark.py` entirely redundant. These fixtures would have to locate the name sever and check that the camera server has registered with it, so if the fixtures don't error then the tests for those things in `test_aardvark.py` are guaranteed to pass. I guess that doesn't really matter too much, but it does effectively reduce 4 tests to 2 fixtures.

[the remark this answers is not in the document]
AnthonyHorton: Done.

AnthonyHorton:
I've brought this up to date with all the changes from `develop` and done a bit of tidying up. Now in a position to make a few functionality improvements.

wtgee:
Thanks for all the work @AnthonyHorton.

Closing in favor of: https://github.com/AstroHuntsman/huntsman-pocs/pull/80

PR #549, opened 2018-08-18
Tracking error fixes

wtgee:
This fixes an error with the direction of the declination tracking adjustment. Also moves the logic into the mount itself.

### DIFF
--- a/pocs/mount/mount.py
+++ b/pocs/mount/mount.py
@@ -320,6 +320,10 @@ class AbstractMount(PanBase):
         the movement of the camera box with respect to the pier, which can be
         determined from the Hour Angle (HA) of the pointing image in the sequence.
 
+        Note:
+            Correction values below 50ms will be skipped and values above 99999ms
+            will be clipped.
+
         Args:
             offset_info (`OffsetError`): A named tuple describing the offset
                 error. See `pocs.images.OffsetError`.
@@ -328,7 +332,7 @@ class AbstractMount(PanBase):
                 the direction of the Dec adjustment.
 
         Returns:
-            dict: Offset corrections for each axis ::
+            dict: Offset corrections for each axis as needed ::
 
                 dict: {
                     # axis: (arcsec, millisecond, direction)
--- a/pocs/mount/mount.py
+++ b/pocs/mount/mount.py
@@ -314,6 +314,12 @@ class AbstractMount(PanBase):
     def get_tracking_correction(self, offset_info, pointing_ha):
         """Determine the needed tracking corrections from current position.
 
+        This method will determine the direction and number of milliseconds to
+        correct the mount for each axis in order to correct for any tracking
+        drift. The Declination axis correction ('north' or 'south') depends on
+        the movement of the camera box with respect to the pier, which can be
+        determined from the Hour Angle (HA) of the pointing image in the sequence.
+
         Args:
             offset_info (`OffsetError`): A named tuple describing the offset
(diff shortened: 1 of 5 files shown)


pocs/observatory.py
```diff
@@ -342,23 +342,49 @@ def update_tracking(self):
         if self.current_offset_info is not None:
             self.logger.debug("Updating the tracking")
 
+            # Get the pier side of pointing image
```
jamessynge: Could you add some more comments? I'm certainly ignorant about what is happening here, and the lack of problem description in the PR isn't helping.

pocs/mount/mount.py
```diff
@@ -72,8 +72,8 @@ def __init__(self, location, commands=None, *args, **kwargs
         self._state = 'Parked'
 
         self.sidereal_rate = ((360 * u.degree).to(u.arcsec) / (86164 * u.second))
-        self.ra_guide_rate = 0.5  # Sidereal
-        self.dec_guide_rate = 0.5  # Sidereal
+        self.ra_guide_rate = 0.9  # Sidereal
```
jamessynge: Does guide rate mean the speed with which we make adjustments to the position between images?

pocs/mount/mount.py
```diff
@@ -311,6 +311,106 @@ def distance_from_target(self):
 
         return separation
 
+    def get_tracking_correction(self, offset_info, pointing_ha):
+        """Determine the needed tracking corrections from current position.
+
+        Args:
+            offset_info (`OffsetError`): A named tuple describing the offset
+                error. See `pocs.images.OffsetError`.
+            pointing_ha (float): The Hour Angle (HA) of the mount at the
```
jamessynge: Is the point here that the mount will do a meridian flip to go from 89 degrees to 91 degrees (90 being pointing to the meridian)? And thus the HA will tell us if we started observing with the cameras on the left or right side of the pier?

pocs/observatory.py
```diff
@@ -342,67 +341,24 @@ def update_tracking(self):
...
+            except AttributeError:
+                pass
+
+            self.logger.debug("Pointing HA: {}".format(pointing_ha))
+            correction_info = self.mount.get_tracking_correction(
+                self.current_offset_info,
+                pointing_ha
+            )
+
+            try:
```
jamessynge: Is it worth putting in a limit, where we don't make a correction if the error is sufficiently low?

wtgee: It happens back in [`get_tracking_correction`](https://github.com/panoptes/POCS/blob/8d6d6aa0926ef4362964eec22530691698e13516/pocs/mount/mount.py#L366) where it checks for anything under 50 ms. Nothing goes into the returned dict so the `correct_tracking` method then just skips that whole axis. 

The 50 ms was determined empirically by me where I noticed when it was shorter than about 40ms it would just happen too quickly and then the system would get stuck waiting. It's also less than 1 pix of movement.

I'll make a note in the docstring to `get_tracking_correction` about how the values are clipped for min max.

pocs/mount/mount.py
wtgee: I added some more info in the docstring.

pocs/mount/mount.py
wtgee: This is documented somewhere but the fact that I don't know where and that you have a question here makes that irrelevant.

I want to turn these into properties of the class, which would be more consistent with all these other values and would allow for an easier docstring. - #552

PR #551, opened 2018-08-19
Auto start social

jamessynge:
Start the social messaging script when a PANOPTES unit starts up.

Attempt (not yet successful) to eliminate the problem with log format (fileline being unknown).

### DIFF
--- a/resources/targets/simple.yaml
+++ b/resources/targets/simple.yaml
@@ -1,8 +1,4 @@
 -
-    name: Kepler 488 b
-    position: 18:51:51.88s +48:20:42.3s
-    priority: 125
--
     name: Kepler 1100
     position: 19h27m29.10s +44d05m15.00s
     priority: 125
--- a/scripts/run_social_messaging.py
+++ b/scripts/run_social_messaging.py
@@ -94,6 +94,10 @@ def run_social_sinks(msg_port, social_twitter, social_slack):
 
 
 if __name__ == '__main__':
+    # This is early in startup to ensure the PANOPTES logging code is installed early.
+    pocs_config = load_config(config_files=['pocs'])
+    the_root_logger = get_root_logger()
+
     parser = argparse.ArgumentParser(
         description='Run social messaging to forward platform messages to social channels.')
     parser.add_argument(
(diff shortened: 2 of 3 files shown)


resources/targets/simple.yaml
```diff
@@ -1,3 +1,7 @@
+-
+    name: Kepler 488 b
+    position: 18:51:51.88s +48:20:42.3s
```
wtgee: 1) this file shouldn't be in this PR. And yes, we need a better way for targets to be changed. :)
2) I would be surprised if this parses in this format, with the `:` instead of the `hms`

scripts/run_social_messaging.py
```diff
@@ -94,6 +94,10 @@ def run_social_sinks(msg_port, social_twitter, social_slack):
 
 
 if __name__ == '__main__':
+    # This is early in startup to ensure the PANOPTES logging code is installed early.
+    pocs_config = load_config(config_files=['pocs'])
```
wtgee: Just a note that it does `pocs` by default so you could just `load_config()` edit: and I see you are just moving it anyway.

scripts/startup/start_panoptes_in_tmux.sh
```diff
@@ -97,7 +97,13 @@ set -x
 
 # Create a window running the zeromq message forwarders.
 # These provide connections between message publishers and subscribers.
-create_and_init_window messaging start_messaging_hub.sh
+create_and_init_window msg_hub start_messaging_hub.sh
+
+# Create a window running the social messaging program. It subscribes to
+# zeromq and posts some messages to Twitter or Slack, if
+# conf_files/pocs_local.yaml has the appropriate config values; if not,
+# the program just exits.
```
wtgee: Does it close the tmux window when that happens or leave a shell?

scripts/startup/start_panoptes_in_tmux.sh
```diff
@@ -97,7 +97,13 @@ set -x
 
 # Create a window running the zeromq message forwarders.
 # These provide connections between message publishers and subscribers.
-create_and_init_window messaging start_messaging_hub.sh
+create_and_init_window msg_hub start_messaging_hub.sh
+
+# Create a window running the social messaging program. It subscribes to
+# zeromq and posts some messages to Twitter or Slack, if
```
wtgee: nitpick: could expand beyond twitter or slack

resources/targets/simple.yaml
jamessynge: Sorry, this wasn't supposed to be part of the change.
And you're right, this format wouldn't parse. OTOH, it would be nice if it did since that is what some sites present for the coords.

scripts/run_social_messaging.py
jamessynge: I'll drop the config_files=['pocs'].

scripts/startup/start_panoptes_in_tmux.sh
jamessynge: Agreed, wasn't sure what to say for a more general description... after all, it could post to any site, not just a social messaging service.

scripts/startup/start_panoptes_in_tmux.sh
jamessynge: It will leave the shell, so you can see any error messages.

PR #557, opened 2018-08-25
Canon init fixes

### LINKED ISSUE #503: DATE-OBS from DSLRs can be wrong
The DATE-OBS fits header is coming directly from the EXIF information embedded in the CR2 file, which corresponds to the time that is set on the camera. This can be off or otherwise inconsistent with the control computer. Should have one source of time for DATE-OBS. Should just match the timestamp in the file name (and from IMAGE_ID)

wtgee:
* Fix date on cameras (closes #503)
* Camera suggestions from Aru.

### DIFF
--- a/pocs/camera/canon_gphoto2.py
+++ b/pocs/camera/canon_gphoto2.py
@@ -35,6 +35,8 @@ class Camera(AbstractGPhotoCamera):
 
         # Properties to be set upon init.
         prop2index = {
+            '/main/settings/datetime': 'now',             # Current datetime
+            '/main/settings/datetimeutc': 'now',          # Current datetime
             '/main/actions/viewfinder': 1,                # Screen off
             '/main/settings/autopoweroff': 0,             # Don't power off
             '/main/settings/reviewtime': 0,               # Screen off after taking pictures
--- a/scripts/take_bias.sh
+++ b/scripts/take_bias.sh
@@ -30,15 +30,15 @@ P=$1
 SAVE_FILE=${2:-"bias-%Y%m%d-%H%M%S.cr2"}
 echo "Taking bias frames on port ${P}: ${SAVE_FILE}"
 
-NUM_FRAMES=${4:-1}
-SHUTTER_INDEX=${3:-52} # 1/4000s on EOS 100D
+NUM_FRAMES=${3:-1}
+SHUTTER_INDEX=${4:-52} # 1/4000s on EOS 100D
 
 # Set to fast speed
 gphoto2 --port=${P} --set-config-index shutterspeed=${SHUTTER_INDEX}
 
 COUNTER=0
 until [[ ${COUNTER} -eq ${NUM_FRAMES} ]]; do
-    gphoto2 --port=${P} --filename=${SAVE_FILE} --capture-image-and-download
+    gphoto2 --port=${P} --filename=${SAVE_FILE}_${COUNTER} --capture-image-and-download
     let COUNTER+=1
 done
 
--- a/scripts/take_pic.sh
+++ b/scripts/take_pic.sh
@@ -7,8 +7,13 @@ echo 'Taking picture'
 echo "T = ${T}s"
 
 # Open shutter
-gphoto2 --port=${P} --set-config eosremoterelease=Immediate \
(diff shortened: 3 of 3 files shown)


pocs/camera/canon_gphoto2.py
```diff
@@ -35,6 +35,8 @@ def connect(self):
 
         # Properties to be set upon init.
         prop2index = {
+            '/main/settings/datetime': 'now',             # Current datetime
```
jamessynge: Note: this is untested!

I see in the manual and elsewhere this config property:
```
/main/actions/syncdatetime
Label: Synchronize camera date and time with PC
Type: TOGGLE
```
It might be another way to achieve this result.

scripts/take_bias.sh
```diff
@@ -30,15 +30,15 @@ P=$1
 SAVE_FILE=${2:-"bias-%Y%m%d-%H%M%S.cr2"}
 echo "Taking bias frames on port ${P}: ${SAVE_FILE}"
 
-NUM_FRAMES=${4:-1}
-SHUTTER_INDEX=${3:-52} # 1/4000s on EOS 100D
+NUM_FRAMES=${3:-1}
```
jamessynge: How about 4 lines in a row for $1 through $4, then the echo "Taking bias ...

And a comment for those unfamiliar with bash details that the default value if the 3rd param isn't supplied is 1, not -1 which one might suspect.

scripts/take_bias.sh
```diff
@@ -30,15 +30,15 @@ P=$1
...
+SHUTTER_INDEX=${4:-52} # 1/4000s on EOS 100D
```
jamessynge: How about a TODO or Issue that we should allow exposure time to be specified, or how to get the list of indices and corresponding exposure times supported by the camera?

scripts/take_bias.sh
```diff
@@ -30,15 +30,15 @@ P=$1
...
 
 # Set to fast speed
 gphoto2 --port=${P} --set-config-index shutterspeed=${SHUTTER_INDEX}
 
 COUNTER=0
 until [[ ${COUNTER} -eq ${NUM_FRAMES} ]]; do
-    gphoto2 --port=${P} --filename=${SAVE_FILE} --capture-image-and-download
+    gphoto2 --port=${P} --filename=${SAVE_FILE}_${COUNTER} --capture-image-and-download
```
jamessynge: Hmm. This seems problematic because the _0, _1, etc. will appear after the file extension (e.g. cr2).
I'd be inclined to add the count to the name only if NUM_FRAMES is > 1, and then to insert it before the file name extension.

scripts/take_pic.sh
```diff
@@ -7,8 +7,13 @@ echo 'Taking picture'
 echo "T = ${T}s"
```
jamessynge: We definitely need some docs at the top of this file! Sigh.

pocs/camera/canon_gphoto2.py
wtgee: I saw this setting but wasn't quite clear how it worked. I toggled it but it didn't seem to update the time.

I'll add some tests. Need to get into those real hardware tests to take a look.

scripts/take_bias.sh
wtgee: I think this is what Aru said she modified so I was waiting to see if we get that.

scripts/take_bias.sh
wtgee: Good point. I didn't think that through. It's easy enough to do a shell substitution (i.e. `${SAVE_FILE/\.cr2/_${COUNTER}\.cr2}` however I suspect the real issue here is the output directory, not really the filename. The default name has a timestamp so would be okay for any NUM_FRAME. What about just having an OUTPUT_DIR argument and keep the default filename?

wtgee: Actually, maybe I don't like that idea. I'll probably just add the counter if NUM_FRAMES > 1 and SAVE_FILE is not empty.

wtgee:
I'm going to close this for now. The camera time fix will come along once I can add some tests and the script for taking pictures will be split into separate PR. Thanks for comments, will incorporate into new PRs.

PR #560, opened 2018-08-25
Fix errors with tracking updates

wtgee:
* Skip axis correction if no correction information
* Fix error

### DIFF
--- a/pocs/mount/mount.py
+++ b/pocs/mount/mount.py
@@ -405,6 +405,9 @@ class AbstractMount(PanBase):
             `error.Timeout`: Timeout error.
         """
         for axis, corrections in correction_info.items():
+            if correction_info is None:
+                continue
+
             offset = corrections[0]
             offset_ms = corrections[1]
             delta_direction = corrections[2]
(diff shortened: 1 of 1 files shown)


pocs/mount/mount.py
```diff
@@ -405,21 +405,24 @@ def correct_tracking(self, correction_info, axis_timeout=30.):
             `error.Timeout`: Timeout error.
         """
         for axis, corrections in correction_info.items():
+            if correction_info is None:
```
jamessynge: This looks misplaced. You'll never get to this line, and in fact will have raised an exception at the previous line if correction_info is None.

So, is this supposed to be `if not axis or not corrections:`?

wtgee: It doesn't need to be because the keys are set in `get_tracking_correction`. And the `.items()` wouldn't iterate over a non-key.

jamessynge: Yes, I get that .items won't return something that isn't a key. But if correction_info is None, then the for loop will crash, so you'll never get to 408.

p.s. I think that None is a valid key in a dictionary, so axis could be None, as could corrections, at line 408, given the full generality of dictionaries.

wtgee: Oh, sorry, I see. Yes, it should be `if corrections is None`, not `correction_info`. I'll add the `axis` in just for good measure.

PR #566, opened 2018-08-28
Coverage calculation accounting for concurrency

AnthonyHorton:
While working on #547 I ran into problems with inaccurate test coverage reports because the existing `coverage` and Travis CI configurations produced results which did not take into account concurrency, in particular scripts started with `subprocess`.  I was able to come up with some changes to the configurations that appear to resolve these problems, and should properly calculate test coverage involving `multiprocessing`, `threading` and `subprocess` concurrency. At @wtgee's suggestion I have split these config changes out into a Pull Request of their own.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,11 +1,15 @@
 [run]
 branch = True
-concurrency = multiprocessing
+concurrency =
+    multiprocessing
+    threading
+    subprocess
 source =
     peas
     pocs
-omit = 
+omit =
 	pocs/utils/data.py
+parallel = True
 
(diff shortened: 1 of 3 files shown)


AnthonyHorton:
@wtgee It would make more difference if `$POCS/scripts/*.py` were included in the coverage calculation! Python scripts launched by tests using `subprocess` are properly counted with this set up, which is the main difference I was after.

jamessynge:
I'm looking forward to trying this out, and hopefully getting some improved coverage.

PR #568, opened 2018-08-29
Add targets Kepler 488 and HAT-P-19

jamessynge:
Add Kepler-488 b, which has a 19.2 mmag dip,
and HAT-P-19 b, which has a depth of 21.6 mmag.

### DIFF
--- a/resources/targets/simple.yaml
+++ b/resources/targets/simple.yaml
@@ -1,4 +1,8 @@
 -
+    name: Kepler 488 b
+    position: 18h51m51.88s +48d20m42.3s
+    priority: 125
+-
     name: Kepler 1100
     position: 19h27m29.10s +44d05m15.00s
     priority: 125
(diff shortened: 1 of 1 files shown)


resources/targets/simple.yaml
```diff
@@ -1,3 +1,7 @@
+-
+    name: Kepler 488 b
```
wtgee: One issue we need to deal with is consistency in naming and fields. These automatically get uploaded to the storage buckets with these names.

I have been putting the name of the host star, not the planet itself. So just `Kepler 488`. This could feasibly change in the future when you have a timed observation and are going after a different exoplanet but until then I think we should just list the host star.

jamessynge: Will do.

wtgee:
Glad this increased our code coverage. :) Clearly something is still not right with that.

PR #570, opened 2018-08-29
Fix default transition

### LINKED ISSUE #565: Default next method incorrect
The call to get the next state incorrectly returns the `park` string instead of the `park` method.

```
        caller = getattr(self, call_method, 'park')
        state_changed = caller()
```

should be 
```
        caller = getattr(self, call_method, self.park)
        state_changed = caller()
```

Mentioned in #548.

wtgee:
Fixes #565

### DIFF
--- a/pocs/state/machine.py
+++ b/pocs/state/machine.py
@@ -172,15 +172,27 @@ class PanStateMachine(Machine):
                 self.sleep(60)
 
     def goto_next_state(self):
+        """Make a transition to the next state.
+
+        Each state is responsible for setting the `next_state` property based
+        off the logic that happens inside the state. This method will look up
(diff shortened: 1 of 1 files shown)


pocs/state/machine.py
```diff
@@ -179,7 +179,7 @@ def goto_next_state(self):
 
         self.logger.debug("Transition method: {}".format(call_method))
 
-        caller = getattr(self, call_method, 'park')
+        caller = getattr(self, call_method, self.park)
```
jamessynge: Could you change the name of call_method to transition_method_name and caller to transition_method? That would help me when coming in here and wondering what is going on.

PR #574, opened 2018-08-31
Improve testing of pocs.scheduler.field.Field

jamessynge:
Require that the field name be non-empty.

This is a precursor to adding support for colon separated hour angle or degree values.

### DIFF
--- a/pocs/scheduler/field.py
+++ b/pocs/scheduler/field.py
@@ -30,6 +30,8 @@ class Field(FixedTarget, PanBase):
         super().__init__(SkyCoord(position, equinox=equinox, frame='icrs'), name=name, **kwargs)
 
         self._field_name = self.name.title().replace(' ', '').replace('-', '')
+        if not self._field_name:
+            raise ValueError('Name is empty')
 
 
 ##################################################################################################
--- a/pocs/tests/test_field.py
+++ b/pocs/tests/test_field.py
@@ -1,5 +1,8 @@
 import pytest
 
+from astropy import units as u
+from astropy.coordinates import Latitude, Longitude
+
 from pocs.scheduler.field import Field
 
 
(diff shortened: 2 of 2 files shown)


jamessynge:
As noted elsewhere, we have an issue with our coverage info. The only file that should see any change is pocs/scheduler/field.py, but we see changes to 4 other files. Perplexing.

jamessynge:
And interestingly, those 4 changes are in areas I've worked on: dome and arduino. Not sure if that is coincidence, or if my test style results in inconsistent coverage.

wtgee:
> Not surprisingly, I've already tested the solution that I have in mind,  which doesn't require that units be passed in to field. Instead, I  modify Field to pass the units (hourangle and degree) into SkyCoord,  which handles both the existing format and the new one. Does that seem  OK?

Actually, no, that limits our flexibility. Besides just entering it with a colon delimiter someone might want to just enter the actual degrees. More likely, as we go toward charting TESS fields we will probably be using galactic coordinates rather than RA/Dec. I think it makes more sense to pass an actual `SkyCoord` to the `Field`.

jamessynge:
What syntax would you propose for the Galactic Coords? Or more generally, what should we put in the targets file to support creating a SkyCoord to be passed into a Field ctor?

wtgee:
The existing syntax is what I would propose since the unit is contained within the string.

I don't think switching the `:` to the appropriate unit is a heavy burden on someone and in the future the hope is people won't be interacting with the yaml files directly anyway so it would be a non-issue. But there is no reason to limit what a `Field` can accept for `SkyCoord` params when really it should work with any `SkyCoord`. I honestly don't see the need to support a unit-less string.

jamessynge:
I meant "what syntax will be used with galactic coords?", and "how is that differentiated from the existing RA and DEC?".

wtgee:
Galactic coordinates make most sense simply in degrees of longitude (l) and latitude (b). We will also have to change the coordinate frame in that case, which means the Field needs to either know about the position, the units, and the frame, or it just needs to accept a `SkyCoord`.

I'm not opposed to having `Field` do some parsing and properly build a `SkyCoord` and ideally it should probably accept either. Might just need to have a larger conversation about the `Field`s and scheduler related changes? 

What is the current goal for this PR and the related link above? To make it easier to add things to the targets file (and make more robust)? Or is this also work toward a larger change with the `Field`s?

(And I think this PR is fine regardless so maybe we should merge and move this conversation?)

jamessynge:
OK, will merge and take up the discussion elsewhere.

Issue #564, opened 2018-08-27
Config File Updates Overview


wtgee:
* [ ] Add lots of documentation to config file to explain options.
* [x] ~~Rely on config file instead of env vars (#555).~~
* [ ] Ability to override individual nested keys.
  * [x] Via config server
  * [x] Via pocs_local.yaml file.
* [x] pyserial options #181 
* [x] Save config in local file #238 
* [x] issues with testing and config #259 
* [x] Target constraints in config #402 

Maybe:
* [x] ~~Have a directory config file in $PANDIR and a script for parsing (discussed in #555).~~
* [x] Add a `get_config_value('mount.serial.port')` that can be used to retrieve a single (nested) value, mostly for use in utility functions. _-or-_
* [x] Add a config query service.
* [x] ~~Convert config to [`ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap) rather than using a simple dict update. This allows for updates to underlying configs and easier overriding.~~
* [ ] Consider using the [pydantic Settings](https://pydantic-docs.helpmanual.io/usage/settings/) module.

Would be nice:
* [x] Web/notebook interface (https://github.com/panoptes/panoptes-tutorials/issues/48)

PR #577, opened 2018-09-02
Cleanup google network utils - WIP

wtgee:
This is a cleanup of some of the network utilities. This is prompted both by the desire to have "real-time" uploads as observations are happening as well as to have a suite of tools that can be used in google compute instances for interacting with the storage.

The main upload from the `observatory` now calls out to the `scripts/process_image_dir.py` script. Changes include the uploading of all files in the directory, not just the fits files, as well as the saving (and thus uploading) of all the jpgs. We could consider making and uploading the timelapse into same directory.

Also allows `solve_field` to handle `fz` without decompression.

Also includes the `clouddb` module for interacting with the cloud sql db. Still need to write a few tests for that but have been using it frequently in PIAA. The PGPASSWORD mechanism will probably update in a few days.


pocs/utils/google/storage.py
```diff
@@ -88,3 +91,239 @@ def upload(self, local_path, remote_path=None):
...
+                                 bucket='panoptes-survey',
+                                 **kwargs):
+    """Upload an observation directory to google cloud storage.
+
+    This is a convenience function for bulk uploading an observation folder to a
+    bucket. This assumes that observations are placed within `/images/fields`
+    and follow the normal naming convention for observations.
+
+    Note:
+        This requires that the command line utility `gsutil` be installed
```
jamessynge: It does seem odd that we are taking two approaches to dealing with GCS, esp. authorization. Should we just require that `gcloud login` (or `gcloud ...service_account...`) has been done?

pocs/utils/google/storage.py
```diff
@@ -88,3 +91,239 @@ def upload(self, local_path, remote_path=None):
...
+        field_dir = dir_name.split('/fields/')[-1]
+        remote_path = os.path.normpath(os.path.join(
+            bucket,
+            pan_id,
+            field_dir
+        ))
+
+        # normpath strips the trailing slash so add here so files go in directory
+        destination = 'gs://{}/'.format(remote_path)
+        run_cmd = [gsutil, '-mq', 'cp', '-r', img_path, destination]
```
jamessynge: I just tried running gsutil locally and it failed with:
```
ERROR: Python 3 and later is not compatible with the Google Cloud SDK. Please use Python version 2.7.x.

If you have a compatible Python interpreter installed, you can use it by setting
the CLOUDSDK_PYTHON environment variable to point to it.
```
How is it working for you?

scripts/process_image_dir.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
```
jamessynge: The filename is not at all descriptive. How about upload_observation.py?

conf_files/pocs.yaml
```diff
@@ -52,6 +50,18 @@ messaging:
     # Must match ports in peas.yaml.
     cmd_port: 6500
     msg_port: 6510
+
+# Google Network #
+panoptes_network:
+    auth_key: # Update
```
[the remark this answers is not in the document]
wtgee: Yes, I think that is preferred so I will change it so it accepts either. I thought I already had actually but either way I'll make sure it works. I think we should encourage the gcloud usage as you suggest.

conf_files/pocs.yaml
```diff
@@ -52,6 +50,18 @@ messaging:
     # Must match ports in peas.yaml.
     cmd_port: 6500
     msg_port: 6510
+
```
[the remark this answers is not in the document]
wtgee: #564

pocs/utils/google/clouddb.py
```diff
@@ -0,0 +1,195 @@
...
+        host (str, optional): Hostname, default localhost.
+        db_user (str, optional): Name of db user, default 'panoptes'.
+        db_name (str, optional): Name of db, default 'postgres'.
+        port (int, optional): DB port.
+
+    Returns:
+        `psycopg2.Connection`: DB connection object.
+    """
+    try:
+        pg_pass = os.environ['PGPASSWORD']
```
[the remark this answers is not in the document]
wtgee: It's set by postgres.

pocs/utils/google/storage.py
wtgee: The `gsutil` is part of `gcloud` so I don't think they are mutually exclusive. 

Agree with the authentication as noted above, so we'll move toward the `gcloud` service key authorization.

pocs/utils/google/storage.py
wtgee: Oh, hm. Perhaps I had this set up earlier under python 2.7. Let me look into my setup. In my head the gsutil had moved up to python 3.

scripts/process_image_dir.py
wtgee: I actually changed it from that name. :)  It includes the cleanup in the script, which is also compressing fits files, making timelapse, etc. so it is doing more than just uploading. But agree it's not the best name whereas just uploading is pretty obvious.

pocs/utils/google/storage.py
wtgee: So gcloud and gsutil come from the same place and both require python 2.7. The gsutil is nice for uploading because it transparently handles parallel uploads while the python3 client library does not. We can use the service key to authenticate with either. If we wanted to completely avoid python2.7 we could just use the client libraries, just missing that parallel ability, which is very nice to have.

The gcloud and gsutil can come from the ubuntu package manager, so part of #578 .

scripts/process_image_dir.py
jamessynge: Perhaps have the cleaning as an option controlled by flag(s), with uploading as the default behavior.

pocs/tests/utils/google/test_google_storage.py
```diff
@@ -0,0 +1,111 @@
...
+import os
+import re
+import shutil
+
+from pocs.utils.config import load_config
+from pocs.utils.error import GoogleCloudError
+from pocs.utils.google.storage import PanStorage, upload_observation_to_bucket
+
+
+pytestmark = pytest.mark.skipif(
```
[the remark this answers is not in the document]
wtgee: Haven't mocked this yet. The idea was to test actual connection but should also be testing process. Will add.

scripts/process_image_dir.py
wtgee: The primary purpose of this script is to cleanup the directory. The uploading (and sending of metadata) are the optional arguments to that.

pocs/utils/google/clouddb.py
```diff
@@ -3,42 +3,43 @@
 from warnings import warn
 
 import psycopg2
```
jamessynge: Please change the name of the path & file to something other than pocs/utils/google/clouddb.py. I get that it is being used to access a Google offering of postgres, but it is being accessed using a generic postgres library, so how about just calling it pocs/utils/db/postgres.py?

scripts/process_image_dir.py
```diff
@@ -5,56 +5,80 @@
...
 
-    def _print(msg):
-        if verbose:
-            print(msg)
-
-    _print("Cleaning observation directory")
-    clean_observation_dir(directory, remove_jpgs=False)
+    _print("Cleaning observation directory: {}".format(directory))
+    try:
+        clean_observation_dir(directory, remove_jpgs=False, verbose=verbose, **kwargs)
```
jamessynge: How about adding remove_jpgs to the command line flags, instead of hard-coding it? Could be in another PR.

scripts/process_image_dir.py
```diff
@@ -5,56 +5,80 @@
...
     if upload:
         _print("Uploading to storage bucket")
-        upload_observation_to_bucket(pan_id, directory, include_files='*')
+
+        uploaded_files_fn = os.path.join(directory, 'uploaded_files.txt')
+
+        if os.path.exists(uploaded_files_fn):
+            _print("Files have been uploaded already, skipping")
+        else:
+            img_path = upload_observation_to_bucket(
```
jamessynge: The name `img_path` implies this is the path of a single image. The `include_files='*'` implies that non-image files may be uploaded. Further, `img_path` is then passed to glob, implying that it is not a glob.

pocs/utils/google/clouddb.py
wtgee: Funny how it evolved to that point and I didn't even realize it. Will change, thanks.

scripts/process_image_dir.py
wtgee: I cleaned this up ab it but it's still not very intelligent. It merely records all files in the directory that were uploaded, assuming the upload was successful.

wtgee:
Growing out of control. I am going to split this PR into three separate ones:

* updates to google storage
* updates for google cloudsql (postgres db connections)
* clean observations script and dirs

PR #578, opened 2018-09-04
Updating install scripts for Ubuntu 18.04 - WIP

jamessynge:
Includes updating some dependencies/requirements.

This doesn't fully work yet. astroscrappy build
during install fails. And this is only tested as
by running it on my laptop; I've not yet been able
to run pytest on POCS.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,33 +1,14 @@
-astroplan
-astropy >= 3.0.0
-ccdproc
-codecov
-coveralls
-dateparser
+# These are packages that are NOT available via our
+# choice of conda repositories; there are less well
+# known repositories that do offer them, but that is
+# more of a security risk.
+
+# Python wrapper for ffmpeg.
 ffmpy
-gcloud
-google-cloud-storage
-matplotlib >= 2.0.0
+
+# Mock object support for Python testing.
 mocket
-numpy >= 1.12.1
-pycodestyle == 2.3.1
-pymongo >= 3.2.2
-pyserial >= 3.1.1
-pytest >= 3.4.0
-python-dateutil >= 2.5.3
-PyYAML >= 3.11
-pyzmq >= 15.3.0
-readline
-requests
-scikit-image >= 0.12.3
-scipy >= 0.17.1
 
-# Jupyter-notebook requires Tornado versions between 4 and 5. See:
-# https://stackoverflow.com/questions/49141525/install-jupyter-notebook-on-miniconda
-# https://github.com/ipython/ipython/issues/8249
-# https://github.com/jupyter/help/issues/324
-
-tornado == 4.5.3
+# State machine library for Python.
 transitions >= 0.4.0
-tweepy
-wcsaxes
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 # known repositories that do offer them, but that is
 # more of a security risk.
 
-# Python wrapper for ffmpeg.
+# Thin Python wrapper for ffmpeg. We *could* replace with our own wrapper.
 ffmpy
 
 # Mock object support for Python testing.
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,12 +14,12 @@ pycodestyle == 2.3.1
 pymongo >= 3.2.2
 pyserial >= 3.1.1
 pytest >= 3.4.0
-python_dateutil >= 2.5.3
+python-dateutil >= 2.5.3
 PyYAML >= 3.11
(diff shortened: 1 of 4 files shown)


AnthonyHorton:
I ran into problems with `astroscrappy` and `healpy` a little while back and the root cause seemed to be an incompatibility between `Cython` and Python 3.7, which `conda` now installs by default. Reverting to Python 3.6 was a successful workaround for me.

requirements.txt
```diff
@@ -1,33 +1,14 @@
...
-codecov
-coveralls
-dateparser
+# These are packages that are NOT available via our
+# choice of conda repositories; there are less well
+# known repositories that do offer them, but that is
+# more of a security risk.
+
+# Python wrapper for ffmpeg.
 ffmpy
```
wtgee: We could probably lose the option. It is just a thin wrapper around ffmpeg. Since we don't have many varying options it would be trivial to write our own script.

scripts/install/conda-packages-list.txt
```diff
@@ -1,3 +1,33 @@
-pandas
+astroplan
+astropy >= 3.0.0
+ccdproc
```
wtgee: Even though conda fixes the problems with this (via astroscrappy), we are not actually using it anywhere in the repo. Might as well just remove.

requirements.txt
```diff
@@ -1,33 +1,14 @@
...
-requests
-scikit_image >= 0.12.3
-scipy >= 0.17.1
 
-# Jupyter-notebook requires Tornado versions between 4 and 5. See:
-# https://stackoverflow.com/questions/49141525/install-jupyter-notebook-on-miniconda
-# https://github.com/ipython/ipython/issues/8249
-# https://github.com/jupyter/help/issues/324
+# Mock object support for Python testing.
+mocket
```
wtgee: Could potentially work around this. Since we have `requests` elsewhere now it might be cleaner to switch TheSkyX calls to `requests` and then use some of the mock tools provided for it, which are available.

requirements.txt
```diff
@@ -1,33 +1,14 @@
...
 
-tornado == 4.5.3
+# State machine library for Python.
 transitions >= 0.4.0
```
wtgee: And technically I'm still a maintainer on this project though I haven't done anything for years. :confused: I can find out if there is any plan.

scripts/install/conda-packages-list.txt
```diff
@@ -1,3 +1,33 @@
...
+google-cloud-sdk
+google-cloud-storage
+matplotlib >= 2.0.0
+numpy >= 1.12.1
+pycodestyle == 2.3.1
+pymongo >= 3.2.2
+pyserial >= 3.1.1
+pytest >= 3.4.0
+pandas
+python
```
wtgee: Does this actually install the python version from here and if so should we specify a version?

scripts/install/conda-packages-list.txt
jamessynge: Remove ccdproc?

requirements.txt
jamessynge: I'm trying to keep changes in this PR to a minimum (i.e. it will be large enough as is). I'd like to focus on just the switch to conda and 18.04. In fact, maybe I already shoved to much into this, and should extract the 18.04 (Mongodb) changes. Sigh. Lots of slow tests required.

scripts/install/conda-packages-list.txt
jamessynge: Specifying a version is problematic. If you say ==3.7, then the constraints are unsolvable.

wtgee: Gotcha.

requirements.txt
wtgee: Agreed, don't want to do it for this PR, mostly just commenting on what I think we could eventually get rid of.

scripts/install/conda-packages-list.txt
wtgee: I think so. It will come through in PIAA probably anyway but might as well get rid of it here since it's not an actual dependency.

requirements.txt
```diff
@@ -1,26 +1,14 @@
...
-ccdproc
-codecov
-coveralls
-dateparser
+# These are packages that are NOT available via our
+# choice of conda repositories; there are less well
+# known repositories that do offer them, but that is
+# more of a security risk.
+
+# Thin Python wrapper for ffmpeg. We *could* replace with our own wrapper.
```
wtgee: #591 replaces this.

Issue #579, opened 2018-09-04
Move pointing logic out of state


wtgee:
Similar to the `observe`, `analyze_recent`, and `update_tracking`, the bulk of the logic for the pointing state should be moved outside of the state itself and into the observatory.

Issue #340, opened 2018-01-14
Decide on correcting separation after pointing


wtgee:
In the pointing class we have a bunch of commented out code that used to be used to sync the mount after each pointing. ioptron manual says not to do this so it's been commented out. Need to capture the functionality elsewhere and remove commented out code.

wtgee:
Code removed in #356

wtgee:
Closed by #580

PR #580, opened 2018-09-04
Correct pointing option

wtgee:
This adds an `auto_correct` option to the pointing configuration that will "sync" the mount with the current coordinates in an attempt to minimize the pointing error.

See also #579, #340 

Note: we used this by default for a long time then removed it as we were debugging pointing/tracking issues on PAN001. I've added it back to PAN001 and have been using it recently.

Edit: this also includes a fix to the pointing headers which is actually used in tracking.

### DIFF
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -38,8 +38,8 @@ mount:
     non_sidereal_available: True
 pointing:
     auto_correct: True
-    threshold: 0.05
-    exptime: 30
+    threshold: 0.05 # degrees
+    exptime: 30 # seconds
     max_iterations: 3
 cameras:
     auto_detect: True
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -38,7 +38,7 @@ mount:
     non_sidereal_available: True
 pointing:
     auto_correct: True
-    threshold: 0.05 # degrees ≈ 180" ≈ 18 pixels 
+    threshold: 500 # arcseconds ~ 50 pixels
     exptime: 30 # seconds
     max_iterations: 3
 cameras:
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -37,6 +37,7 @@ mount:
         baudrate: 9600
     non_sidereal_available: True
 pointing:
+    auto_correct: True
     threshold: 0.05
     exptime: 30
     max_iterations: 3
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -38,7 +38,7 @@ mount:
     non_sidereal_available: True
 pointing:
     auto_correct: True
-    threshold: 0.05 # degrees
+    threshold: 0.05 # degrees ≈ 180" ≈ 18 pixels 
     exptime: 30 # seconds
     max_iterations: 3
 cameras:
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -37,7 +37,7 @@ mount:
         baudrate: 9600
     non_sidereal_available: True
 pointing:
-    auto_correct: True
+    auto_correct: False
     threshold: 500 # arcseconds ~ 50 pixels
     exptime: 30 # seconds
     max_iterations: 3
--- a/pocs/state/states/default/pointing.py
+++ b/pocs/state/states/default/pointing.py
@@ -20,7 +20,7 @@ def on_enter(event_data):
     pointing_config = pocs.config['pointing']
     num_pointing_images = pointing_config.get('max_iterations', 3)
     should_correct = pointing_config.get('auto_correct', False)
-    pointing_threshold = pointing_config.get('threshold', False)
+    pointing_threshold = pointing_config.get('threshold', 0.05)  # degrees
 
     try:
         pocs.say("Taking pointing picture.")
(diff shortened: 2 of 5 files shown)


pocs/state/states/default/pointing.py
```diff
@@ -65,8 +72,9 @@ def on_enter(event_data):
                 sleep(wait_interval)
                 wait_time += wait_interval
 
+            # Analyze pointing
             if pocs.observatory.current_observation is not None:
```
jamessynge: Why is this not `if observation is not None:`

conf_files/pocs.yaml
```diff
@@ -37,8 +37,9 @@ mount:
         baudrate: 9600
     non_sidereal_available: True
 pointing:
-    threshold: 0.05
-    exptime: 30
+    auto_correct: True
```
jamessynge: If `auto_correct: False`, what benefit would we get? I think that logging in that case should at least be INFO level.

pocs/state/states/default/pointing.py
```diff
@@ -49,6 +55,7 @@ def on_enter(event_data):
                     except Exception as e:
                         pocs.logger.error("Problem waiting for images: {}".format(e))
 
+            # Wait for images to complete
```
jamessynge: We definitely need to put together a helper for this kind of loop, shared with observing.py (or whatever the name is... the one I recently changed).

pocs/state/states/default/pointing.py
```diff
@@ -82,6 +90,31 @@ def on_enter(event_data):
...
+                # Correct the pointing
+                if should_correct and separation > pointing_threshold:
+                    pocs.say("I'm still a bit away from the field so I'm going to get closer.")
+
+                    # Tell the mount we are at the field, which is the center
+                    pocs.say("Syncing with the latest image...")
+                    has_field = pocs.observatory.mount.set_target_coordinates(pointing_image.pointing)
+                    pocs.logger.debug("Coords set, calibrating")
+
+                    # Sync the mount
```
jamessynge: Definitely need a comment that explains what it means to sync the mount.

wtgee: Hoping to get that answer from iOptron. :) I'll add a basic comment.

conf_files/pocs.yaml
jamessynge: Now that I think about it, is `auto_correct: False` supposed to completely suppress the pointing state?

pocs/state/states/default/pointing.py
```diff
@@ -18,6 +16,12 @@ def on_enter(event_data):
 
     pocs.next_state = 'parking'
 
+    # Get pointing parameters
+    pointing_config = pocs.config['pointing']
+    num_pointing_images = pointing_config.get('max_iterations', 3)
+    should_correct = pointing_config.get('auto_correct', False)
```
jamessynge: Seems like False is the wrong default.

joshwalawender:
On other mounts, there are two similar options called "sync" and something else like "update pointing" or something like that.  They are different and we should be careful about whether we are using sync correctly.  On other mounts I've used, you want to "sync" once per night (i.e. on the first target), then use the other option for subsequent pointing updates.

This is because sync applies offsets to the internal pointing model, while the update pointing adds points to the internal model.  The iOptron may not do this, but I wanted to put this here as a caution, so we are sure we understand the correct strategy.

conf_files/pocs.yaml
wtgee: It suppresses the pointing correction, so an initial 30sec image will be taken but the actual `calibrate_mount` command will never be called. We use a pointing image for the WCS tracking correction (and a few other things).

pocs/state/states/default/pointing.py
wtgee: In the end I might change the default in the config to be False, it sort of depends on what we find out from iOptron about usage of the command. I kept False as default here until we confirm as it's easier to change in the config.

wtgee:
> This is because sync applies offsets to the internal pointing model, while the update pointing adds points to the internal model. The iOptron may not do this, but I wanted to put this here as a caution, so we are sure we understand the correct strategy.

Agreed, and that's why we originally removed this code. We were having the tracking issues on PAN001 and we weren't sure if it was the hardware or software. 

I've changed the default to False so that this PR doesn't change any behavior. I've been using it on PAN001 and PAN008 and it's nice to actually have target in center.

wtgee:
I've been using this on PAN001 and it seems to be working well as coded. Defaults make this disabled unless turned on, so I think we would be okay to merge. Not sure it's worth waiting to hear from iOptron on specific answers.

pocs/state/states/default/pointing.py
jamessynge: My point was that, IIUC, the old behavior was equivalent to should_correct==True, so the default should be True.

wtgee: The very old behavior was equal to that but that has not how we have been operating for about 2 years.

To be clear, we always do a pointing image at the beginning of the observation sequence. This option is making it do a "sync" after the initial pointing image, reslew to intended coordinates, then perform another pointing image. It can do this in a loop (`max_iterations`) until it reaches the `threshold`. Usually that's just one additional pointing image.

PR #581, opened 2018-09-06
Fix pretty image linking

wtgee:
If the `images.jpg` file doesn't already exist the currently logic skips the linking. Fix for that.

Also, note the use of `suppress` for the expected exception. Learned this from the Huntsman folks and have started using it in the case where appropriate (such as here). I think we should encourage the same throughout project (where appropriate).

### DIFF
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -3,10 +3,11 @@ import subprocess
 import shutil
 import re
 
+from contextlib import suppress
+
 from matplotlib import pyplot as plt
 from warnings import warn
 
-from astropy import units as u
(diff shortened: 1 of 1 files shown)


jamessynge:
I'm curious why you withdrew this. Since my unit filled its disk, and I wiped out LOTS of stuff from $PANDIR/images, the state machine and latest observation images are no longer linked. Would this have addressed that?

wtgee:
That's exactly what happened to me and what I was attempting to fix but it didn't work. Looking at what I did it was mostly cleanup but not a fix. For some reason a stale symlink seems to give it problems, need to play with it a little more.

PR #582, opened 2018-09-06
Support both old and new Google Cloud libraries

jamessynge:
It appears that the Python package gcloud has been
replaced by google.cloud. This change allows for both
packages to be used until we update all units.


wtgee:
This is changed in #577 as well. Are files not uploading right now?

jamessynge:
It isn't that files weren't uploading. This was triggered by my PR to support installing on 18.04 and using conda more. Conda doesn't have the older gcloud library.

PR #583, opened 2018-09-08
Update mongodb install to support Ubuntu 18.04

jamessynge:
This is extracted from #578.


jamessynge:
Yes, we should definitely worry about it. I just haven't given it enough thought yet.

PR #585, opened 2018-09-09
Fix install-dependencies.sh for Ubuntu 18.04

jamessynge:
- Update for conda 4.4
- Avoid astroscrappy build problem by requiring Python 3.6
- Make the install process quieter by logging the output of
  the astrometry make commands, instead flooding the terminal
  with content most folks won't care about.
- Add some more context messages to the output.
- Create an logs/install directory specific to this install.- Move all of the command line arg handling to the start of the
file.
- Add comments to most functions.
- Apply most fixes suggested by shellcheck, e.g. quoting
  variable expansions and replacing fgrep with grep -F.
- Insert changes into the appropriate file used by the shell
  at login.

### DIFF
--- a/scripts/install/install-dependencies.sh
+++ b/scripts/install/install-dependencies.sh
@@ -536,12 +536,11 @@ function checksum_panoptes_env() {
 # Install latest version of Miniconda (Anaconda with very few packages; any that
 # are needed can then be installed).
 function install_conda() {
-  local -r the_destination="${CONDA_INSTALL_DIR}"
-  if [[ -d "${the_destination}" ]] ; then
+  if [[ -d "${CONDA_INSTALL_DIR}" ]] ; then
     echo_bar
     echo
-    echo "Removing previous miniconda installation from ${the_destination}"
-    rm -rf "${the_destination}"
+    echo "Removing previous miniconda installation from ${CONDA_INSTALL_DIR}"
+    rm -rf "${CONDA_INSTALL_DIR}"
   fi
 
   echo_bar
@@ -550,17 +549,17 @@ function install_conda() {
   local -r the_script="${PANDIR}/tmp/miniconda.sh"
   wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh \
        -O "${the_script}"
-  bash "${the_script}" -b -p "${the_destination}"
+  bash "${the_script}" -b -p "${CONDA_INSTALL_DIR}"
   rm "${the_script}"
 
   # As per the Anaconda 4.4 release notes, one is supposed to add the following
   # to .bash_profile, .bashrc or wherever is appropriate:
-  #    . $CONDA_LOCATION/etc/profile.d/conda.sh
+  #    . $CONDA_INSTALL_DIR/etc/profile.d/conda.sh
   #    conda activate <name-of-desired-environment>
-  # Where CONDA_LOCATION is where Anaconda or miniconda was installed.
+  # Where CONDA_INSTALL_DIR is where Anaconda or miniconda was installed.
   # We do the first step here. Later we'll activate the PANOPTES environment.
 
-  . "${the_destination}/etc/profile.d/conda.sh"
+  . "${CONDA_SH}"
 }
 
 # Add additional repositories in which conda should search for packages.
@@ -572,6 +571,7 @@ function add_conda_channels() {
   # package. And by default the most recently added repository is treated
   # as the highest priority repository. Here we use prepend to be clear
   # that we want astropy to be highest priority.
+  . "${CONDA_SH}"
   if (conda config --show channels | grep -F -q astropy) ; then
     conda config --remove channels astropy
   fi
@@ -588,6 +588,7 @@ function add_conda_channels() {
 function prepare_panoptes_conda_env() {
   # Use the base Anaconda environment until we're ready to
   # work with the PANOPTES environment.
+  . "${CONDA_SH}"
   conda activate base
 
   # Determine if the PANOPTES environment already exists.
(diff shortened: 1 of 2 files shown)


jamessynge:
This script doesn't quite work right w.r.t. detecting the presence of conda due to https://github.com/conda/conda/issues/7753. I'll workaround that by checking for the environment variables that they do export.

jamessynge:
After a ton of experiments and commits, this is ready for review. @wtgee Could you handle that?

jamessynge:
Oh, note that it is currently configured to install the "latest" version of astrometry (i.e. not stable). Let me know if you want me to return it to 0.72 or to the current latest, 0.76.

scripts/install/install-dependencies.sh
```diff
@@ -253,223 +587,288 @@ function install_latest_cfitsio() {
...
+  echo_bar
+  echo
+  echo "Unpacking ${FN}."
+  tar zxf "${FN}"
+
+  if [[ "${ASTROMETRY_VERSION}" == "latest" ]] ; then
+    # We need to know the version that is in the tar file in order to
+    # enter that directory.
+    ASTROMETRY_VERSION=""
+    for version in $(find astrometry.net-* -maxdepth 0 -type d | sed "s/astrometry.net-//")
```
wtgee: Can't we just rename the directory when we untar or wget it?

scripts/install/install-dependencies.sh
```diff
@@ -253,223 +587,288 @@ function install_latest_cfitsio() {
...
+  local -r make_py_log="${INSTALL_LOGS_DIR}/make-astrometry-py.log"
+  local -r make_install_log="${INSTALL_LOGS_DIR}/make-astrometry-install.log"
   cd "${SCRATCH_DIR}"
 
   echo_bar
-  echo  
+  echo
   echo "Fetching astrometry into directory $(pwd)"
-
   wget "${URL}"
```
wtgee: Rename dir here with `-O` flag. Comment below.

scripts/install/install-dependencies.sh
```diff
@@ -253,223 +587,288 @@ function install_latest_cfitsio() {
...
+    install_astrometry
+  fi
 }
 
+# Run a POCS script for installing the indices used by astrometry
+# to solve fields.
 function install_astrometry_indices() {
+  echo_bar
+  echo
+  echo "Fetching astrometry and other indices."
```
wtgee: Don't we still somewhere need to tell the astrometry.cfg file about the data dir? Or does it default to `data` inside it's base dir?

scripts/install/install-dependencies.sh
```diff
@@ -253,223 +587,288 @@ function install_latest_cfitsio() {
...
 
+# Install packages using the APT package manager. Works on Debian based systems
+# such as Ubuntu.
 if [[ "${DO_APT_GET}" -eq 1 ]] ; then
   install_apt_packages
 fi
 
-
+# Install Mongodb, a document database. Used by POCS for storing environment
+# readings, from which weather plots are generated.
```
wtgee: Also stores observations metadata. Not really important but comment undersells how much we use it to some extent.

scripts/install/install-dependencies.sh
```diff
@@ -253,223 +587,288 @@ function install_latest_cfitsio() {
...
-  # an install with just about two commands (wget script, then run script).
-  pip install -r "${POCS}/requirements.txt"
+  echo_bar
+  echo
+  echo "Installing python packages using pip."
+  pip install --quiet --requirement "${POCS}/requirements.txt"
 fi
 
-
+# Install cfitsio, native tools for reading and writing FITS files.
```
wtgee: Did cfitsio move to a conda package and if so does it need a separate option for installing? Would we ever not install cfitsio?

wtgee:
What files does this depend on?  Ideally I want to wget this file and run it and do nothing else but I guess it's not quite to that point.

Pulled down the file. Env variables weren't set. Pulled down the `default-env-vars.sh`. Running it doesn't actually set my env vars, so I have to manually do that before I can run script. I know we want to clean up env vars otherwise but seems like install script should set them up if they don't exist.

wtgee:
Because some of this runs as sudo it requires that the env variables be set up for the sudo user. Not sure if that is intended or needed.

```bash
panoptes@pan008-nuc:~/Downloads$ head install-dependencies.sh 
#!/bin/bash -e

# Run with --help to see your options. With no options, does a complete install
# of dependencies, though attempts to reuse existing installs.

THIS_DIR="$(dirname "$(readlink -f "${0}")")"
THIS_PROGRAM="$(basename "${0}")"

if [[ -z "${PANDIR}" || -z "${POCS}" || -z "${PAWS}" || -z "${PANLOG}" ||
      -z "${PANUSER}" ]] ; then

panoptes@pan008-nuc:~/Downloads$ echo $PANDIR
/var/panoptes

panoptes@pan008-nuc:~/Downloads$ echo $POCS
/var/panoptes/POCS

panoptes@pan008-nuc:~/Downloads$ echo $PAWS
/var/panoptes/PAWS

panoptes@pan008-nuc:~/Downloads$ echo $PANLOG
/var/panoptes/logs

panoptes@pan008-nuc:~/Downloads$ echo $PANUSER
panoptes

panoptes@pan008-nuc:~/Downloads$ ./install-dependencies.sh 
mkdir: cannot create directory ‘tmp’: Permission denied

panoptes@pan008-nuc:~/Downloads$ sudo ./install-dependencies.sh 
Please set the Panoptes environment variables, then re-run this script.

panoptes@pan008-nuc:~/Downloads$ 
```

jamessynge:
Our [Intel NUC Setup](https://docs.google.com/document/d/16Z87vPiAsEx9yAN5EwsPXn6x5PMAZWMvJz0p2aF6cE8/edit#heading=h.yfyg7tyy1g1h) doc describes how to run this, including how to setup the environment variables.

My intent though is to eventually have a "setup computer" script, that is used right after creating the first user (e.g. "james") on the computer, and takes care of creating the "panoptes" user, sets up the cron jobs, etc., leaving you with a system that is ready to go... though we'd need some second action that actually enables POCS to run automatically. This new script would certainly be derived from this one.

wtgee:
Tried latest version while following docs and worked well.

One note, the install script gives instructions to follow at the end, which I did. This is slightly different than what the docs say. I made a comment in the docs.

PR #588, opened 2018-09-11
Make sure to use existing observation

wtgee:
When selecting the same observation make sure to use the actual same observation object. The new observation object, while the same in principle, does not have a pointing image attached to it, which is used in the tracking corrections.

### DIFF
--- a/pocs/state/states/default/scheduling.py
+++ b/pocs/state/states/default/scheduling.py
@@ -32,6 +32,9 @@ def on_enter(event_data):
 
             if existing_observation and observation.name == existing_observation.name:
                 pocs.say("I'm sticking with {}".format(observation.name))
+
+                # Make sure we are using existing observation (with pointing image)
+                pocs.observatory.current_observation = existing_observation
                 pocs.next_state = 'tracking'
             else:
                 pocs.say("Got it! I'm going to check out: {}".format(observation.name))


pocs/state/states/default/scheduling.py
```diff
@@ -32,6 +32,9 @@ def on_enter(event_data):
 
             if existing_observation and observation.name == existing_observation.name:
                 pocs.say("I'm sticking with {}".format(observation.name))
+
+                # Make sure we are using existing observation (with pointing image)
+                pocs.observatory.current_observation = existing_observation
```
jamessynge: Why would pocs.observatory.current_observation have changed since we earlier got its value? Does pocs.observatory.get_observation() cause a change?

wtgee: get_observation does cause a change because the observation file could have been reloaded. That said, I'm not sure this change is the best approach.

It's somewhat a question of should we allow the observation currently in progress to have its parameters (such as exposure time) changed. My feeling would be no, in which case the above is okay. There is then a question of reloading of the targets file during operation and what to do if there is a target with the same name but different parameters from the observation that is currently in progress. Mostly just need to be careful about being explicit to the user about what will happen.

Going to think about this a little more.

wtgee: Ok, spent a little time with this. If the config item `scheduler.check_file` is True, (it is False by default) then `read_field_file` is called before each call to `get_observation`. This effectively rebuilds the `Observations`, which includes a change to the `seq_time`. That means without the above line the images would be placed in a different folder and effectively be a different observation.

The only issue that would arise from the change above is if you wanted to dynamically change the current observation. For example, if you were observing for 120 seconds and had taken 60 pictures but wanted to suddenly switch to 60 seconds while still staying on the same observation, i.e. no slewing, all images in same folder.

I think we can safely not support that feature for now as it is more important that, for example, images 60-70 go into the same folder as 1-60 even if we have re-read the targets file.

wtgee: Pointer back to here in #615

PR #589, opened 2018-09-11
Camera updates

wtgee:
Ideally the take_pic.sh script will be re-written in python to handle any pictures (e.g. bias and regular) but this is some small cleanup before that time.

* Cleaning up script
* Adding default time for cameras - #503 
* Caption is optional on jpgs (it's a slow operation)
* No resizing of jpg

Note that this was also waiting to get some of the camera hardware testing infrastructure in place but I wanted to make sure to get these updates out.

### DIFF
--- a/pocs/camera/canon_gphoto2.py
+++ b/pocs/camera/canon_gphoto2.py
@@ -35,26 +35,31 @@ class Camera(AbstractGPhotoCamera):
 
         # Properties to be set upon init.
         prop2index = {
+            '/main/actions/viewfinder': 1,                # Screen off
+            '/main/capturesettings/autoexposuremode': 3,  # 3 - Manual; 4 - Bulb
+            '/main/capturesettings/continuousaf': 0,      # No auto-focus
+            '/main/capturesettings/drivemode': 0,         # Single exposure
+            '/main/capturesettings/focusmode': 0,         # Manual (don't try to focus)
+            '/main/capturesettings/shutterspeed': 0,      # Bulb
+            '/main/imgsettings/imageformat': 9,           # RAW
+            '/main/imgsettings/imageformatcf': 9,         # RAW
+            '/main/imgsettings/imageformatsd': 9,         # RAW
+            '/main/imgsettings/iso': 1,                   # ISO 100
+            '/main/settings/autopoweroff': 0,             # Don't power off
+            '/main/settings/capturetarget': 0,            # Capture to RAM, for download
             '/main/settings/datetime': 'now',             # Current datetime
             '/main/settings/datetimeutc': 'now',          # Current datetime
-            '/main/actions/viewfinder': 1,                # Screen off
-            '/main/settings/autopoweroff': 0,             # Don't power off
             '/main/settings/reviewtime': 0,               # Screen off after taking pictures
-            '/main/settings/capturetarget': 0,            # Capture to RAM, for download
-            '/main/imgsettings/imageformat': 9,           # RAW
-            '/main/imgsettings/imageformatsd': 9,         # RAW
-            '/main/imgsettings/imageformatcf': 9,         # RAW
-            '/main/imgsettings/iso': 1,                   # ISO 100
-            '/main/capturesettings/focusmode': 0,         # Manual (don't try to focus)
-            '/main/capturesettings/continuousaf': 0,      # No auto-focus
-            '/main/capturesettings/autoexposuremode': 3,  # 3 - Manual; 4 - Bulb
-            '/main/capturesettings/drivemode': 0,         # Single exposure
-            '/main/capturesettings/shutterspeed': 0,      # Bulb
         }
+
+        owner_name = 'Project PANOPTES'
+        artist_name = self.config.get('unit_id', owner_name)
+        copyright = 'owner_name {}'.format(owner_name, current_time().datetime.year)
+
         prop2value = {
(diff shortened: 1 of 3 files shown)


scripts/cr2_to_jpg.sh
```diff
@@ -1,10 +1,37 @@
...
+}
+
+if [ $# -eq 0 ]; then
+    usage
+    exit 1
+fi
 
 FNAME=$1
-NAME=$2
+CAPTION=$2
```
jamessynge: Please consider running shellcheck on this and following much of its advice. Here it would probably recommend quoting $2. I would recommend:  `CAPTION="${2}"`

pocs/camera/canon_gphoto2.py
```diff
@@ -35,6 +35,8 @@ def connect(self):
 
         # Properties to be set upon init.
         prop2index = {
+            '/main/settings/datetime': 'now',             # Current datetime
```
jamessynge: What do you think about sorting these lines lexically?

scripts/cr2_to_jpg.sh
```diff
@@ -1,10 +1,37 @@
...
 
 JPG=${FNAME/cr2/jpg}
```
jamessynge: How about: `JPG="${FNAME%.cr2}.jpg"`
This will only replace `.cr2` at the end of the string, not in the middle.

scripts/cr2_to_jpg.sh
```diff
@@ -1,10 +1,37 @@
...
 
+echo "Converting ${FNAME} to jpg."
```
jamessynge: jpg was probably supposed to be `${JPG}`.

scripts/cr2_to_jpg.sh
```diff
@@ -18,6 +45,12 @@ else
     fi
 fi
 
-# Make thumbnail from jpg.
-convert ${JPG} -thumbnail 1280x1024 -background black -fill red \
-    -font ubuntu -pointsize 24 label:"${NAME}" -gravity South -append ${JPG}
+if [[ $CAPTION ]]
```
jamessynge: How about: `if [[ -n "${CAPTION}" ]]`

scripts/take_pic.sh
```diff
@@ -1,14 +1,45 @@
...
+echo "PORT = ${PORT}"
+echo "TIME = ${TIME}s"
+echo "FILE = ${FILENAME}"
 
 # Open shutter
-gphoto2 --port=${P} --set-config eosremoterelease=Immediate \
-         --wait-event=${T}s --set-config eosremoterelease=4 --wait-event-and-download=2s \
-         --filename "${F}"
+gphoto2 --port=${PORT} \
+        --set-config shutterspeed=0 \
```
jamessynge: Could you add comments at the end of the lines where we have these magic numbers and symbols?

scripts/take_pic.sh
```diff
@@ -1,14 +1,45 @@
...
+# Take a picture via gphoto2.
+# 
+# This will change the camera setting to bulb and then take
+# an exposure for the requested amount of time.
+##################################################
+ $ $(basename $0) PORT EXPTIME OUTFILE
+ 
+ Options:
+  PORT              USB port as reported by gphoto2 --auto-detect, e.g. usb:001,004.
+  EXPTIME           Exposure time in seconds, should be greater than 1 second.
```
jamessynge: Must this be an integer?

scripts/take_pic.sh
```diff
@@ -1,14 +1,45 @@
...
+  OUTFILE           Output filename (with .cr2 extension).
```
jamessynge: With "appropriate" extension; e.g. .cr2 for Canon RAW images.

scripts/take_pic.sh
```diff
@@ -1,14 +1,45 @@
-#!/bin/bash
+#!/bin/bash -e
 
-P=$1
-T=$2
-F=$3
+usage() {
+  echo -n "##################################################
+# Take a picture via gphoto2.
```
jamessynge: Maybe a comment that this seems limited to working with Canon EOS images.

scripts/cr2_to_jpg.sh
wtgee: I didn't even know of shellcheck! I don't do much actual bash scripting. Thanks!

pocs/camera/canon_gphoto2.py
wtgee: Done. Not sure it makes it more clear but a good idea regardless.

scripts/cr2_to_jpg.sh
wtgee: I thought it might be too long to and redundant to have the full path for both since we don't offer the possibility of a different name or output folder but since that's not explicit it's probably best to clarify.

wtgee: Actually, I changed it to "Converting CR2 to ${JPG}".

scripts/take_pic.sh
wtgee: Do you have a better syntax than this?

```bash
# Open shutter
gphoto2 --port=${PORT} \
        --set-config shutterspeed=0 `#Always set to bulb` \
        --set-config capturetarget=0 `#Capture to RAM for download` \
        --set-config eosremoterelease=Immediate \
```

PR #590, opened 2018-09-12
Parking cleanups

wtgee:
* Cleanup to `parked` state to make it easier to understand
* Ctrl-c while pocs_shell is running will finish loop through state
machine rather than call park directly on the mount. This allows for
housekeeping (including uploading) of observations even when machine
is manually killed.

Edit: ~~Note: in future might be worth adding an option such that when you
press Ctrl-c you are prompted for if you want immediate park and shutdown
or if should run through states.~~ I think I would prefer to move to #592

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -193,9 +193,12 @@ Hardware names: {}   (or all for all hardware)'''.format(
             try:
                 self.pocs.run()
             except KeyboardInterrupt:
-                print_warning('POCS interrupted, skipping states and parking')
-                self.pocs.observatory.mount.home_and_park()
-                self._running = False
+                print_warning('POCS interrupted, parking')
+                if self.pocs.state not in ['sleeping', 'housekeeping', 'parked', 'parking']:
+                    self.pocs.park()
+                else:
+                    self.pocs.observatory.mount.home_and_park()
+                self._obs_run_retries = 0  # Don't retry
             finally:
                 print_info('POCS stopped.')
         else:
--- a/pocs/state/states/default/parked.py
+++ b/pocs/state/states/default/parked.py
@@ -4,24 +4,26 @@ def on_enter(event_data):
     pocs = event_data.model
     pocs.say("I'm parked now. Phew.")
 
-    has_valid_observations = pocs.observatory.scheduler.has_valid_observations
+    if pocs.run_once is True:
+        pocs.say("Done running loop, going to clean up and sleep!")
+        pocs.next_state = 'housekeeping'
+    elif pocs.should_retry is False:
+        pocs.say("Done with retrying loop, going to clean up and sleep!")
+        pocs.next_state = 'housekeeping'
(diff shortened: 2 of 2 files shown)


bin/pocs_shell
```diff
@@ -193,9 +193,12 @@ Hardware names: {}   (or all for all hardware)'''.format(
             try:
                 self.pocs.run()
             except KeyboardInterrupt:
-                print_warning('POCS interrupted, skipping states and parking')
-                self.pocs.observatory.mount.home_and_park()
-                self._running = False
+                print_warning('POCS interrupted, parking')
```
jamessynge: Perhaps add the TODO here for the behavior you suggested in the PR?

pocs/state/states/default/parked.py
```diff
@@ -4,24 +4,26 @@ def on_enter(event_data):
     pocs = event_data.model
     pocs.say("I'm parked now. Phew.")
```
jamessynge: Let's drop the Phew.

pocs/state/states/default/parked.py
```diff
@@ -4,24 +4,26 @@ def on_enter(event_data):
...
 
-    has_valid_observations = pocs.observatory.scheduler.has_valid_observations
+    if pocs.run_once is True:
+        pocs.say("Done running loop, going to clean up and sleep!")
+        pocs.next_state = 'housekeeping'
+    elif pocs.should_retry is False:
+        pocs.say("Done with retrying loop, going to clean up and sleep!")
+        pocs.next_state = 'housekeeping'
+    else:
+        has_valid_observations = pocs.observatory.scheduler.has_valid_observations
```
jamessynge: Looks like this var is only used in one place.

pocs/state/states/default/parked.py
```diff
@@ -4,24 +4,26 @@ def on_enter(event_data):
...
 
-    if has_valid_observations:
-        if pocs.is_safe():
-            if pocs.should_retry is False or pocs.run_once is True:
-                pocs.say("Done retrying for this run, going to clean up and shut down!")
-                pocs.next_state = 'housekeeping'
-            else:  # This branch will only happen if there is an error causing a shutdown
+        if has_valid_observations:
```
jamessynge: I take it that `has_valid_observations` being True means that those observations could be taken now, so it isn't yet Astronomical dawn, or some such.

bin/pocs_shell
wtgee: Would prefer to move to #592 actually. Can make a note.

pocs/state/states/default/parked.py
wtgee: It doesn't capture the logic of whether those observations are observable or not, it just captures whether or not there is even something to try to schedule. This mostly comes into play if targets are removed while running. Not a widely used feature right now in terms of actively adding/removing fields while running.

PR #591, opened 2018-09-13
Timelapse update

wtgee:
* Removing ffmpy requirement
* Can overwrite timelapse
* Default timelapse goes into same directory as observation

### DIFF
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -223,6 +223,10 @@ def make_timelapse(
 
     Returns:
         str: Name of output file
+
+    Raises:
+        error.InvalidSystemCommand: Raised if ffmpeg command is not found.
+        FileExistsError: Raised if fn_out already exists and overwrite=False.
     """
     if fn_out is None:
         head, tail = os.path.split(directory)
@@ -241,6 +245,9 @@ def make_timelapse(
         raise FileExistsError("Timelapse exists. Set overwrite=True if needed")
 
     ffmpeg = shutil.which('ffmpeg')
+    if ffmpeg is None:
+        raise error.InvalidSystemCommand("ffmpeg not found, can't make timelapse")
+
     inputs_glob = os.path.join(directory, '*.{}'.format(file_type))
 
     try:
@@ -334,9 +341,12 @@ def clean_observation_dir(dir_name, *args, **kwargs):
         if len(jpg_list) > 0:
 
             # Create timelapse
-            _print('Creating timelapse for {}'.format(dir_name))
-            video_file = make_timelapse(dir_name)
-            _print('Timelapse created: {}'.format(video_file))
+            try:
+                _print('Creating timelapse for {}'.format(dir_name))
+                video_file = make_timelapse(dir_name)
+                _print('Timelapse created: {}'.format(video_file))
+            except Exception as e:
+                _print("Problem creating timelapse: {}".format(e))
 
             # Remove jpgs
             _print('Removing jpgs')
--- a/pocs/utils/images/__init__.py
+++ b/pocs/utils/images/__init__.py
@@ -209,10 +209,10 @@ def make_timelapse(
         **kwargs):
     """Create a timelapse.
 
-    A timelapse is created from all the jpg images in a given `directory`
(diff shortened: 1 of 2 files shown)


pocs/utils/images/__init__.py
```diff
@@ -211,6 +216,9 @@ def create_timelapse(directory, fn_out=None, file_type='jpg', **kwargs):
         fn_out (str, optional): Full path to output file name, if not provided,
             defaults to `directory` basename.
         file_type (str, optional): Type of file to search for, default 'jpg'.
+        overwrite (bool, optional): Overwrite timelapse if exists, default False.
+        timeout (int): Timeout for making movie, default 60 seconds.
```
jamessynge: I'm curious, how long does this take on the RPi?

pocs/utils/images/__init__.py
```diff
@@ -224,29 +232,48 @@ def create_timelapse(directory, fn_out=None, file_type='jpg', **kwargs):
...
 
-    fn_dir = os.path.dirname(fn_out)
-    os.makedirs(fn_dir, exist_ok=True)
+    if verbose:
+        print("Timelapse file: {}".format(fn_out))
+
+    if os.path.exists(fn_out) and not overwrite:
+        raise FileExistsError("Timelapse exists. Set overwrite=True if needed")
+
+    ffmpeg = shutil.which('ffmpeg')
```
jamessynge: Check for no ffmpeg?

pocs/utils/images/__init__.py
```diff
@@ -201,7 +199,14 @@ def _make_pretty_from_cr2(fname, timeout=15, **kwargs):  # pragma: no cover
...
+        directory,
+        fn_out=None,
+        file_type='jpg',
+        overwrite=False,
+        timeout=60,
+        verbose=False,
+        **kwargs):
     """Create a timelapse
 
     A timelapse is created from all the jpg images in a given `directory`
```
jamessynge: jpg is the default type of file, but we allow for others. What other types would ffmpeg support? Probably gif and png. Speaking of which, I've wondered about using png in place of jpg to avoid lossy artifacts.

pocs/utils/images/__init__.py
```diff
@@ -224,29 +232,48 @@ def create_timelapse(directory, fn_out=None, file_type='jpg', **kwargs):
...
-    fn_dir = os.path.dirname(fn_out)
-    os.makedirs(fn_dir, exist_ok=True)
+    if verbose:
+        print("Timelapse file: {}".format(fn_out))
+
+    if os.path.exists(fn_out) and not overwrite:
+        raise FileExistsError("Timelapse exists. Set overwrite=True if needed")
+
+    ffmpeg = shutil.which('ffmpeg')
     inputs_glob = os.path.join(directory, '*.{}'.format(file_type))
```
jamessynge: What is it that orders the files added to the video? Just the lexical order?

pocs/utils/images/__init__.py
```diff
@@ -224,29 +232,48 @@ def create_timelapse(directory, fn_out=None, file_type='jpg', **kwargs):
...
+            # Don't wait forever
+            outs, errs = proc.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.kill()
+            outs, errs = proc.communicate()
+        finally:
+            if verbose:
+                print(outs)
+                print(errs)
+            fn_out = None
```
jamessynge: I suspect this belongs in the except block above, along with the `proc.kill()`. Otherwise fn_out is always None when returning.

pocs/utils/images/__init__.py
wtgee: jpgs come directly out of the CR2 without us having to create/convert anything so it's the easiest. Would be open to another solution if practical.

wtgee: Updated docs to not be jpg specific.

pocs/utils/images/__init__.py
wtgee: I think ffmpeg is intelligent about ordering them but worth checking on. I usually sort after the glob but here we're just passing the glob.

Issue #205, opened 2017-12-18
PanMessaging.send_message doesn't ensure channel name has no space in it.


jamessynge:
PanMessaging.receive_message requires that it be able to split the received message into a channel name and a json encoded message body.

We should change the top-level encoding used to be a json encoded list of two elements, the channel name and the dict containing the structured message.

wtgee:
@jamessynge, you did some recent work with the messaging. Could you speak to any update on this Issue?

wtgee:
Ping @jamessynge

jamessynge:
Not done. Minimal change would of course be to just throw an exception if there is a space in the name.

wtgee:
Was closed by #593.

PR #593, opened 2018-09-15
Ensure that channel names don't have spaces

jamessynge:
Throws an exception if a channel name doesn't match
a regular expression, one which doesn't include spaces.
Fixes issue #205.

Enforce that the value of a message is either a string
or a dict (that was already effectively required, but
now the failure message will be clear).

Greatly expand the class level documentation in PanMessaging.

### DIFF
--- a/pocs/utils/messaging.py
+++ b/pocs/utils/messaging.py
@@ -184,15 +184,13 @@ class PanMessaging(object):
         if not isinstance(channel, str):
             raise ValueError('Channel name must be a string')
         elif not self.name_re.fullmatch(channel):
-            raise ValueError('Channel name ("{}") is not valid'
-                .format(channel))
+            raise ValueError('Channel name ("{}") is not valid'.format(channel))
 
         if isinstance(message, str):
             message = {
                 'message': message,
-                'timestamp': current_time().isot.replace(
-                    'T',
-                    ' ').split('.')[0]}
+                'timestamp': current_time().isot.replace('T', ' ').split('.')[0]
+            }
         elif isinstance(message, dict):
             message = self.scrub_message(message)
         else:
--- a/pocs/utils/messaging.py
+++ b/pocs/utils/messaging.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 import zmq
 
 import yaml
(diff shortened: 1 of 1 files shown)


pocs/utils/messaging.py
```diff
@@ -14,14 +15,79 @@
...
+    an instance of PanMessaging, on which they can then call
+    receive_message.
+
+    Messages are sent to channels, a name that can be used to allow
+    a high-level partitioning of messages. A channel name may not
+    include whitespace. Among the currently used channel names are:
+
+      * PANCHAT (sent from POCS.say)
+      * PAWS-CMD (sent from PAWS websockets.py)
+      * POCS (sent by class POCS)
```
wtgee: I don't recall what this is used for.

pocs/utils/messaging.py
```diff
@@ -14,14 +15,79 @@
...
+    PanMessaging, on which they can then call send_message.
+    Subscribers should call PanMessaging.create_subscriber to create
+    an instance of PanMessaging, on which they can then call
+    receive_message.
+
+    Messages are sent to channels, a name that can be used to allow
+    a high-level partitioning of messages. A channel name may not
+    include whitespace. Among the currently used channel names are:
+
+      * PANCHAT (sent from POCS.say)
```
wtgee: Seeing this makes me think there are some `PAN-CHAT` out there as well. Should double-check. Thanks for list.

pocs/utils/messaging.py
```diff
@@ -14,14 +15,79 @@
...
+      * telemetry:commands (in ArduinoIO... new)
+      * camera:commands (in ArduinoIO... new)
+
+    And some other channels are used in tests:
+
+      * TEST-CHANNEL (test_messaging.py)
+      * RUNNING (test_pocs.py)
+      * POCS-CMD (test_pocs.py)
+
+    The method receive_message will return messages from all channels;
```
wtgee: While this is true, the intended usage is to pass the channel to the `create_subcriber` and then no filtering is needed by the user. This should be the recommended behaviour.

pocs/utils/messaging.py
```diff
@@ -14,14 +15,79 @@
...
+    the caller must check the returned channel name to determine if
+    the message value is of interest.
+
+    Note: PAWS doesn't use PanMessaging, which will likely result in
+    problems as we evolve PanMessaging and the set of channels.
+    TODO: Figure out how to share PanMessaging with PAWS.
```
wtgee: I think the answer is that we will be changing PAWS.

pocs/utils/messaging.py
```diff
@@ -111,20 +177,24 @@ def send_message(self, channel, message):
...
+            raise ValueError('Channel name ("{}") is not valid'.format(channel))
 
         if isinstance(message, str):
             message = {
                 'message': message,
-                'timestamp': current_time().isot.replace(
-                    'T',
-                    ' ').split('.')[0]}
-        else:
+                'timestamp': current_time().isot.replace('T', ' ').split('.')[0]
```
wtgee: `current_time(pretty=True)` is the same thing, this code just hasn't been changed since then. Can you update please?

pocs/utils/messaging.py
```diff
@@ -188,13 +259,13 @@ def scrub_message(self, message):
             if isinstance(v, Time):
                 v = str(v.isot).split('.')[0].replace('T', ' ')
 
-            # Hmmmm
+            # Hmmmm. What is going on here? We need some documentation.
```
wtgee: Yup. I can't remember what case this was handling. I'll try to find out.

PR #596, opened 2018-09-16
Fix some bugs and lots of warnings reported by pyflakes

jamessynge:
Fixed some "unknown variable" errors, such as port when
it should have been self.port, or the corruption of
nickname by someone pasting the word position into the
middle of it.

Converted some string literals to raw string literals so
that the regular expression is not invalid; e.g 'PAN\d\d\d'
is invalid because \d is not a valid escape sequence. The
author likely meant \d to be the regular expression for
matching digits, which should either be '\\d' or r'\d'.

Removed unused imports. Marked some imports in __init__.py
files with "# pragma: no flakes" so that they can be exported
cleanly.

Replaced "from x import *" with explicit imports in libfli.py.

Removed many unused variables 'e' in statements of
the form: except Name as e.

Fixed uses of pytest.raises in test_social_messaging where
the asserts about the contents of the raised exception were
at the wrong indentation level, so were never executed.

### DIFF
--- a/pocs/camera/sbigudrv.py
+++ b/pocs/camera/sbigudrv.py
@@ -8,20 +8,15 @@ enums and structs defined in the library C-header to Python dicts and
 ctypes.Structures, plus a class (SBIGDriver) to load the library
 and call the single command function (SBIGDriver._send_command()).
 """
-import platform
 import ctypes
 from ctypes.util import find_library
 from warnings import warn
-import _ctypes
-import os
 import time
 from threading import Timer, Lock
 
 import numpy as np
 from numpy.ctypeslib import as_ctypes
 from astropy import units as u
-from astropy.io import fits
-from astropy.time import Time
 
 from pocs.base import PanBase
 from pocs.utils.images import fits as fits_utils
--- a/pocs/utils/google/storage.py
+++ b/pocs/utils/google/storage.py
@@ -39,7 +39,7 @@ class PanStorage(object):
         super(PanStorage, self).__init__()
 
         self.unit_id = load_config()['pan_id']
-        assert re.match('PAN\d\d\d', self.unit_id) is not None
+        assert re.match(r'PAN\d\d\d', self.unit_id) is not None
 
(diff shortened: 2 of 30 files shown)


pocs/camera/sbigudrv.py
```diff
@@ -8,20 +8,15 @@
 ctypes.Structures, plus a class (SBIGDriver) to load the library
 and call the single command function (SBIGDriver._send_command()).
 """
-import platform
 import ctypes
 from ctypes.util import find_library
 from warnings import warn
-import _ctypes
```
wtgee: @AnthonyHorton is there any chance this is doing some magic simply by importing?

pocs/utils/google/storage.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, bucket_name, auth_key=None, project_id='panoptes-survey'):
         super(PanStorage, self).__init__()
 
         self.unit_id = load_config()['pan_id']
-        assert re.match('PAN\d\d\d', self.unit_id) is not None
+        assert re.match(r'PAN\d\d\d', self.unit_id) is not None
```
wtgee: I have these changed in #577 as well but I'll just merge after this.

pocs/utils/google/storage.py
```diff
@@ -51,10 +51,11 @@ def __init__(self, bucket_name, auth_key=None, project_id='panoptes-survey'):
 
         try:
             self.bucket = self.client.get_bucket(bucket_name)
-        except Forbidden as e:
+        except Forbidden:
             raise error.GoogleCloudError(
-                "Storage bucket does not exist or no permissions. " +
-                "Ensure that the PANOPTES_CLOUD_KEY variable is properly set"
+                "Storage bucket does not exist or no permissions. "
+                "Ensure that the PANOPTES_CLOUD_KEY variable is properly set "
```
wtgee: This is removed in #577 as we don't use this env var anymore. Like above, I can either merge after or if you want to remove here.

PR #600, opened 2018-09-17
Update google instructions

wtgee:
One of #599 . Basic instructions for getting authenticated on the google network.

This includes some config items that will be used in forthcoming PRs but should be okay to merge here.

Issue #599, opened 2018-09-16
Observations Cleanup Overview


wtgee:
PR #577 was an update to some of the google network connections to reflect usage on the running units as well as for use with PIAA when using google compute instances. However, that PR grew out of control so it is being split. Items to be accomplished

* [x] Properly document authentication routines. #600
* [x] Update google storage utilities. #601
* [x] Update google cloudsql access. #602
* [x] Run cloudsql proxy on startup. #626 

**Note:** based on review, some of the items have changed since we don't want to manage DB access. Checkboxes below reflect updates (some are not necessarily related to google network connections):

* [ ] Update the morning cleanup routines for each observation (or run-time?). ~~#628~~
  * [x] Remove cloudsql items. #652
  * [x] Update uploads to include manifest log. #654
  * [x] Timelapse and jpg options. #655
  * [x] Update cleanup script. #687 
  * [ ] Consider cleanup of upload scripts. https://github.com/panoptes/POCS/pull/654#discussion_r221798352
  * [x] Use CloudFunction endpoint for updating headers. https://github.com/panoptes/panoptes-network/pull/3
  * [ ] Consider more options for cleanup (timelapse on servers, etc).
  * [x] Make the image upload async. It currently can take a long time to upload images and it's very annoying to have to wait if for some reason you need to wait.

wtgee:
Closing this as (mostly) complete. Additional development or ideas should have new Issues.

PR #601, opened 2018-09-17
Update google storage

wtgee:
Part of #599, depends on #600.

* Storage relies on gcloud authentication (#600) removal of previous env variable
* Methods added:
    * `upload` renamed to `upload_file`
    * `get_file_blob` / `get_file_blobs` - returns blob data
    * `download_file` - get the actual file
    * `lookup_fits_header` stream just the fits header from the storage
* Utility functions:
    * `upload_observation_to_bucket` - uses `gsutil` rather than the underlying gcloud because that allows for easy parallelization.

ToDo: Figure out how to mock tests. Current tests are designed for a live system and actually communicate with the network (requires flag to test).


wtgee:
Mocking suggestions from @jamessynge 

>Does this require "real" cloud storage, or do you have support for a mock somewhere?
> FWIW:

- [Mock Cloud API in gsutil](https://github.com/GoogleCloudPlatform/gsutil/blob/159ea317217dd7a3cb4160b2939edd3f4d09b5b9/gslib/tests/mock_cloud_api.py)
- [A fake GCS server for testing](https://github.com/fsouza/fake-gcs-server) (no idea if any good).

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
...
+def test_bad_bucket(auth_key):
+    with pytest.raises(AssertionError):
+        PanStorage('fake-bucket')
+
+    with pytest.raises(GoogleCloudError):
+        PanStorage('fake-bucket', auth_key=auth_key)
+
+
+@pytest.fixture(scope="function")
+def storage(auth_key):
```
jamessynge: I can't help but think that there should be a way that we can have an auth_key that is not in github for all the world to see, but that can be used by each tester for accessing the test bucket without having to use the same credentials that give access to the primary bucket. Of course, maybe you've already addressed this.

If not, travis-ci does support secret values being passed into tests.

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
+import pytest
```
jamessynge: Could you add some comments please?

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
...
+    os.unlink(temp_fn)
+
+
+def test_delete(storage):
+    """
+    Note: The util wrappers don't provide a way to delete because we generally
+    will not want people to delete things. However it's good to test and we
+    want to remove the above files
+    """
+    remote_path = '{}/pong.txt'.format(storage.unit_id)
```
jamessynge: I'd very much like to avoid depending on tests running serially and in a specific order, unless there is a fixture that is used to ensure that happens. I sometimes run using the xdist or random plugins in order to try find mistakes where our tests are non-hermetic.

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
...
+    assert remote_path == returned_remote_path
+    assert storage.bucket.blob(returned_remote_path).exists()
+    os.unlink(temp_fn)
+
+
+def test_delete(storage):
+    """
+    Note: The util wrappers don't provide a way to delete because we generally
+    will not want people to delete things. However it's good to test and we
+    want to remove the above files
```
jamessynge: Both files aren't removed.

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
...
+    assert re.match(r'PAN\d\d\d', storage.unit_id), storage.logger.error(
+        "Must have valid pan_id. Please change your conf_files/pocs_local.yaml")
+
+
+def test_bucket_exists(storage):
+    assert storage.bucket.exists()
+
+
+def test_file_upload_no_prepend(storage):
+    temp_fn = 'ping.txt'
```
jamessynge: The fixed names don't allow for multiple runs of this test concurrently (e.g. by travis.ci, you, Anthony and me).

pocs/utils/google/storage.py
```diff
@@ -1,40 +1,42 @@
...
             project_id (str, optional): Project id hosting the bucket. Default 'panoptes-survey'
-
         Raises:
             error.GoogleCloudError: Error raised if valid connection cannot be formed for
                 given project, bucket, and authorization.
         """
         self.logger = get_root_logger()
         assert auth_key is not None and os.path.exists(auth_key), self.logger.error(
-            "Cannot use google storage without PANOPTES_CLOUD_KEY variable set.")
+            "Cannot use google storage without a valid auth_key.")
```
jamessynge: Do you want to convert the assert on unit_id to an exception?

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+            if 'pointing' in blob.name and not include_pointing:
+                continue
+            elif filter_ext is not None and blob.name.endswith(filter_ext) is False:
+                continue
+            else:
+                objs.append(blob)
+
+        return sorted(objs, key=lambda x: x.name)
+
+    def download_file(self, remote_path, save_dir='.', force=False, callback=None):
```
jamessynge: Perhaps the default for save_dir should be either $HOME/Downloads or $TMPDIR? The latter is probably better.

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+        """Downloads (and uncompresses) the image blob data.
+
+        Args:
+            remote_path (str|`google.cloud.storage.blob.Blob`): Blob or path to remote blob.
+                If just the blob name is given then file will be downloaded.
```
jamessynge: Not sure how "blob name" relates to remote_path.

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+            save_dir (str, optional): Path for saved file, defaults to current directory.
+            force (bool, optional): Force a download even if file exists locally, default False.
```
jamessynge: Perhaps call this overwrite, which we use elsewhere?

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+            callback (callable, optional): A callable object that gets called at end of
+                function.
+
+        Returns:
+            str: Path to local (uncompressed) FITS file
+        """
+        # Get blob object if just a path
```
jamessynge: Totally a nit: Please end sentences with a period.

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+        that are 80 bytes long each. The Header Unit always ends with the single
+        word 'END' on a line (not necessarily line 36).
+
+        Here the header is streamed from Storage until the 'END' is found, with
+        each line given minimal parsing.
+
+        See https://fits.gsfc.nasa.gov/fits_primer.html for overview of FITS format.
+
+        Args:
+            remote_blob (google.cloud.storage.blob.Blob): Blob object
```
jamessynge: Any reason to not allow a string as done in the prior method?

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+            i = 2  # We skip the compression header info
+
+        headers = dict()
+
+        streaming = True
+        while streaming:
+            # Get a header card
+            start_byte = 2880 * (i - 1)
+            end_byte = (2880 * i) - 1
+            b_string = remote_blob.download_as_string(start=start_byte, end=end_byte)
```
jamessynge: Unless you know that the FITS header is usually just one block, I'd be inclined to simply download the first N blocks of size 2880 bytes (e.g. 50 blocks), then operate on each of those blocks (or really treat it as a longer sequence of 80 byte lines. This will very likely be faster than downloading two 2880 blocks.

Or test to see if I'm right.

pocs/utils/google/storage.py
```diff
@@ -97,3 +99,247 @@ def upload(self, local_path, remote_path=None):
...
+                        v = int(v)
+
+                    headers[k.strip()] = v
+
+            i += 1
+
+        return headers
+
+
+def upload_observation_to_bucket(pan_id,
```
jamessynge: How about merging most of the body of this into the class above at some point? A TODO will be fine.

pocs/tests/utils/google/test_storage.py
wtgee: I think there is a generic testing auth_key. I saw reference to that somewhere, will try to find.

pocs/utils/google/storage.py
wtgee: We know that the first block is the compression header information (for `.fz` packed files) so we are skipping that block and starting at block 2. After that we don't know how many blocks it will be.

I'll add some comments so it's clear what is trying to be done then we can figure out if that is what actually should be done. :)

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
...
+
+pytestmark = pytest.mark.skipif(
+    not pytest.config.option.test_cloud_storage,
+    reason="Needs --test-cloud-storage to run"
+)
+
+
+@pytest.fixture(scope="module")
+def auth_key():
+    local_config = load_config('pocs_local', ignore_local=True)
```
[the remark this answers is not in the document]
wtgee: We now rely on authentication being done for the system as a whole. As such, I am removing the ability to to pass an auth_key to the `PanStorage` object directly. That also then makes some of this testing infrastructure disappear.

pocs/tests/utils/google/test_storage.py
```diff
@@ -0,0 +1,111 @@
...
+from pocs.utils.google.storage import PanStorage, upload_observation_to_bucket
+
+
+pytestmark = pytest.mark.skipif(
+    not pytest.config.option.test_cloud_storage,
+    reason="Needs --test-cloud-storage to run"
+)
+
+
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
wtgee: Same as above.

pocs/utils/google/storage.py
wtgee: Good call, thanks.

PR #602, opened 2018-09-17
Update google clouddb

wtgee:
Part of #599, depends on #600.

Adds the `cloud_sql_proxy` from google for connecting to cloud sql instances. Also includes a simple wrapper script that will pull the setup information from the config.

Adds a few utility functions:

* `get_db_proxy_conn`
* `get_cursor`
* `meta_insert`
* `add_header_to_db`

**Note** As of now this still requires the user to set a PGPASSWORD to authenticate against the actual database table. We might want to look at ways to mitigate that.

### DIFF
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -55,6 +55,29 @@ messaging:
     # Must match ports in peas.yaml.
     cmd_port: 6500
     msg_port: 6510
+
+############ Google Network ############
+# By default all images are stored on google
+# cloud servers and we also use a few google
+# services to store metadata, communicate with
+# servers, etc.
+#
+# See $POCS/pocs/utils/google/README.md for
+# details about authentication.
+########################################
+panoptes_network:
+    image_storage: True
+    project_id: panoptes-survey
+    buckets:
+        images: panoptes-survey
+    cloudsql_instances:
+        meta:
+            location: us-central1
+            database: panoptes-meta
+        tess:
+            location: us-central1
+            database: tess-catalog            
+        
 #Enable to output POCS messages to social accounts
 # social_accounts:
 #     twitter:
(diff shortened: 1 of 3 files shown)


conf_files/pocs.yaml
```diff
@@ -55,6 +55,31 @@ messaging:
     # Must match ports in peas.yaml.
     cmd_port: 6500
     msg_port: 6510
+
+############ Google Network ############
+# By default all images are stored on google
```
jamessynge: I'm sure we could fit this text on two lines instead of four! ;-)

scripts/connect_clouddb_proxy.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+
+####################################
+# connect_clouddb_proxy.py
+#
+# This is a simple wrapper that looks up
```
jamessynge: I recall advocating for 80 column lines, but this is really short! ;-)

scripts/connect_clouddb_proxy.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+import subprocess
+from pprint import pprint
+
+from pocs.utils.config import load_config
+
+
+def main(instance_id, local_port, verbose=False):
+    proxy_cmd = os.path.join(os.environ['POCS'], 'bin', 'cloud_sql_proxy')
+    connection_str = '-instances={}=tcp:{}'.format(instance_id, local_port)
```
jamessynge: `connection_str` seems like the wrong name for this var. Not sure if the following makes sense:
```
connection_str = '{}=tcp:{}'.format(instance_id, local_port)
instances_arg = '-instances={}'.format(connection_str)
```

scripts/connect_clouddb_proxy.py
wtgee: Soon I'm going to push for 2-character lines! ;)

PR #604, opened 2018-09-18
Remove unused exception variables

jamessynge:
Caught by pyflakes.


bin/peas_shell
```diff
@@ -354,7 +354,7 @@ class PanSensorShell(cmd.Cmd):
             sensor = getattr(self, sensor_name)
             try:
                 sensor.capture(store_result=True, send_message=True)
-            except Exception as e:
+            except Exception:
```
wtgee: Should we be figuring out what exceptions this might be?

jamessynge: I've got a replacement coming for this code, so don't want to bother.

PR #605, opened 2018-09-18
Use topic terminology from ZeroMQ

### LINKED ISSUE #597: Switch 'channel' and 'msg_type' to 'topic'
There are three names for 'topic'. Let's have just one.
@wtgee Let me know if you agree, in which case I'll make the change (i.e. assign it back to me).

jamessynge:
We were using both channel and msg_type to refer to what ZeroMQ
calls topics. No need for three terms, so this changes them
all to topic.

Fixes #597.

### DIFF
--- a/pocs/utils/messaging.py
+++ b/pocs/utils/messaging.py
@@ -146,7 +146,7 @@ class PanMessaging(object):
 
         Args:
             port (int):         The port (on localhost) to bind to.
-            topic (str):      Which topic topic to subscribe to.
+            topic (str):      Which topic or topic prefix to subscribe to.
 
         """
         obj = cls()
--- a/pocs/utils/messaging.py
+++ b/pocs/utils/messaging.py
@@ -32,9 +32,9 @@ class PanMessaging(object):
     an instance of PanMessaging, on which they can then call
     receive_message.
 
-    Messages are sent to channels, a name that can be used to allow
-    a high-level partitioning of messages. A channel name may not
-    include whitespace. Among the currently used channel names are:
+    Messages are sent to topics, a name that can be used to allow
+    a high-level partitioning of messages. A topic name may not
+    include whitespace. Among the currently used topic names are:
 
       * PANCHAT (sent from POCS.say)
       * PAWS-CMD (sent from PAWS websockets.py)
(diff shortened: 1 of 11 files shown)


pocs/utils/messaging.py
```diff
@@ -32,9 +32,9 @@ class PanMessaging(object):
     an instance of PanMessaging, on which they can then call
     receive_message.
 
-    Messages are sent to channels, a name that can be used to allow
-    a high-level partitioning of messages. A channel name may not
-    include whitespace. Among the currently used channel names are:
+    Messages are sent to topics, a name that can be used to allow
```
wtgee: Is this the best place to record all the current topics?

pocs/utils/messaging.py
```diff
@@ -46,25 +46,20 @@ class PanMessaging(object):
       * telemetry:commands (in ArduinoIO... new)
       * camera:commands (in ArduinoIO... new)
 
-    And some other channels are used in tests:
+    And some other topics are used in tests:
 
-      * TEST-CHANNEL (test_messaging.py)
+      * Test-Topic (test_messaging.py)
```
wtgee: Seems like we now have a mix of naming conventions. I know they are all allowed but should we have some logical scheme?

pocs/utils/messaging.py
```diff
@@ -77,14 +72,14 @@ class PanMessaging(object):
...
     a byte array of this format:
-        <channel-name><space><serialized-value>
+        <topic-name><space><serialized-value>
     """
     logger = get_root_logger()
 
-    # Channel names must consist of the characters.
-    name_re = re.compile('[a-zA-Z][-a-zA-Z0-9_.:]*')
+    # Topic names must consist of the characters.
+    topic_name_re = re.compile('[a-zA-Z][-a-zA-Z0-9_.:]*')
```
wtgee: Don't need the `r'` with `re.compile`?

pocs/utils/messaging.py
```diff
@@ -146,16 +141,16 @@ def create_publisher(cls, port, bind=False, connect=True):
...
 
     @classmethod
-    def create_subscriber(cls, port, channel='', bind=False, connect=True):
+    def create_subscriber(cls, port, topic='', bind=False, connect=True):
         """ Create a listener
 
         Args:
             port (int):         The port (on localhost) to bind to.
-            channel (str):      Which topic channel to subscribe to.
+            topic (str):      Which topic topic to subscribe to.
```
wtgee: repeat of `topic`

pocs/utils/messaging.py
jamessynge: My plan is to develop a scheme for future topics and to also write general documentation about topics and filtering. I've created issue #606 for that.

pocs/utils/messaging.py
jamessynge: That change was deliberate, just to test that the title-case form was supported.

pocs/utils/messaging.py
jamessynge: Raw strings are only needed when the string includes characters to be escaped AND you don't want to double escape them... or multi-line strings.

pocs/utils/messaging.py
jamessynge: Fixed.

PR #611, opened 2018-09-20
Cache astrometry index files

wtgee:
Cache the astrometry index files as they never change. Helps with situations where astrometry.net is down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,6 @@ script:
   - coverage combine
 
 after_success:
-    - if [[ $TRAVIS_PYTHON_VERSION == 3.6* ]]; then
-          bash <(curl -s https://codecov.io/bash);
-      fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6* ]]; then
+        bash <(curl -s https://codecov.io/bash);
+    fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ services:
 python:
   - "3.6"
 cache:
-  - pip
+  pip: true
+  directories:
+    - /var/panoptes/astrometry/data
 env:
   - POCS=$TRAVIS_BUILD_DIR PANDIR=/var/panoptes PANUSER=$USER ARDUINO_VERSION=1.8.1
 before_install:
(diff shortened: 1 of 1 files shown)


wtgee:
I'm not entirely sure this is working yet and cache errors don't cause build errors. Need to look into calling order a bit to figure out why still getting permission errors on the directory.

.travis.yml
```diff
@@ -1,20 +1,19 @@
...
   - "3.6"
 env:
-  - POCS=$TRAVIS_BUILD_DIR PANDIR=/var/panoptes PANUSER=$USER ARDUINO_VERSION=1.8.1
-before_cache:
-    - sudo mkdir -p $PANDIR/astrometry/data && sudo chmod -R 777 $PANDIR
-cache:
-  pip: true
-  directories:
-    - $PANDIR/astrometry/data
+  - POCS=$TRAVIS_BUILD_DIR
```
jamessynge: Thanks, much nicer to have them each on their own line.

.travis.yml
```diff
@@ -1,20 +1,19 @@
...
-  pip: true
-  directories:
-    - $PANDIR/astrometry/data
+  - POCS=$TRAVIS_BUILD_DIR 
+  - PANDIR=/var/panoptes 
+  - PANUSER=$USER 
+  - ARDUINO_VERSION=1.8.1
 before_install:
-    - sudo mkdir -p $PANDIR/astrometry/data && sudo chmod -R 777 $PANDIR
+    - sudo mkdir -p $PANDIR && sudo chmod -R 775 $PANDIR
```
jamessynge: Out of curiosity, why is the mode not 755?

.travis.yml
```diff
@@ -1,20 +1,19 @@
...
+    - sudo chown -R $PANUSER:$PANUSER $PANDIR
+    - sudo addgroup $USER $PANUSER
```
jamessynge: Doesn't addgroup require logging out and back in before it will take effect?

.travis.yml
wtgee: Just changed. I didn't realize `$USER==$PANUSER` here.

.travis.yml
wtgee: Removed as per above.

Issue #95, opened 2017-07-15
Camera.is_primary is wrong


jamessynge:
I provided a "primary: uid" line in pocs_local.yaml, but the wrong camera is considered primary. pocs_local.yaml:

```
cameras:
    auto_detect: True
    # Arbitrary choice of primary, but now labeled as camera #0,
    # and plugged in to jack #0.
    primary: 6575fc
    devices:
    -
        model: canon_gphoto2
    -
        model: canon_gphoto2
```

And the serial numbers of the cameras start and end as follows:
```
7bab97...cdc74a
6575fc...a6c489
```

wtgee:
This should be fixed by #650. Reopen if still a problem.

PR #612, opened 2018-09-20
Remove camera creation from Observatory

wtgee:
* First step toward pulling cameras out of Observatory.
* Basic `create_cameras_from_config` function. This is basically copied over from the `_create_cameras` that was in `Observatory`, with no intended behavior change.
* Creation removed from `Observatory` .
 * Add/remove camera methods.
* Updates to observatory init (and testing).
* **Behavior Change** Accessing `observatory.primary_camera` will now always return a camera if `observatory.has_cameras`. If no primary camera was set, the first camera in the list will become the primary.


bin/pocs_shell
```diff
@@ -152,7 +153,8 @@ class PocsShell(Cmd):
             simulator = []
 
         try:
-            observatory = Observatory(simulator=simulator)
+            cameras = create_cameras_from_config()
+            observatory = Observatory(simulator=simulator, cameras=cameras)
```
jamessynge: FWIW, while thinking about adding a supervisor and better messaging, and also while reviewing Anthony's distributed cameras PR (which includes a nameserver), I wondered about introducing some kind of device registry. Something we can discuss in the future.

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
...
+
 from pocs.camera.camera import AbstractCamera  # pragma: no flakes
 from pocs.camera.camera import AbstractGPhotoCamera  # pragma: no flakes
+
+from pocs.utils import list_connected_cameras
+from pocs.utils import logger as logger_module
+
+
+def create_cameras_from_config(config=None, logger=None, **kwargs):
+    """Creates a camera object(s)
```
jamessynge: Creates camera object(s) based on the config.

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
...
+
+    Loads the cameras via the configuration.
```
jamessynge: Drop this line.

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
...
+
+    Creates a camera for each camera item listed in the config. Ensures the
+    appropriate camera module is loaded.
+
+    Note:
+        This does not actually make a connection to the camera. To do so,
+        call 'camera.connect()' explicitly.
```
jamessynge: ... on each camera.

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
...
+
+    Args:
+        **kwargs (dict): Can pass a `cameras` object that overrides the info in
+            the configuration file. Can also pass `auto_detect`(bool) to try and
+            automatically discover the ports.
+
+    Returns:
+        OrderedDict: An ordered dictionary of created camera objects.
```
jamessynge: What is the key? And why is it ordered?

jamessynge: Or None if there are no cameras declared in the config.

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
+from collections import OrderedDict
+
+from pocs.utils import error
+from pocs.utils import load_module
+from pocs.utils.config import load_config
+
 from pocs.camera.camera import AbstractCamera  # pragma: no flakes
 from pocs.camera.camera import AbstractGPhotoCamera  # pragma: no flakes
+
+from pocs.utils import list_connected_cameras
```
jamessynge: FWIW, seems like the kind of function that could be moved to this file.

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
...
+    for cam_num, device_config in enumerate(device_info):
+        cam_name = 'Cam{:02d}'.format(cam_num)
+
+        if not a_simulator:
+            camera_model = device_config.get('model')
+
+            # Assign an auto-detected port. If none are left, skip
+            if auto_detect:
+                try:
+                    camera_port = ports.pop()
```
jamessynge: I'm perplexed by this. If the config file lists two canon cameras with the prefixes of their serial numbers, how do we know the order in which to pop the ports?

pocs/camera/__init__.py
```diff
@@ -1,2 +1,157 @@
...
+        error.CameraNotFound: Description
+        error.PanError: Description
+    """
+    if not logger:
+        logger = logger_module.get_root_logger()
+
+    if not config:
+        config = load_config(**kwargs)
+
+    if 'cameras' not in config:
```
jamessynge: Why is this not after the line: camera_info = kwargs_or_config('cameras')

pocs/observatory.py
```diff
@@ -47,10 +47,15 @@ def __init__(self, *args, **kwargs):
         self.mount = None
         self._create_mount()
 
-        self.logger.info('\tSetting up cameras')
-        self.cameras = OrderedDict()
-        self._primary_camera = None
-        self._create_cameras(**kwargs)
+        if not cameras:
+            cameras = OrderedDict()
```
jamessynge: Why not just raise an exception?

pocs/observatory.py
```diff
@@ -107,6 +128,42 @@ def current_observation(self, new_observation):
     def has_dome(self):
         return self.dome is not None
 
+
+##########################################################################
+# Device Getters/Setters
+##########################################################################
+
+    def add_camera(self, cam_name, camera):
```
jamessynge: Do you intend these two add/remove methods to be used to dynamically add and remove cameras during operation?

pocs/tests/test_observatory.py
```diff
@@ -25,7 +26,11 @@ def simulator():
 @pytest.fixture
 def observatory(config, simulator):
     """Return a valid Observatory instance with a specific config."""
-    obs = Observatory(config=config, simulator=simulator, ignore_local_config=True)
+    cameras = create_cameras_from_config(config)
```
jamessynge: local=False 

Or some such to prevent the pocs_local.yaml from being picked up?

bin/pocs_shell
wtgee: Yes, I think we will definitely want to change how the Observatory interacts with and discovers devices. Hopefully this gets us closer to doing that.

pocs/camera/__init__.py
wtgee: This is only if auto_detect is true, which means that it is ignoring the config. Serial number is set when camera object is initialized by lookup (via gphoto2)

pocs/observatory.py
wtgee: Yes, eventually that would be the goal.

pocs/tests/test_observatory.py
wtgee: Doesn't the `config` fixture handle that?

pocs/camera/__init__.py
jamessynge: Ah. Then I may have misunderstood how to auto_detect and primary work. I included the serial number prefixes and primary:True in pocs_local.yaml on PAN006. My goal is to make sure that a specific camera is treated as primary every time. And ideally we'd get a warning if a specified camera is not found, or if another camera is found.

pocs/tests/test_observatory.py
jamessynge: Oh, it may do so. NM.

pocs/camera/__init__.py
wtgee: I think #95 is still true unfortunately. This function could certainly be cleaned up some. Happy to do it here if we want just to make sure it gets done soon but I imagine it will all need to be revisited anyway if we set up separate process controls for the camera. Which hopefully happens soon. :)

pocs/observatory.py
wtgee: I could. `create_dome_from_config` is returning None so I'll go for consistency with that (although see comment below).

wtgee:
> This new framework will be really useful for refactoring my distributed camera code to allow for camera addition/removal without restarting everything from scratch every time.

I would prefer to merge this before #547 so we can keep some things cleaner, but it will definitely be the bigger change for you. I'll go ahead and get this merged though so that work can start.

AnthonyHorton:
> I would prefer to merge this before #547 so we can keep some things cleaner, but it will definitely be the bigger change for you. I'll go ahead and get this merged though so that work can start.

Oh, absolutely. What I meant by my comment was just that I'm especially keen to see this merged as soon as possible so that I can use it as a guide for refactoring of #547, which (unlike this PR) is still some way off being ready to merge.

PR #613, opened 2018-09-20
Use unit_id for sequence and image ids

### LINKED ISSUE #586: Fix sequence and image ids
The sequence_id and image_id (SEQ_ID and IMG_ID) are built from the `<unit_id>_<camera_id>_<sequence_time>` but the value for `unit_id` is being pulled from the `config['name']`. This has created problems for PAN006 where there is a custom name instead of a unit_id. Now that we actually have a unit_id parameter in the config, the sequence_id and image_id should be changed.

wtgee:
* **Behavior Change** Enables `pan_id` in the config by default. `PAN000` as simulator?

Closes #586

### DIFF
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -1,6 +1,18 @@
 ---
+######################### PANOPTES UNIT ########################################
+# name:   Can be anything you want it to be. This name is displayed in several
+#         places and should be a "personal" name for the unit.
+#
+# pan_id: This is an identification number assigned by the PANOPTES team and is
+#         the official designator for your unit. This id is used to store image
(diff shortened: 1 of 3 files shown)


conf_files/pocs.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: Generic PANOPTES Unit
-# pan_id: PANxyz   # This unit's assigned identifier.
+pan_id: PAN000   # This unit's assigned identifier.
```
jamessynge: ... PAN000 is a "fake" id that no unit uses.

wtgee: I greatly expanded this.

PR #616, opened 2018-09-20
Only use Mongo DB type during local testing - Local Testing with 1/3rd the wait!

wtgee:
~~* Adds `--test-all-databases` option which is False by default but enabled for travis tests.~~

Edit: Added `--test-databases [<db_name>]` option to support testing of certain db types.

Call with, e.g. `pytest -xv --test-databases file memory`. Supports an all option (which travis uses). Default 'mongo' only.

Helps with #191 .

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -47,10 +47,10 @@ def pytest_addoption(parser):
         help="Tests cloud strorage functions." +
         "Requires $PANOPTES_CLOUD_KEY to be set to path of valid json service key")
     group.addoption(
-        "--test-all-databases",
-        action="store_true",
-        default=False,
-        help="If all the database types should be tested, default False uses mongo")
+        "--test-databases",
+        nargs="+",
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -214,6 +219,10 @@ def can_connect_to_mongo():
 
 @pytest.fixture(scope='function', params=['mongo', 'file', 'memory'])
```
jamessynge: FWIW, I'd prefer an additional option that is the list of dbs to be tested, which then would provide the value for params.

wtgee: That's almost what I did. Would be easy enough to change if we feel it is worth it do to right now.

wtgee: I changed it to support this now. Call with, e.g. `pytest -xv --test-databases file memory`.  Supports an `all` option (which travis uses). Default 'mongo' only.

PR #622, opened 2018-09-22
Fixes for focuslynx

wtgee:
* Fix incorrectly called methods
* Handle missing uid for formatting (why is uid missing?)
* Better handling of errors

More from #547

### DIFF
--- a/pocs/focuser/focuslynx.py
+++ b/pocs/focuser/focuslynx.py
@@ -288,10 +288,6 @@ class Focuser(AbstractFocuser):
         Utility function that handles the common aspects of sending commands and
         parsing responses.
         """
-        name = self.port
-        if hasattr(self, '_focuser_config'):
-            name = self.uid
-
         # Make sure we start with a clean slate
         self._serial_port.reset_output_buffer()
         self._serial_port.reset_input_buffer()
@@ -299,14 +295,14 @@ class Focuser(AbstractFocuser):
         self._serial_port.write(command_str.encode('ascii'))
         response = str(self._serial_port.readline(), encoding='ascii').strip()
         if not response:
-            message = "No response to command '{}' from focuser {}".format(command_str, name)
+            message = "No response to command '{}' from focuser {}".format(command_str, self.uid)
             self.logger.error(message)
             raise RuntimeError(message)
 
         # Should always get '!' back unless there's an error
         if response != '!':
             message = "Error sending command '{}' to focuser {}: {}".format(
-                command_str, name, response)
+                command_str, self.uid, response)
             self.logger.error(message)
             raise RuntimeError(message)
 
@@ -314,7 +310,7 @@ class Focuser(AbstractFocuser):
         command_echo = str(self._serial_port.readline(), encoding='ascii').strip()
         if command_echo != expected_reply:
             message = "Expected reply '{}' from {}, got '{}'".format(
-                expected_reply, name, command_echo)
+                expected_reply, self.uid, command_echo)
             self.logger.error(message)
             raise RuntimeError(message)
 
--- a/pocs/focuser/focuslynx.py
+++ b/pocs/focuser/focuslynx.py
@@ -65,9 +65,8 @@ class Focuser(AbstractFocuser):
                 else:
                     raise ValueError('Max position must be greater than min position!')
             else:
-                message = "Specified max_position {} greater than focuser max {}!".format(max_position,
-                                                                                          self._max_position)
-                warn(message)
+                self.logger.warning("Specified max_position {} greater than focuser max {}!",
+                                    max_position, self._max_position)
 
         if initial_position is not None:
             self.position = initial_position
@@ -96,8 +95,8 @@ class Focuser(AbstractFocuser):
     @uid.setter
     def uid(self, nickname):
         if len(nickname) > 16:
-            self.logger.warning('Truncated nickname {} to {} (must be <= 16 characters)'.format(nickname,
-                                                                                                nickname[:16]))
+            self.logger.warning('Truncated nickname {} to {} (must be <= 16 characters)',
+                                nickname, nickname[:16])
             nickname = nickname[:16]
         command_str = '<F{:1d}SCNN{}>'.format(self._focuser_number, nickname)
         self._send_command(command_str, expected_reply='SET')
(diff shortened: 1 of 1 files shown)


pocs/focuser/focuslynx.py
```diff
@@ -271,8 +272,8 @@ def _initialise(self):
         self.logger.info('{} initialised'.format(self))
 
     def _get_hub_info(self):
-        self._hub_info = self._send_comman(command_str='<FHGETHUBINFO>',
-                                           expected_reply='HUB INFO')
+        self._hub_info = self._send_command(command_str='<FHGETHUBINFO>',
```
jamessynge: Oh, that's a good one to find! I wonder why pyflakes or one of our other linters didn't catch that.

pocs/focuser/focuslynx.py
```diff
@@ -294,19 +295,36 @@ def _send_command(self, command_str, expected_reply):
         Utility function that handles the common aspects of sending commands and
         parsing responses.
         """
+        name = self.port
+        if hasattr(self, '_focuser_config'):
```
jamessynge: That seems like an odd test for whether or not self.uid is present & valid. Perhaps a name attr is needed, initialized where _focuser_config is first set in (presumably in __init__).

pocs/focuser/focuslynx.py
```diff
@@ -294,19 +295,36 @@ def _send_command(self, command_str, expected_reply):
...
         self._serial_port.reset_output_buffer()
         self._serial_port.reset_input_buffer()
         # Send command
         self._serial_port.write(command_str.encode('ascii'))
-        # Should always get '!' back unless there's an error
         response = str(self._serial_port.readline(), encoding='ascii').strip()
-        if response != '!':
-            message = "Error sending command '{}' to {}: {}".format(command_str, self.uid, response)
+        if not response:
+            message = "No response to command '{}' from {}".format(command_str, name)
```
jamessynge: Maybe the error messages here could include the literal string "focuser", i.e. "No response to command 'foo' from focuser 'bar'"

pocs/focuser/focuslynx.py
wtgee: I thin it doesn't (can't?) check things that are class methods, which is a little unfortunate.

pocs/focuser/focuslynx.py
wtgee: @AnthonyHorton can you maybe speak to why it was set up this way in #547? I cleaned it up a bit to assume the port is available and then do the check for the `_focuser_config`, so it is slightly different but functionally the same.  Would we be okay to change the `.name` property to better handle this situation rather than do it here?

AnthonyHorton: Yeah, this was a middle of the night fix!  All the changes here stem from receiving extremely unhelpful errors for a situation where the FocusLynx focus controller was powered off. The initialisation would get as far as opening the serial port (because that's just a USB-RS232 converter that *does* have power) but then gets no response to the first command it sends, a request for hub info. At that point I was getting a confusing `AttributeError` about `_focuser_config` being missing because it was trying to put the UID into the error message, and the getter for UID tried to get it from `_focuser_config`, which isn't set until getting the focuser config from the controller with the next command.

I make some quick changes so that the error handling would still produce an error message in the case of an error occurring before `_focuser_config` has been set, which enabled me to discover that the actual problem was that no response was being received at all, so I added another handler for that specific case, then went and checked the power switch on the side of the focus controller...

Bit wary of doing anything with the `.name` property. All the `Camera` and `Focuser` classes have this property and at the moment use it in a fairly consistent manner. Would prefer to avoid any `FocusLynx` specific behaviour with that. Maybe give the UID getter a fallback option where it returns `port` if `_focuser_config` is not set?

Side note: for the FocusLynx focusers the 'UID' is __not__ guaranteed to always be unique. There are no serial numbers accessible to external code, the only available identifier is the user set 'nickname' which is saved in the focus controller, and that can be any string the user wants. I don't think non-unique IDs would actually cause problems for focusers (unlike cameras, where it would mess things up) but something to watch out for.

pocs/focuser/focuslynx.py
```diff
@@ -299,14 +293,27 @@ def _send_command(self, command_str, expected_reply):
...
+        if not response:
+            message = "No response to command '{}' from focuser {}".format(command_str, self.uid)
+            self.logger.error(message)
+            raise RuntimeError(message)
+
+        # Should always get '!' back unless there's an error
         if response != '!':
-            message = "Error sending command '{}' to {}: {}".format(command_str, self.uid, response)
+            message = "Error sending command '{}' to focuser {}: {}".format(
+                command_str, self.uid, response)
```
AnthonyHorton: Need to add a fallback option to the `uid` getter otherwise this will fail in the case where `_focuser_config` has not been set yet.

AnthonyHorton:
I pushed a commit with a few other small bug fixes for FocusLynx to `pyro_cameras` this morning too (were  sitting on the Raspberry Pi from more middle of the night bug squashing during the recent commissioning run).  They should move into here too.

wtgee:
> I pushed a commit with a few other small bug fixes for FocusLynx to pyro_cameras this morning too (were sitting on the Raspberry Pi from more middle of the night bug squashing during the recent commissioning run). They should move into here too.

Was going to ask if that was okay to move here. I'll include them, thanks.

pocs/focuser/focuslynx.py
wtgee: I got an email with your comment but can't seem to find it here. Add back the check for the `_focuser_config` and show the port otherwise?

wtgee: N/M, I see the comment about the `try/except`. Added in `uid` getter and gives back port if not.

AnthonyHorton: I wrote a detailed response to your request for an explanation of the changes, which GitHub almost immediately hid as 'outdated' when you pushed another commit... Its up there, just need to click 'show outdated' until you find it.

pocs/focuser/focuslynx.py
```diff
@@ -28,6 +28,7 @@ class Focuser(AbstractFocuser):
     Additional positional and keyword arguments are passed to the base class, AbstractFocuser. See
     that class for a complete list.
     """
+
```
jamessynge: Doesn't the args section belong in the docstring for __init__ rather than the class?

wtgee: There's not actually a preference according to the google style other than to say it should be consistent for a project. i think we have a lot of inconsistency in this regard. :(

wtgee: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

PR #625, opened 2018-09-23
Timer Utility Change

wtgee:
* `Timeout` renamed to `Timer` to reflect usage. Note, I did not rename the serialutil equivalent.
* Generic helper method in POCS to wait for events, essentially a wrapper around the `Timer`.
* Cleanup of `observing` and `pointing` states to use `wait_for_events`

### DIFF
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -13,6 +13,9 @@ from pocs.observatory import Observatory
 from pocs.state.machine import PanStateMachine
 from pocs.utils import current_time
 from pocs.utils import get_free_space
+from pocs.utils import Timer
+from pocs.utils import listify
+from pocs.utils import error
 from pocs.utils.messaging import PanMessaging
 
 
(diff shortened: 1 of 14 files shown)


pocs/core.py
```diff
@@ -434,6 +437,65 @@ def sleep(self, delay=2.5, with_status=True):
...
+                        events,
+                        timeout,
+                        sleep_delay=1,
+                        status_interval=10,
+                        msg_interval=30,
+                        event_type='generic'):
+        """Wait for event(s) to be set.
+
+        This method will wait for a maximum of `timeout` seconds for all of the
+        `events` to complete. The loop will `check_messages` each time in case of
```
jamessynge: Will check at least every `sleep_delay` seconds for the events to be done, and also for interrupts and bad weather. Will log debug messages approximately every `status_interval` seconds, and will output status messages approximately every `status_interval` seconds.

pocs/core.py
```diff
@@ -434,6 +437,65 @@ def sleep(self, delay=2.5, with_status=True):
...
+            if now >= next_status_time:
+                self.status()
+                next_status_time += status_interval
+                now = current_time()
+
+            if timer.expired():
+                raise error.Timeout
+
+            # Sleep for a little bit.
+            time.sleep(sleep_delay.value)
```
jamessynge: The .value makes me think you're passing in an astropy time quantity rather than the documented float.

pocs/core.py
```diff
@@ -434,6 +437,65 @@ def sleep(self, delay=2.5, with_status=True):
...
+        next_msg_time = start_time + msg_interval
+
+        while not all([event.is_set() for event in events.values()]):
+            self.check_messages()
+            if self.interrupted:
+                self.logger.info("Waiting for events has been interrupted")
+                break
+
+            now = current_time()
+            if now >= next_msg_time:
```
jamessynge: FWIW, you could choose to use Timer for this too, but that isn't trivial because it isn't a PeriodicTimer. Perhaps we should write such a class... sometime.

pocs/state/states/default/observing.py
```diff
@@ -1,12 +1,6 @@
...
 from pocs import utils as pocs_utils
 
-SLEEP_SECONDS = 1.0
-STATUS_INTERVAL = 10. * u.second
-WAITING_MSG_INTERVAL = 30. * u.second
-MAX_EXTRA_TIME = 60 * u.second
+MAX_EXTRA_TIME = 60  # seconds
 
 
 def on_enter(event_data):
```
jamessynge: Can probably streamline the docstring now.

pocs/state/states/default/pointing.py
```diff
@@ -34,44 +31,24 @@ def on_enter(event_data):
...
-
-                pocs.logger.debug('Waiting for images: {} seconds'.format(wait_time))
-                pocs.status()
+            pocs.logger.debug("Taking pointing image {}/{} on: {}",
+                              img_num, num_pointing_images, primary_camera)
 
-                if wait_time > timeout:
-                    raise error.Timeout("Timeout waiting for pointing image")
+            # Start the exposure
+            camera_event = primary_camera.take_observation(
```
jamessynge: Is the type of camera_event `list of events`? If so, please rename as camera_events, observation_done_events, or some such.

pocs/utils/__init__.py
```diff
@@ -49,7 +49,7 @@ def flatten_time(t):
 
 
 # This is a streamlined variant of PySerial's serialutil.Timeout.
-class Timeout(object):
+class Timer(object):
```
jamessynge: I suppose this is a good change since I see we already have an error type called Timeout. However, note also that we're using threading.Timer in some places. If you are concerned by that, you could consider calling this TimeLimit, which is not a symbol that we use.

jamessynge:
I forgot to say, can you come up with any kind of focused test?

pocs/core.py
wtgee: Thanks, I think I was getting lazy. :)

pocs/core.py
wtgee: Changed to support both. It was somewhere in between.

pocs/state/states/default/pointing.py
wtgee: Here it's just a single event. The `wait_for_events` handles either a single Event or a list.

pocs/utils/__init__.py
wtgee: `CountdownTimer`?

wtgee:
Note that test coverage should increase after #627.

PR #626, opened 2018-09-23
CloudSQL Proxy Startup

wtgee:
Adding a script for starting the google cloudsql proxy, connecting to the meta db by default.

@jamessynge is there any way to test these startup scripts?

Part of #599

### DIFF
--- a/scripts/startup/start_panoptes_in_tmux.sh
+++ b/scripts/startup/start_panoptes_in_tmux.sh
@@ -99,6 +99,10 @@ set -x
 # These provide connections between message publishers and subscribers.
 create_and_init_window msg_hub start_messaging_hub.sh
 
+# Create a window running the Google CloudSQL proxy script, connecting
+# to the meta database.
+create_and_init_window meta_db start_cloudsql_proxy.sh meta
+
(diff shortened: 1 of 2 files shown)


jamessynge:
Sadly there isn't yet a way to test these scripts. I'll give it some thought.

scripts/startup/start_panoptes_in_tmux.sh
```diff
@@ -99,6 +99,10 @@ set -x
 # These provide connections between message publishers and subscribers.
 create_and_init_window msg_hub start_messaging_hub.sh
 
+# Create a window running the Google CloudSQL proxy script, connecting
+# to the meta database.
+create_and_init_window meta_db start_cloudsql_proxy.sh meta
```
jamessynge: What does this imply about **needing** an internet connection to run POCS?

wtgee: Good point. It actually also implies that you have done the authentication with Google and doesn't handle if not.

More generally, I was thinking about how we could have a "setup" for the startup script to choose which scripts get started. The social messaging, for instance, would be the same. Created #643.

PR #627, opened 2018-09-23
Fixing tests related to image solving with simulator

wtgee:
* Use unsolved image so we can test solving
* Pass `skip_solved=False` during `analyze_recent`. Helps with testing when
the same image is used repeatedly, shouldn't affect normal usage.
* Add some extra logging
* Misc cleanup

### DIFF
--- a/pocs/camera/simulator.py
+++ b/pocs/camera/simulator.py
@@ -83,7 +83,7 @@ class Camera(AbstractCamera):
 
     def _fake_exposure(self, filename, header, exposure_event):
         # Get example FITS file from test data directory
-        file_path = "{}/pocs/tests/data/{}".format(os.getenv('POCS'), 'solved.fits')
+        file_path = os.path.join(os.getenv('POCS'), 'pocs', 'tests', 'data', 'unsolved.fits')
         fake_data = fits.getdata(file_path)
 
         if header['IMAGETYP'] == 'Dark Frame':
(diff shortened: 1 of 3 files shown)


AnthonyHorton:
An aside on the image solving, we noticed during the recent testing with Huntsman that if given ludicrously large images the solving will timeout. We saw this consistently when one of the 50 Mpixel FLI ML50100 cameras was enabled, which due to its name would become the default primary imager. We got a warning in the log about analysis failing due to solving timing out for every image we took.

wtgee:
> An aside on the image solving, we noticed during the recent testing with Huntsman that if given ludicrously large images the solving will timeout. We saw this consistently when one of the 50 Mpixel FLI ML50100 cameras was enabled, which due to its name would become the default primary imager. We got a warning in the log about analysis failing due to solving timing out for every image we took.

There is a `timeout` param than can be passed through `get_solve_field` that will let it keep going. I use that on the Pi where the solve can take 45-60 seconds. Default is 30.

But we should probably figure out if the solving can be done asynchronously (and/or if needed at all for Huntsman observing).

AnthonyHorton:
> There is a `timeout` param than can be passed through `get_solve_field` that will let it keep going. I use that on the Pi where the solve can take 45-60 seconds. Default is 30.

I suspected that there was a parameter for this somewhere.
 
> But we should probably figure out if the solving can be done asynchronously (and/or if needed at all for Huntsman observing).

Definitely. The processing that goes on at the end of each science exposure can take quite a while. It's not a huge issue for 5+ minute long galaxy exposures but it is contributing to some substantial gaps in temporal coverage when trying bright star transits. Should look at deferring as much as possible to after the next exposure has started.

Issue #555, opened 2018-08-24
Consider dropping most environment variables


jamessynge:
We have PANDIR, POCS, etc. And then we have config files that have the values of those envvars. It seems like we should have as few envvars as possible, and infer them where possible, or record them in config files where not.

@wtgee and @AnthonyHorton, your thoughts please.

jamessynge:
I was reviewing @AnthonyHorton's change which adds a pyro config file, and that includes a path starting with /var/pantoptes. It seems to me that we should try to avoid duplicating the paths.

So, I'd like to either extend our YAML support with the ability to evaluate env vars, or reduce the use of the env vars... or both if there are some that really aren't needed.

wtgee:
A mapping for directories specifically? Might be a good idea to pull that out of pocs.yaml and put directly in $PANDIR.

AnthonyHorton:
I'm supportive of the idea of removing the dependence on environment variables, and instead putting the items we really need in the config files and inferring the rest. On general principles keeping all the important configuration in the `conf_files` directory, rather than having some of it in user's `.bashrc`, `.profile`, etc. seems like a good idea to me.

FWIW the setting of `base: /var/panoptes` in the `pyro_camera.yaml` file that I added isn't a new thing. The `directories` section of that file is just a copy of the `directories` section in `pocs.yaml`. We've had a situation for a long time were we have an environment variable for something (`PANDIR`) as well as a config entry for the same thing (`config['directories']['base']`), which is obviously a recipe for confusing errors if they ever have different values. We should definitely be setting these things in only one place, & my preference would be the config files.

Having `$POCS`, `$PANDIR`, etc. defined in your shell can be handy, but of course it will still be possible to do that even if POCS itself no longer uses the environment variables. We could make that easy by having a script that parses the config files and sets the corresponding environment variables for you, or adds them to your `.profile`.

In terms of what to include `$PANDIR` and `$PANUSER` (or their config file equivalents) seem to be the only essential ones.

wtgee:
Huntsman runs everything as `huntsman` user under `/var/huntsman`. Works great to just set PANUSER and PANDIR. I believe our travis testing also uses PANUSER.

I've started stripping a few env variables in the PRs I am working on.

I've added #564 with some other config related items I've been thinking about. Please add anything to that overview issue.

wtgee:
[...]

PR #628, opened 2018-09-24
Observation cleanup cleanup

wtgee:
Updates to image uploading

* Option to not remove JPGs.
* Option to overwrite timelapse.
* Metadata storage pulled from config
* Script defaults to False for uploading of images and metadata
* Bash script to perform generic uploading
	* :star2:  Adds a manifest log so uploads can be safely resumed/retried. :star2: 

Note: We are currently using the underlying `gsutil cp` mechanism. This was switched (a long time ago) from [`gsutil rsync`](https://cloud.google.com/storage/docs/gsutil/commands/rsync). Might be worth exploring if there are advantages to using that instead.

> **Note:** ~~The cloudsql proxy handles the connections to the server instance but we still need passwords for the actual database users. Not sure how to handles this other than have more env variables (#555). Could feasibly set up a cloud-based REST api so this script just sends json entries to some public endpoint, which handles the actual db connection. This would also save us having to actually run the proxy. For the most part we are inserting/updating and don't really need to read the db.~~

I'm now relying on the [`~/.pgpass` file](https://www.postgresql.org/docs/9.4/static/libpq-pgpass.html) being set by the user. We need to still figure out a mechanism to let people know what the password is.

> I've also added an entry in the config for the service_account_key location. We could standardize the location and name of the file and get rid of this. 

Part of #599, ~~(sort of)~~ depends on #626.


wtgee:
I've been using this on PAN001 for a few days now and is working well, including uploading of metadata to our cloudsql instance. Need to figure out how we want to distribute passwords but otherwise I think it's looking good.

conf_files/pocs.yaml
```diff
@@ -66,6 +66,22 @@ messaging:
     cmd_port: 6500
     msg_port: 6510
 
+########################## Observations ########################################
+# Observations consist of a sequence of images and are grouped together
+# according # to the time the sequence started and finish when the mount
```
jamessynge: `#` in the middle of the line.

conf_files/pocs.yaml
```diff
@@ -66,6 +66,22 @@ messaging:
...
+# slews to new coordinates. There will be one folder per camera per target
+# per observation. An example folder structure would be:
+#
+#   $PANDIR/images/fields/PAN001/Hd189733/14d3bd/20180901T120001/
+#
+# In this folder will be stored JPG and FITS images. A timelapse of the
+# observation can be made (on per camera) and the JPGs optionally removed
```
jamessynge: Does this imply that constructing the timelapse requires the JPEG images?

conf_files/pocs.yaml
```diff
@@ -66,6 +66,22 @@ messaging:
     cmd_port: 6500
     msg_port: 6510
 
+########################## Observations ########################################
+# Observations consist of a sequence of images and are grouped together
```
jamessynge: This sentence feels a bit awkward to me. How would this be:

An observation folder contains a contiguous sequence of images of a target/field recorded by a single camera, with no slewing of the mount during the sequence; there may be tracking adjustments during the observation.

conf_files/pocs.yaml
```diff
@@ -66,6 +66,22 @@ messaging:
...
+# according # to the time the sequence started and finish when the mount
+# slews to new coordinates. There will be one folder per camera per target
+# per observation. An example folder structure would be:
+#
+#   $PANDIR/images/fields/PAN001/Hd189733/14d3bd/20180901T120001/
```
jamessynge: I don't think the unit id is in the path.

conf_files/pocs.yaml
```diff
@@ -74,19 +90,20 @@ messaging:
 # details about authentication.
 ################################################################################
 panoptes_network:
-    image_storage: True
-    metadata_storage: True
+    image_storage: False
```
jamessynge: Please add comments explaining these True/False fields.

conf_files/pocs.yaml
```diff
@@ -66,6 +66,22 @@ messaging:
...
+# slews to new coordinates. There will be one folder per camera per target
+# per observation. An example folder structure would be:
+#
+#   $PANDIR/images/fields/PAN001/Hd189733/14d3bd/20180901T120001/
+#
+# In this folder will be stored JPG and FITS images. A timelapse of the
+# observation can be made (on per camera) and the JPGs optionally removed
+# afterward.
+################################################################################
+observations:
```
jamessynge: TODO: Add some fields to control cleanup in the future (e.g. after 30 days, if the images have been uploaded and the disk has less than 30GB free, then remove the oldest images until we have at least 30GB free).

pocs/observatory.py
```diff
@@ -199,43 +198,82 @@ def get_observation(self, *args, **kwargs):
 
         return self.current_observation
 
-    def cleanup_observations(self):
+    def cleanup_observations(self, timeout=600):
```
jamessynge: FWIW, cleanup doesn't seem the correct term. The main focus of this method seems to be uploading.

pocs/observatory.py
```diff
@@ -199,43 +198,82 @@ def get_observation(self, *args, **kwargs):
...
         """Cleanup observation list
 
         Loops through the `observed_list` performing cleanup tasks. Resets
         `observed_list` when done
 
+        Args:
+            timeout (int, optional): Timeout for each directory to upload files,
+                defaults to 600 seconds (10 minutes).
```
jamessynge: The slowest speed that will work for uploading a full night's worth of data within 24 hours is about 128Kbps (i.e. a modest speed DSL connection). I think that we should use that as our measure of whether it is uploading fast enough. For now, I'd just suggest increasing that default to at least an hour, but later we could use a continuously running process which uploads files that it is informed about over zeromq, and reports progress the same way. That way we're not using Popen.communicate's crude timeout technique.

pocs/utils/db/postgres.py
```diff
@@ -141,25 +148,38 @@ def add_header_to_db(header, conn=None, logger=None):
         'exp_time': header['EXPTIME'],
         'ra_rate': header['RA-RATE'],
         'pocs_version': header['CREATOR'],
-        'piaa_state': header['PSTATE'],
+        'piaa_state': header.get('PSTATE', 'initial'),
```
jamessynge: Why are we using get sometimes, and sometimes not?

scripts/connect_clouddb_proxy.py
```diff
@@ -16,16 +16,22 @@
 from pocs.utils.config import load_config
 
 
-def main(instance_id, local_port, verbose=False):
+def main(instances, key_file, verbose=False):
```
jamessynge: FWIW, there are inconsistent file names involved here: cloud_sql_proxy and connect_clouddb_proxy.py (i.e. _sql vs. db).

scripts/connect_clouddb_proxy.py
```diff
@@ -41,10 +47,14 @@ def main(instance_id, local_port, verbose=False):
...
-    parser.add_argument('-d', '--database', default=None, required=True,
-                        help="Database, currently 'meta' or 'tess' from the config.")
-    parser.add_argument('-v', '--verbose', action='store_true', default=False,
-                        help="Print results to stdout")
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--from_config', default=True, action='store_true',
+                       help="Connect to all instances listed in the config file, default True.")
+    group.add_argument('--database', default=None,
+                       help="Connect to a specific database listed in the config.")
+    parser.add_argument('--key_file', default=None, help="JSON service account key location.")
```
jamessynge: Maybe we should settle on a default location, perhaps baked into pocs.yaml, such as `$HOME/.pocs_secrets/cloud_sql_proxy.key`.

scripts/connect_clouddb_proxy.py
```diff
@@ -53,22 +63,46 @@ def main(instance_id, local_port, verbose=False):
...
+
+    # Try to lookup service account key from config if none provided
+    key_file = args.key_file
+    if not key_file:
+        try:
+            key_file = network_config['service_account_key']
+        except KeyError:
+            pass
+
+        if not key_file or not os.path.isfile(key_file):
```
jamessynge: Extra indentation.

scripts/connect_clouddb_proxy.py
```diff
@@ -41,10 +47,14 @@ def main(instance_id, local_port, verbose=False):
...
     parser = argparse.ArgumentParser(description="Connect to a google CloudSQL via a local proxy")
-    parser.add_argument('-d', '--database', default=None, required=True,
-                        help="Database, currently 'meta' or 'tess' from the config.")
-    parser.add_argument('-v', '--verbose', action='store_true', default=False,
-                        help="Print results to stdout")
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--from_config', default=True, action='store_true',
+                       help="Connect to all instances listed in the config file, default True.")
+    group.add_argument('--database', default=None,
+                       help="Connect to a specific database listed in the config.")
```
jamessynge: ... else will connect to all instances listed in the config.

scripts/connect_clouddb_proxy.py
```diff
@@ -41,10 +47,14 @@ def main(instance_id, local_port, verbose=False):
...
     # Get the command line option
     parser = argparse.ArgumentParser(description="Connect to a google CloudSQL via a local proxy")
-    parser.add_argument('-d', '--database', default=None, required=True,
-                        help="Database, currently 'meta' or 'tess' from the config.")
-    parser.add_argument('-v', '--verbose', action='store_true', default=False,
-                        help="Print results to stdout")
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--from_config', default=True, action='store_true',
+                       help="Connect to all instances listed in the config file, default True.")
+    group.add_argument('--database', default=None,
```
jamessynge: Now called instance in the config. Do you want to make those consistent?

scripts/upload_files.sh
```diff
@@ -0,0 +1,55 @@
...
+SEARCH_PATH="$1"
+DESTINATION="$2"
+MANIFEST_PATH="$3"
+echo 'Uploading files'
+echo "SEARCH_PATH = ${SEARCH_PATH}"
+echo "DESTINATION = ${DESTINATION}"
+echo "MANIFEST_PATH = ${MANIFEST_PATH}"
+
+# Loops until all files are uploaded. See `gsutil cp` documentation.
+until gsutil -mq cp -c -L "${MANIFEST_PATH}" -r "${SEARCH_PATH}" "${DESTINATION}"; do
```
jamessynge: I've never seen `until` in bash before. Interesting.

scripts/upload_image_dir.py
```diff
@@ -1,97 +1,101 @@
...
+    config = load_config()
+    try:
+        pan_id = config['pan_id']
+    except KeyError:
+        raise error.GoogleCloudError("Can't upload without a valid pan_id in the config")
 
-    return len(imgs)
+    _print("Cleaning observation directory: {}".format(directory))
+    try:
+        clean_observation_dir(directory, remove_jpgs=remove_jpgs, include_timelapse=make_timelapse,
```
jamessynge: Interesting order of cleanup and upload. I suppose we could do with choices such as:
- Create timelapse
- Create jpg
- Upload fits
- Upload jpg
- Upload timelapse
- Extra upload pattern/files
- Exclude upload pattern/files
- Remove jpg
- Remove timelapse
- Remove fits
- Remove extra pattern/files
- Remove empty directory

And so on. For super small disk systems, we'd want the ability to discard the fits file as soon as it was used to generate the jpg and had been uploaded.

I'm not really asking for anything at this time. Just a bit worried. Maybe we simply need to make sure that we create a very clean separation between the robotic telescope (capturing of data, tracking the sky) and the data processing/upload system.

jamessynge:
Regarding pgpass, I'm a bit concerned in general about this exposure of access to the db. Since the point is for a unit to upload data and metadata, I'd rather expose a simple API for doing so rather than exposing read-write access to the db to all units. Thoughts?

pocs/observatory.py
wtgee: The cleanup is the main purpose of this method, namely to remove old files and to fpack the FITS files. The making of a timelapse and uploading are optional to that. Cleanup always has to happen, uploading does not.

pocs/utils/db/postgres.py
wtgee: When I want to set a default if not provided. The other cases should probable be handled better.

scripts/connect_clouddb_proxy.py
wtgee: That was somewhat intentional since I think users might not know what "cloud_sql" refers to but could figure out "cloud_db". I'll make consistent though.

scripts/connect_clouddb_proxy.py
wtgee: Yes, this is described here: https://github.com/panoptes/POCS/tree/develop/pocs/utils/google#service-account-key

scripts/connect_clouddb_proxy.py
wtgee: The instance is the postgres instance that holds the different databases. The name they pass here is the name for the actual database (actually a shortened version and could be anything).

We should be able to hide the `instance` details from the user and `database` is a lot more intuitive.

scripts/upload_files.sh
wtgee: I thought that was an interesting usage. Working well so far.

scripts/upload_image_dir.py
wtgee: Yes, I will need to make something some time regarding FITS storage as the Pi doesn't have a lot.

wtgee:
> Regarding pgpass, I'm a bit concerned in general about this exposure of access to the db. Since the point is for a unit to upload data and metadata, I'd rather expose a simple API for doing so rather than exposing read-write access to the db to all units. Thoughts?

I tend to agree. After using this fora  few days it's working well but I don't think the connection is a great idea. Some of this was waiting on the Cloud Functions to mature a bit and yesterday I started playing with that exact idea of an API endpoint that we could just post the headers to in json format.

As such, this PR might change or close to some extent.

scripts/upload_image_dir.py
jamessynge: I've been assuming that a Pi will either require high-speed internet access or a local disk (USB connected). On possibility for a Pi system is to omit creating JPGs and timelapse files locally, and only do that in the cloud. Requires a strong separation between PAWS (including JPG and timelapse creation) and POCS (operating the observatory, without a UI, just an API, maybe a CLI).

jamessynge:
Maybe the PR has gotten too big. Perhaps we need to break it up into semantically separated parts, perhaps after a discussion in an issue?

wtgee:
Closing this as scope has changed. See #599 for details.

PR #630, opened 2018-09-25
Fixing observing/pointing wait events

wtgee:
Removing units from the MAX_EXTRA_TIME for consistency.

This is fixing a bug introduced in #625.

@jamessynge note that this is fixing those problems with testing the flaws in our testing I was talking about. The bug was causing a shutdown in the pointing state but most of the tests in `test_pocs` report as passing because POCS itself was absorbing the errors.

### DIFF
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -3,6 +3,7 @@ import pytest
 
 import time
 from datetime import datetime as dt
+from astropy import units as u
 
 from pocs.utils import current_time
 from pocs.utils import listify
@@ -69,6 +70,10 @@ def test_countdown_timer_non_blocking():
     timer = CountdownTimer(0)
     assert timer.is_non_blocking
     assert timer.time_left() == 0
+
+    for arg, expected_duration in [(2, 2.0), (0.5, 0.5), (1 * u.second, 1.0)]:
+        timer = CountdownTimer(arg)
+        assert timer.duration == expected_duration
 
 
 def test_countdown_timer():
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -1,12 +1,13 @@
 import os
 import pytest
 
+import time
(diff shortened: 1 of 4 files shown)


pocs/tests/utils/test_utils.py
```diff
@@ -51,3 +52,30 @@ def test_has_camera_ports():
...
+    assert timer.is_non_blocking
+    assert timer.time_left() == 0
+
+
+def test_countdown_timer():
+    timer = CountdownTimer(1)
+    assert timer.time_left() > 0
+    assert timer.expired() is False
+    assert timer.is_non_blocking is False
+    time.sleep(1)
```
jamessynge: I'd prefer a sleep time that is greater than the count down time. Alternately, you could have a loop that checks every 0.01 seconds (for example), and tracks whether the time_left is decreasing each time, and goes to zero before 1.1 seconds (for example) has passed.

pocs/tests/utils/test_utils.py
```diff
@@ -51,3 +52,30 @@ def test_has_camera_ports():
...
+        assert CountdownTimer('d')
+
+    with pytest.raises(ValueError):
+        assert CountdownTimer(current_time())
+
+    with pytest.raises(AssertionError):
+        assert CountdownTimer(-1)
+
+
+def test_countdown_timer_non_blocking():
```
jamessynge: Could you add tests with float and u.second? For example:
```
for arg, expected_duration in [(2, 2.0), (0.5, 0.5), (1 * u.second, 1.0)]:
    timer = CountdownTimer(arg)
    assert timer.duration == expected_duration
...
```

pocs/tests/utils/test_utils.py
wtgee: Good idea. I update this, let me know if you think still needs improvement.

pocs/tests/utils/test_utils.py
wtgee: Done, thanks.

PR #642, opened 2018-09-29
Camera image pathname cleanup

### LINKED ISSUE #561: Observation should have dir_name property
In a number of places where we build a filename using an `Observation` we are doing this kid of thing:

```
        image_dir = "{}/fields/{}/{}/{}/".format(
            self.config['directories']['images'],
            observation.field.field_name,
            self.uid,
            observation.seq_time,
        )
```

Would be nicer to have something like:

```
image_dir = os.path.join([
    observation.dir_name,
    self.uid,
    observation.seq_time
])
```

or even

```
image_dir = observation.get_dir(uid=self.uid)
```

It would also be nice to later have a stored copy of the directory, which would make cleanup easier.

### LINKED ISSUE #563: Image paths have double slash
Notice the double slash between the folder name and the image name.
```
D0826 08:15:35.939 canon_gphoto2.py:154 Converting CR2 -> FITS: /var/panoptes/images/fields/Kic8462852/6575fc/2018082
6T045541//20180826T081329.cr2
D0826 08:15:35.952 canon_gphoto2.py:154 Converting CR2 -> FITS: /var/panoptes/images/fields/Kic8462852/7bab97/2018082
6T045541//20180826T081329.cr2
```

wtgee:
This gives a `directory` property to the `Observation`, which makes it easier for the camera to build path names for the images. Also just generic cleanup on using `os.path.join` instead of string formatting.

Close #561 
Fixes #563 

Some overlap with #547


pocs/camera/camera.py
```diff
@@ -423,25 +418,31 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
 
         start_time = headers.get('start_time', current_time(flatten=True))
 
+        if not observation.seq_time:
```
jamessynge: FWIW, it seems to me that cameras should have no knowledge of observations; e.g. a camera should get asked to take one image and put it in a specific file, nothing more.

pocs/scheduler/observation.py
```diff
@@ -107,6 +109,24 @@ def seq_time(self):
     def seq_time(self, time):
         self._seq_time = time
 
+    @property
+    def directory(self):
+        """Return the directory for this Observation.
+
+        This return the base directory for the Observation. This does *not* include
```
jamessynge: This comment makes perfect sense. Given that, and my previous comment, it would seem that Observatory or POCS would be responsible for taking the observation directory and creating the path for a specific camera.

pocs/camera/camera.py
wtgee: To some extent I agree and think this could be re-worked, but this is part of the `take_observation` method, which necessitates an `observation` be present. The lower-level `take_exposure` method does not require this knowledge and instead requires a filename be passed.

pocs/scheduler/observation.py
wtgee: In addition to the comments above, the camera could feasibly take an "observation" without the use of POCS or an Observatory. There is a `test_observation` in `test_cameras.py` that does just this. Not sure how useful that is however.

pocs/camera/camera.py
jamessynge: Perhaps a TODO?

wtgee: Issue added here: https://github.com/panoptes/POCS/issues/656

PR #644, opened 2018-09-30
Don't use the astropy unit

wtgee:
This indicates we are still having the problem where a failure in the movement through the POCS states isn't registering as a test failure.

### DIFF
--- a/pocs/state/states/default/observing.py
+++ b/pocs/state/states/default/observing.py
@@ -13,7 +13,7 @@ def on_enter(event_data):
     pocs.next_state = 'parking'
 
     try:
-        maximum_duration = pocs.observatory.current_observation.exp_time + MAX_EXTRA_TIME
+        maximum_duration = pocs.observatory.current_observation.exp_time.value + MAX_EXTRA_TIME
 
         # Start the observing.
         camera_events_info = pocs.observatory.observe()


pocs/state/states/default/observing.py
```diff
@@ -13,7 +13,7 @@ def on_enter(event_data):
     pocs.next_state = 'parking'
 
     try:
-        maximum_duration = pocs.observatory.current_observation.exp_time + MAX_EXTRA_TIME
+        maximum_duration = pocs.observatory.current_observation.exp_time.value + MAX_EXTRA_TIME
```
jamessynge: How about adding a helper method to the observation that returns numeric seconds (e.g. does a conversion of the time to u.second before extracting .value)?

wtgee: I want to incorporate the idea of a readout time, which makes it seem like it should be that of the camera rather than the observation.

Going to merge this and think about other issue.

AnthonyHorton: The non-gphoto2 Camera classes already have a `readout_time` property, though it's currently not used for much.

wtgee: > The non-gphoto2 Camera classes already have a `readout_time` property, though it's currently not used for much.

Yes, that's what I was thinking of. If I add anything it will be an expansion of that existing property.

Issue #645, opened 2018-09-30
Make sure testing is covering entire state machine


wtgee:
During testing of POCS we are having errors in the states that the pocs instance is handling. Since the tests just look for conditions at startup and shutdown of pocs it fails to recognize that a test is not exercising the code completely. Mostly revealed through coverage tests.

PR #646, opened 2018-09-30
Add the cameras to the POCS tests

wtgee:
This was missed during #612 but is also indicative of #645.


wtgee:
That's the kind of test coverage bump I want...

wtgee:
> Can't really tell what the effect of this change is.

This was merely a fix that should have been put in place with #612 (the camera dependency injection). Without this change the Observatory object is not getting passed any cameras so was erroring.

Which, now that I write that, means that we should have a check in the Observatory such that it cannot leave the `ready` state if there are no cameras attached.

PR #648, opened 2018-09-30
Small scheduler fixes

wtgee:
* Scheduler file was being read twice.
* Output to show why field was invalid if cannot add.
* Setting a field list clobbers a file list so do it first.

Edit: **Behavior Change**: I added a small change so that the fields file is only read upon scheduler initialization if `config['scheduler']['check_file']=False`. So read at beginning if you only want to read once, otherwise read each time call is made to `get_observation`.


pocs/scheduler/scheduler.py
```diff
@@ -40,8 +40,8 @@ def __init__(self, observer, fields_list=None, fields_file=None,
 
         assert isinstance(observer, Observer)
 
-        self._fields_file = fields_file
         self._fields_list = fields_list
+        self._fields_file = fields_file
```
jamessynge: If the order matters here (as apparently it does), please add a comment.

pocs/observatory.py
```diff
@@ -242,13 +242,14 @@ def get_observation(self, *args, **kwargs):
         self.logger.debug("Getting observation for observatory")
 
         # If observation list is empty or a reread is requested
+        reread_fields_file = False
         if (self.scheduler.has_valid_observations is False or
                 kwargs.get('reread_fields_file', False) or
                 self.config['scheduler'].get('check_file', False)):
-            self.scheduler.read_field_list()
+            reread_fields_file = True
```
jamessynge: Any reason that this isn't:
```
        reread_fields_file = (
            self.scheduler.has_valid_observations is False or
            kwargs.get('reread_fields_file', False) or
            self.config['scheduler'].get('check_file', False))

Comment (wtgee) [reply pocs/observatory.py]: Yes, much nicer, thanks.

File: pocs/scheduler/scheduler.py [hunk 3]
```diff
@@ -40,6 +40,8 @@ def __init__(self, observer, fields_list=None, fields_file=None,
 
         assert isinstance(observer, Observer)
 
+        # Settig the fields_list directly will clobber anything
```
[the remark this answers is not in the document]
wtgee: Oops, too eager on the merge. I'll get it changed.

PR #654, opened 2018-10-01
Add manifest log for image uploads

wtgee:
* Adds a manifest log so that uploads can be resumed/retried if fail and will also not re-upload if upload has succeeded in the past.

This is pulled from #628 and is a part of #599.


pocs/utils/google/storage.py
```diff
@@ -317,6 +317,8 @@ def upload_observation_to_bucket(pan_id,
     """
     assert os.path.exists(dir_name)
     assert re.match(r'PAN\d\d\d', pan_id) is not None
+    # Note: this just makes sure the command exists, it is called from bash file.
```
jamessynge: Doesn't of course ensure, since it is only an assert. If you want these asserts to be meaningful, how about converting them to raise an exception?

scripts/upload_files_to_bucket.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash -e
```
jamessynge: Any reason to not use python? Then we could have flags instead of positional args.

wtgee: This came as part of #628, which included a python wrapper around this shell script. The cleanup script is still forthcoming so I expect it will be the same. 

I guess we could cut out the middle-man and just call the bash part in a subprocess directly.

wtgee: I'll make a note in #599.

PR #655, opened 2018-10-01
Updates to cleanup

wtgee:
* Timelapse optional (default True).
* Removing jpgs optional (default False).


wtgee:
> I continue be concerned about the number of changes that have been going in without tests.

I could add some to this for sure. Will do before merge.

PR #659, opened 2018-10-02
Separate coarse autofocus from fine autofocus

AnthonyHorton:
Change in behaviour for autofocus so that a coarse focus is not automatically followed followed by a fine focus. This is required for the changes planned for https://github.com/AstroHuntsman/huntsman-pocs/issues/69

Also incorporates some improvements to the autofocus tests (checking for created files), and fixes a bug introduced by #621 that prevented the final focus position image being taken when plotting was disabled, which is now the default.


pocs/camera/camera.py
```diff
@@ -300,43 +300,41 @@ def autofocus(self,
...
                 saturated pixel mask (determine size of masked regions), default 10
-            coarse (bool, optional): Whether to begin with coarse focusing,
-                default False
-            plots (bool, optional: Whether to write focus plots to images folder,
-                default False.
-            blocking (bool, optional): Whether to block until autofocus complete,
-                default False
+            coarse (bool, optional): Whether to perform a coarse focus, otherwise will perform
+                a fine focus. Default False.
+            plots (bool, optional: Whether to write focus plots to images folder, default False.
```
wtgee: Would prefer to see this called `make_plots` but we could make a separate issue to change things like that all at once.

pocs/focuser/focuser.py
```diff
@@ -294,35 +293,16 @@ def autofocus(self,
...
-        focus_params['coarse'] = False
-
-        fine_thread = Thread(target=self._autofocus, kwargs=focus_params)
-        fine_thread.start()
-
+        focus_thread = Thread(target=self._autofocus, kwargs=focus_params)
+        focus_thread.start()
         if blocking:
-            fine_event.wait()
+            focus_event.wait()
```
wtgee: Should include some ultimate timeout in the `wait`

pocs/focuser/focuser.py
```diff
@@ -490,14 +463,13 @@ def _autofocus(self,
 
         final_focus = self.move_to(best_focus)
 
+        final_fn = "{}_{}.{}".format(final_focus, "final", self._camera.file_extension)
```
wtgee: Should this include the `focus_type`? I don't seem to see it set anywhere.

AnthonyHorton: That's a good idea. It should. I'll put that in.

pocs/focuser/focuser.py
AnthonyHorton: Yes. Needs to be a rather conservative one though, as it's a bit involved to predict exactly how long it will take.

AnthonyHorton: That might be a job for another PR actually. Want to add timeouts to the camera's threaded operations more generally to prevent the lock ups we sometimes get.

pocs/camera/camera.py
AnthonyHorton: OK. Yes, there are probably quite a few of these sorts of terminology inconsistencies to be fixed.

PR #668, opened 2018-10-03
Observation current_exp change

wtgee:
Changed to `current_exp_num` to be more consistent.

This is now just a count of the `exposure_list`, which means we don't need to do specific bookkeeping.

Adds a change to actually reset `exposure_list` when `reset` is called.

### DIFF
--- a/pocs/scheduler/observation.py
+++ b/pocs/scheduler/observation.py
@@ -66,7 +66,7 @@ class Observation(PanBase):
         self._min_duration = self.exp_time * self.min_nexp
         self._set_duration = self.exp_time * self.exp_set_size
 
-        self.pointing_image = None
+        self.pointing_images = OrderedDict()
 
         self._image_dir = self.config['directories']['images']
         self._seq_time = None
@@ -159,6 +159,19 @@ class Observation(PanBase):
             return list(self.exposure_list.items())[-1]
         except IndexError:
             self.logger.warning("No exposure available")
+
+    @property
+    def pointing_image(self):
+        """Return the last pointing image.
+
+        Returns:
+            tuple: `image_id` and full path of most recent pointing image from
+                the primary camera.
+        """
+        try:
+            return list(self.pointing_images.items())[-1]
+        except IndexError:
+            self.logger.warning("No pointing image available")
 
 
 ##################################################################################################
--- a/pocs/scheduler/observation.py
+++ b/pocs/scheduler/observation.py
@@ -60,13 +60,12 @@ class Observation(PanBase):
         self.min_nexp = min_nexp
         self.exp_set_size = exp_set_size
         self.exposure_list = OrderedDict()
+        self.pointing_images = OrderedDict()
 
         self.priority = float(priority)
 
         self._min_duration = self.exp_time * self.min_nexp
         self._set_duration = self.exp_time * self.exp_set_size
-
-        self.pointing_images = OrderedDict()
 
         self._image_dir = self.config['directories']['images']
         self._seq_time = None
--- a/pocs/scheduler/observation.py
+++ b/pocs/scheduler/observation.py
@@ -73,6 +73,8 @@ class Observation(PanBase):
 
         self.merit = 0.0
 
+        self.reset()
+
         self.logger.debug("Observation created: {}".format(self))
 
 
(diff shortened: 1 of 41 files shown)


wtgee:
This is hanging every time in the `test_run_complete`, but only when done on travis. I'm tempted to close as there is no clear need for this change but I would like to discover why it is failing. Leaving open for a few days.

pocs/scheduler/observation.py
```diff
@@ -157,10 +164,10 @@ def last_exposure(self):
 ##################################################################################################
 
     def reset(self):
```
jamessynge: It would be great if this method were called by `__init__`, so that they can't get out of sync.

pocs/scheduler/observation.py
```diff
@@ -127,6 +125,15 @@ def directory(self):
                                            self.field.field_name)
             return self._directory
 
+    @property
+    def current_exp_num(self):
```
jamessynge: What if we're currently taking another exposure? Is that counted or not?

pocs/scheduler/observation.py
```diff
@@ -127,6 +125,15 @@ def directory(self):
...
+        """ Return the current number of exposures.
```
jamessynge: Style question: why the leading space before the start of the docstring contents (i.e. the space before Return).

pocs/state/states/default/observing.py
```diff
@@ -27,7 +27,6 @@ def on_enter(event_data):
         pocs.logger.warning("Problem with imaging: {}".format(e))
         pocs.say("Hmm, I'm not sure what happened with that exposure.")
     else:
-        pocs.observatory.current_observation.current_exp += 1
         pocs.logger.debug('Finished with observing, going to analyze')
```
jamessynge: Drop the blank line?

wtgee:
> This is hanging every time in the `test_run_complete`, but only when done on travis. I'm tempted to close as there is no clear need for this change but I would like to discover why it is failing. Leaving open for a few days.

The hang here is coming from the camera simulator trying to use the same image each time. On my local I now get:

```
D1005 04:26:48.701 Taking 2.0 second exposure on Cam01: /var/panoptes//images/fields/Kic8462852/SC0910/20160909T080000/20160909T080000.fits              
E1005 04:26:50.721 Error writing image to /var/panoptes//images/fields/Kic8462852/SC0550/20160909T080000/20160909T080000.fits!
E1005 04:26:50.721 Error writing image to /var/panoptes//images/fields/Kic8462852/SC0910/20160909T080000/20160909T080000.fits!
E1005 04:26:50.721 File '/var/panoptes//images/fields/Kic8462852/SC0550/20160909T080000/20160909T080000.fits' already exists.
E1005 04:26:50.721 File '/var/panoptes//images/fields/Kic8462852/SC0910/20160909T080000/20160909T080000.fits' already exists.
```

and it loops forever. Will work on getting a fix for it. Will be somewhat related to other camera tests.

pocs/scheduler/observation.py
jamessynge: I've not seen that style before. Where does it come from? It it widely used in the python community? In astropy?

jamessynge: And looking at PEP8 (https://www.python.org/dev/peps/pep-0008/?#documentation-strings), it doesn't use that style:
```
PEP 257 describes good docstring conventions. Note that most importantly, the """ that ends a multiline docstring should be on a line by itself:

"""Return a foobang

Optional plotz says to frobnicate the bizbaz first.
"""
```

I followed the link to PEP 257, which also doesn't contain any example of a leading space.

wtgee: I use this: https://packagecontrol.io/packages/AutoDocstring

Makes docstring maintenance much nicer.

jamessynge:
Regarding test_run_complete, I can't help but note the // (double slash) in the paths. Fixing that probably won't help... but might!

wtgee:
> Regarding test_run_complete, I can't help but note the // (double slash) in the paths. Fixing that probably won't help... but might!

Both you and @AnthonyHorton pointed it out so I would be a fool to ignore fixing it!

pocs/scheduler/observation.py
wtgee: Yeah, I checked my [usual resource](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) and don't see it anywhere. I'll make sure it's gone going forward.

PR #669, opened 2018-10-03
Minor debug log output changes

wtgee:
* Removing debug statement that was used for testing. Currently it defies the whole point of having the `msg_interval` if we just print debug items anyway.
* Get rid of silly precision in weather reading:
    ```
    Weather Safety: True [74 sec old - 2018-10-03 06:13:07.759000]
    ```
    to
    ```
    Weather Safety: True [74 sec old - 2018-10-03 06:13:07]
    ```

### DIFF
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -372,8 +372,10 @@ class POCS(PanStateMachine, PanBase):
             timestamp = record['date'].replace(tzinfo=None)  # current_time is timezone naive
             age = (current_time().datetime - timestamp).total_seconds()
 
-            self.logger.debug(
-                "Weather Safety: {} [{:.0f} sec old - {}]".format(is_safe, age, timestamp))
+            self.logger.debug("Weather Safety: {} [{:.0f} sec old - {:%d-%m-%Y %H:%M:%S}]",
+                              is_safe,
+                              age,
+                              timestamp)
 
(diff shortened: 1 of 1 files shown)


wtgee:
That's a funny coverage reading. I decrease the amount of lines in the file but don't change the number of lines covered, just the overall percentage.

pocs/core.py
```diff
@@ -372,8 +372,10 @@ def is_weather_safe(self, stale=180):
             timestamp = record['date'].replace(tzinfo=None)  # current_time is timezone naive
             age = (current_time().datetime - timestamp).total_seconds()
 
-            self.logger.debug(
-                "Weather Safety: {} [{:.0f} sec old - {}]".format(is_safe, age, timestamp))
+            self.logger.debug("Weather Safety: {} [{:.0f} sec old - {:%d-%m-%Y %H:%M:%S}]",
```
AnthonyHorton: Wot no ISO 8601? ;-)

Shouldn't it be `{:%Y-%m-%d %H:%M:%S}` to format the date-time as in the PR description?

wtgee: Uh, yes, it should be. I was focus on the microseconds. :)

wtgee:
I'm not sure why this PR isn't updating but the tests have passed.

PR #670, opened 2018-10-03
Camera test tweaks from #547

AnthonyHorton:
Some minor tweaks to the camera tests taken from #547. They're not really distributed camera specific so I've pulled them out here to simplify that monster PR a bit.

### DIFF
--- a/pocs/tests/test_camera.py
+++ b/pocs/tests/test_camera.py
@@ -307,7 +307,7 @@ def test_observation(camera, images_dir):
     """
     field = Field('Test Observation', '20h00m43.7135s +22d42m39.0645s')
     observation = Observation(field, exp_time=1.5 * u.second)
-    observation.seq_time = 'seq_time'
+    observation.seq_time = '19991231T235959'
     camera.take_observation(observation, headers={})
     time.sleep(7)
     observation_pattern = os.path.join(images_dir, 'fields', 'TestObservation',
--- a/pocs/tests/test_camera.py
+++ b/pocs/tests/test_camera.py
@@ -33,14 +33,14 @@ def images_dir(tmpdir_factory):
 @pytest.fixture(scope='module', params=zip(params, ids), ids=ids)
 def camera(request, images_dir):
     if request.param[0] == SimCamera:
-        camera = request.param[0](focuser={'model': 'simulator',
-                                           'focus_port': '/dev/ttyFAKE',
-                                           'initial_position': 20000,
-                                           'autofocus_range': (40, 80),
-                                           'autofocus_step': (10, 20),
-                                           'autofocus_seconds': 0.1,
-                                           'autofocus_size': 500,
-                                           'autofocus_keep_files': False})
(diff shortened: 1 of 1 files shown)


pocs/tests/test_camera.py
```diff
@@ -297,14 +301,18 @@ def test_exposure_not_connected(camera):
...
 
 
-def test_observation(camera):
+def test_observation(camera, images_dir):
     """
     Tests functionality of take_observation()
     """
     field = Field('Test Observation', '20h00m43.7135s +22d42m39.0645s')
     observation = Observation(field, exp_time=1.5 * u.second)
+    observation.seq_time = 'seq_time'
```
wtgee: I don't think we rely on any pattern match anywhere for the seq_time but this seems a little odd to just call it 'seq_time'. Why not a dummy pattern that looks like a datestamp?

pocs/tests/test_camera.py
```diff
@@ -297,14 +301,18 @@ def test_exposure_not_connected(camera):
...
     camera.take_observation(observation, headers={})
     time.sleep(7)
+    observation_pattern = os.path.join(images_dir, 'fields', 'TestObservation',
```
wtgee: Note that you can now 

```
os.path.join(
    observation.directory, 
    camera.uid, 
    observation.seq_time,
    '*.fits*'
)
```
Unless the `images_dir` from the fixture is pointing somewhere else.

pocs/tests/test_camera.py
```diff
@@ -68,7 +68,7 @@ def camera(request, images_dir):
 
 
 @pytest.fixture(scope='module')
-def counter():
+def counter(camera):
```
wtgee: Does this make any functional changes or fixes?

pocs/tests/test_camera.py
AnthonyHorton: Good point.

pocs/tests/test_camera.py
AnthonyHorton: `images_dir` here is the `/tmp/pytest-of-<username>/pytest-N` directory that comes from the temporary directory fixture. I'll have to have a think about whether `observation.directory` would be the same here.

pocs/tests/test_camera.py
AnthonyHorton: I did this so that there are separate counters for each `camera`. This doesn't have an effect in `develop` right now as (usually) the tests only run with the simulated camera, but it is necessary for the tests to run in `pyro-cameras` (where both simulated and simulated distributed cameras are tested) and would also be required if you were running the tests with real SBIG or FLI cameras attached and configured.

pocs/tests/test_camera.py
AnthonyHorton: I'm pretty sure those left over images are coming from the tests in `test_observatory.py`, otherwise these tests (which are checking for FITS files appearing in the temporary directory) would be failing.

PR #620, opened 2018-09-22
Camera changes, mostly cleanup

wtgee:
Some from #547, some from latest Huntsman run. These are mostly the small and unimportant changes.

@AnthonyHorton can you look over and make sure it has everything you are expecting? There were some differences with the `library_path` being called in the init method that I was unsure of.

Edit: the `fli.py` file has changes for the constants, the other files are just pep8 changes.

PR #671, opened 2018-10-04
Fix breaking bug in Birger focuser

AnthonyHorton:
At some point in recent changes a `[0]` was accidentally removed from a line in `pocs/focuser/birger.py`. The effect of this is to cause the initialisation of the Birger focuser to fail. This was discovered today when attempting to run on real hardware.

I've also taken the liberty of removing a `\t\t\t` from a log message in `pocs/camera/sbig.py` which was a leftover from back when we were using lots of tabs to format our logging.

### DIFF
--- a/pocs/camera/sbig.py
+++ b/pocs/camera/sbig.py
@@ -39,7 +39,7 @@ class Camera(AbstractCamera):
                 self.ccd_cooling_enabled = True
             else:
                 self.ccd_cooling_enabled = False
-            self.logger.info('\t\t\t {} initialised'.format(self))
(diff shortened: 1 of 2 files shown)


AnthonyHorton:
It's somewhat documented. The `_send_command` docstring says that it returns a possibly empty list containing the lines of the response from the focuser.  There are a bunch of commands that return a 1 line response, and consequently a bunch of methods that do `response[0].rstrip()` to pull the string out of the list of lines. All the `[0].rstrip()`'s were moved in #620, but one of the `[0]`'s got dropped in the process.

PR #672, opened 2018-10-04
Decouple custom exceptions from PanBase

wtgee:
We have some dependencies in the custom exceptions file that make them hard to use outside of a full install of all POCS dependencies. In particular, I want to be able to load anything from `pocs.utils` without having to import `PanBase`. This is useful for, e.g. Cloud Functions.

It is unclear if we were using anything from `AstropyWarning` to our advantage but if so we are not testing for it.

Edit: Oh wow, I think this got rid of our surplus of astropy warnings. Go figure.

### DIFF
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -24,6 +24,11 @@ def test_error():
 
     assert str(e_info.value) == 'PanError'
 
+    with pytest.raises(SystemExit) as e_info:
+        raise error.PanError(msg="Testing exit", exit=True)
+    assert e_info.type == SystemExit
+    assert str(e_info.value) == 'PanError: Testing exit'
+
 
 def test_bad_load_module():
     with pytest.raises(error.NotFound):
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -14,12 +14,12 @@ from pocs.camera import list_connected_cameras
 
 
 def test_error(capsys):
-    with pytest.raises(Exception) as e_info:
+    with pytest.raises(error.PanError) as e_info:
         raise error.PanError(msg='Testing message')
 
     assert str(e_info.value) == 'PanError: Testing message'
 
-    with pytest.raises(Exception) as e_info:
+    with pytest.raises(error.PanError) as e_info:
         raise error.PanError()
 
     assert str(e_info.value) == 'PanError'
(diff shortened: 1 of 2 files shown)


pocs/tests/utils/test_utils.py
```diff
@@ -9,12 +9,34 @@
 from pocs.utils import listify
 from pocs.utils import load_module
 from pocs.utils import CountdownTimer
-from pocs.utils.error import NotFound
+from pocs.utils import error
 from pocs.camera import list_connected_cameras
 
 
+def test_error(capsys):
+    with pytest.raises(Exception) as e_info:
```
jamessynge: Should this be PanError instead of Exception? i.e. why test for Exception being raised?

pocs/tests/utils/test_utils.py
```diff
@@ -9,12 +9,34 @@
...
+        raise error.PanError(msg='Testing message')
+
+    assert str(e_info.value) == 'PanError: Testing message'
+
+    with pytest.raises(Exception) as e_info:
+        raise error.PanError()
+
+    assert str(e_info.value) == 'PanError'
+
+    with pytest.raises(SystemExit) as e_info:
```
jamessynge: As a reader, this is definitely odd. Why should it be this way? It seems that the only thing that PanError(exit=True) provides is a log.error followed by system.exit. It seems oddly hidden in raising an exception.

pocs/utils/error.py
```diff
@@ -1,28 +1,39 @@
...
-            else:
-                self.logger.error('{}: {}'.format(self.__class__.__name__, msg))
-                self.msg = msg
+            logger.error('{}: {}'.format(self.__class__.__name__, msg))
 
-    def exit_program(self, msg='No reason specified'):
+        if exit:
+            self.exit_program(msg)
+
+        self.msg = msg
```
jamessynge: How about setting the msg immediately on entry to `__init__`? Then you can make use of `__str__` to generate the log message.

pocs/utils/error.py
```diff
@@ -1,28 +1,39 @@
...
+
+    def exit_program(self, msg=None):
         """ Kills running program """
+        if not msg:
+            msg = 'No reason specified'
         print("TERMINATING: {}".format(msg))
         sys.exit(1)
 
+    def __str__(self):
+        error_str = 'PanError'
```
jamessynge: Does `self.__class__.__name__` equal the derived class while in `__init__` of PanError? If so, it would be nice to have the name of the derived class used instead of PanError.

pocs/tests/utils/test_utils.py
wtgee: Yes, I'm not sure this is how we should be counting on the system exit. This is a bit of a legacy and probably could be fixed up. I think it is only used in one location.

wtgee: Made #676

PR #673, opened 2018-10-04
Decouple serial utils

wtgee:
Follow up to #672, this removes the `PanBase` dependency from the serial utilities.

@jamessynge this works well with the tests. Any issues you can think of?

### DIFF
--- a/pocs/tests/test_rs232.py
+++ b/pocs/tests/test_rs232.py
@@ -58,7 +58,7 @@ def test_detect_bogus_scheme(handler):
 
 def test_custom_logger(handler, fake_logger):
     s0 = rs232.SerialData(port='no_op://', logger=fake_logger)
-    s0.debug('Testing logger')
+    s0.logger.debug('Testing logger')
 
 
 def test_basic_no_op(handler):
--- a/pocs/tests/test_rs232.py
+++ b/pocs/tests/test_rs232.py
@@ -16,6 +16,11 @@ def test_missing_port():
         rs232.SerialData()
 
 
+def test_custom_logger(fake_logger):
+    s0 = rs232.SerialData(logger=fake_logger)
+    s0.debug('Testing logger')
+
+
 def test_non_existent_device():
     """Doesn't complain if it can't find the device."""
     port = '/dev/tty12345698765'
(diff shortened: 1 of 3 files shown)


pocs/tests/test_rs232.py
```diff
@@ -56,6 +56,11 @@ def test_detect_bogus_scheme(handler):
     assert '# bogus #' in repr(excinfo.value)
 
 
+def test_custom_logger(handler, fake_logger):
+    s0 = rs232.SerialData(port='no_op://', logger=fake_logger)
+    s0.logger.debug('Testing logger')
```
jamessynge: What validation could we have here?

pocs/utils/rs232.py
```diff
@@ -30,7 +30,7 @@ def _parse_json(line, logger, min_error_pos=0):
 
 # Note: get_serial_port_info is replaced by tests to override the normal
 # behavior, so don't change the name without fixing the tests.
-def get_serial_port_info():
+def get_serial_port_info():  # pragma: no cover
```
jamessynge: We could exercise it in a test, even if it returns an empty list. What do you think?

pocs/utils/rs232.py
```diff
@@ -39,14 +39,45 @@ def get_serial_port_info():
...
+class SerialData(object):
     """SerialData wraps a PySerial instance for reading from and writing to a serial device.
 
     Because POCS is intended to be very long running, and hardware may be turned off when unused
     or to force a reset, this wrapper may or may not have an open connection to the underlying
     serial device. Note that for most devices, is_connected will return true if the device is
     turned off/unplugged after a connection is opened; the code will only discover there is a
     problem when we attempt to interact with the device.
+
+    .. doctest::
```
jamessynge: Yippee!

pocs/tests/test_rs232.py
wtgee: I was a bit stumped on that. We could write different messages to different log files I suppose. I'll see if I can add something.

pocs/utils/rs232.py
wtgee: These *should* also be counting toward our coverage but that's yet to be verified.

PR #674, opened 2018-10-04
TheSkyX Updates

wtgee:
* Remove `PanBase` dependency from TSX
* Reduce buffer size on socket
* Some additional scripts for dome and autoguider.

### DIFF
--- a/pocs/dome/__init__.py
+++ b/pocs/dome/__init__.py
@@ -1,6 +1,6 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
 
-import pocs
+from pocs.base import PanBase
 import pocs.utils
 import pocs.utils.logger as logger_module
 
@@ -33,7 +33,7 @@ def create_dome_from_config(config, logger=None):
     return dome
 
 
-class AbstractDome(pocs.base.PanBase):
+class AbstractDome(PanBase):
     """Abstract base class for controlling a non-rotating dome.
 
     This assumes that the observatory 'dome' is not a classic rotating
--- a/pocs/utils/theskyx.py
+++ b/pocs/utils/theskyx.py
@@ -1,10 +1,10 @@
 import socket
 
-from pocs.base import PanBase
 from pocs.utils import error
+from pocs.utils.logger import get_root_logger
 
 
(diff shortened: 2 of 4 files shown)


pocs/dome/__init__.py
```diff
@@ -1,6 +1,6 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
 
-import pocs
+from pocs.base import PanBase
```
jamessynge: Why change this import? I think it undoes a change I made earlier, IIRC.

wtgee: This was failing with just `import pocs`. We don't have anything in the `pocs/__init__.py` file so it wasn't pulling it in.

Could change to `import pocs.base` and then leave `pocs.base.PanBase` but I just imported this specifically.

wtgee: Looks like you changed it back in #418. I think things are much cleaner now.

pocs/utils/theskyx.py
```diff
@@ -1,18 +1,19 @@
...
+class TheSkyX(object):
 
     """ A socket connection for communicating with TheSkyX
 
 
     """
 
     def __init__(self, host='localhost', port=3040, connect=True, *args, **kwargs):
-        super(TheSkyX, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
```
AnthonyHorton: Just curious, why still have an explicit call to `super().__init__()` when subclassing from `object`? Does this actually do anything, or is it just there in case the base class is changed?

pocs/dome/__init__.py
wtgee: @jamessynge did you have concerns about this affecting the astrohaven domes?

pocs/utils/theskyx.py
wtgee: No, I was going to remove it but kept it for good measure. Will remove.

pocs/dome/__init__.py
jamessynge: Not a concern about the dome, just that we will benefit from a consistent style ... especially if it works well! ;-)

PR #678, opened 2018-10-05
Remove support for the 'fileline' attribute in log formats

jamessynge:
This non-standard attribute has been the cause of many problems,
so reverting back to using only the default attributes.

### DIFF
--- a/conf_files/log.yaml
+++ b/conf_files/log.yaml
@@ -7,8 +7,7 @@ logger:
         datefmt: '%H:%M:%S'
       detail:
         style: '{'
-        # See FilenameLineFilter in logger.py for fileline description
-        format: '{levelname:.1s}{asctime}.{msecs:03.0f} {fileline:20s} {message}'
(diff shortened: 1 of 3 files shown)


conf_files/log.yaml
```diff
@@ -7,8 +7,7 @@ logger:
         datefmt: '%H:%M:%S'
       detail:
         style: '{'
-        # See FilenameLineFilter in logger.py for fileline description
-        format: '{levelname:.1s}{asctime}.{msecs:03.0f} {fileline:20s} {message}'
+        format: '{levelname:.1s}{asctime}.{msecs:03.0f} {filename} {lineno} {message}'
```
wtgee: I'd prefer we still have a fixed length for `filename` and `lineno` so the the `message` always start with a left flush. I could live without but I think it makes reading easier.

jamessynge: Done.

PR #679, opened 2018-10-05
Dome Simulator in local config test failure.

wtgee:
If a dome simulator is explicitly added to the `pocs_local.yaml` file it breaks the testing, revealing deeper problems with the load config.

The issue is that the config is stored globablly, so with the move to hardware that is created before the Observatory we now can have the config loaded in a global way by hardware.

This is a temporary fix to an issue reported privately by @jamessynge. A deeper examination of the config loading w.r.t. local files is warranted.

### DIFF
--- a/pocs/tests/test_observatory.py
+++ b/pocs/tests/test_observatory.py
@@ -27,11 +27,13 @@ def simulator():
 @pytest.fixture
 def observatory(config, simulator):
     """Return a valid Observatory instance with a specific config."""
-    cameras = create_cameras_from_config(config)
     obs = Observatory(config=config,
                       simulator=simulator,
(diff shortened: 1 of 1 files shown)


wtgee:
> I'm a bit confused by this, i.e. why is it effective. But if you found that it is, let's go with it.

The cameras were being created first so they were calling `PanBase` before anything else, which was populating the global `_config`. But the camera creation is ignoring the `config` and the `ignore_local_config` so it loads the local dome simulator.

wtgee:
We will start to run into this problem more as we pull the mount, scheduler, dome, etc creation out of Observatory. That's why I say it's only a temp fix and not a real solution.

PR #680, opened 2018-10-06
Scrub paths from config load

wtgee:
Get rid of a few double-paths.

### DIFF
--- a/pocs/utils/config.py
+++ b/pocs/utils/config.py
@@ -130,10 +130,10 @@ def _parse_config(config):
     if 'directories' in config:
         base_dir = os.getenv('PANDIR')
         for dir_name, rel_dir in config['directories'].items():
-            if not rel_dir.startswith('/'):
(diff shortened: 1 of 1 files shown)


pocs/utils/config.py
```diff
@@ -131,7 +131,9 @@ def _parse_config(config):
         base_dir = os.getenv('PANDIR')
         for dir_name, rel_dir in config['directories'].items():
             if not rel_dir.startswith('/'):
-                config['directories'][dir_name] = '{}/{}'.format(base_dir, rel_dir)
+                config['directories'][dir_name] = os.path.normpath(os.path.join(
```
jamessynge: How about:
```
for dir_name, rel_dir in config['directories'].items():
    abs_dir = os.path.normpath(os.path.join(base_dir, rel_dir))
    if abs_dir != rel_dir:
        logger.debug('Modified "%s" directory to: %s', dir_name, abs_path)
        config['directories'][dir_name] = abs_dir
```

PR #681, opened 2018-10-06
Reduce pocs test redundancy (and fix camera simulator)

wtgee:
This changes and fixes a few things related to testing and the simulators:

* Make sure camera simulator is always using temp dir, not just during testing.
* Use a `simulator.yaml` targets file. We spend a *lot* of time loading the targets, especially in test_observatory.py. This reduces that a lot.
* Combined some tests in test_pocs that were causing the entire POCS object to be created for simple property tests. Should keep same coverage.
* Camera simulator exposure times reduced to 1s.

### DIFF
--- a/pocs/camera/simulator.py
+++ b/pocs/camera/simulator.py
@@ -38,6 +38,11 @@ class Camera(AbstractCamera):
             kwargs['exp_time'] = 1
             self.logger.debug("Trimming camera simulator exposure to 1 s")
 
+        # If POCSTIME is set for testing then the simulator will generate
+        # duplicate filenames each time so we pass a filename here. Full path
+        # is added in `camera._setup_obseravtion`.
+        filename = 'simulator_{:02d}.{}'.format(self.file_extension)
+
         return super().take_observation(observation,
                                         headers,
                                         filename,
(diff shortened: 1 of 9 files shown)


pocs/camera/simulator.py
```diff
@@ -34,9 +34,14 @@ def connect(self):
...
-        if exp_time > 2:
-            kwargs['exp_time'] = 2
-            self.logger.debug("Trimming camera simulator exposure to 2 s")
+        if exp_time > 1:
+            kwargs['exp_time'] = 1
+            self.logger.debug("Trimming camera simulator exposure to 1 s")
+
+        # If POCSTIME is set for testing then the simulator will generate
+        # duplicate filenames each time so we pass a filename here. Full path
+        # is added in `camera._setup_obseravtion`.
```
jamessynge: Typo in _setup_obseravtion

pocs/tests/conftest.py
```diff
@@ -15,14 +15,22 @@
...
 @pytest.fixture(scope='function')
-def config():
+def config(images_dir):
     pocs.base.reset_global_config()
 
     global _one_time_config
     if not _one_time_config:
         _one_time_config = load_config(ignore_local=True, simulator=['all'])
         _one_time_config['db']['name'] = 'panoptes_testing'
+        _one_time_config['directories']['images'] = images_dir
```
jamessynge: Will this line execute often enough? Or should it indented less?

wtgee: Would that be true of the db as well? Or because that is a global is it fine anyway?

Issue #653, opened 2018-10-01
Arduino Simulator Fixes


wtgee:
I am getting errors in the arduino simulator tests in unrelated PRs. In addition to that. the `pocs/serial_handlers/protocol_arduinosimulator.py` file reports coverage tests on almost every PR, pointing to some non-hermetic testing.

```
pocs/tests/test_arduino_io.py::test_basic_arduino_io FAILED                                                                                                                                                                                                                                                            [  3%]
                                                                                                     
========================================================================================================================================================== FAILURES ==========================================================================================================================================================
___________________________________________________________________________________________________________________________________________________ test_basic_arduino_io ____________________________________________________________________________________________________________________________________________________
                                                                                       
serial_handlers = True, memory_db = <pocs.utils.database.PanMemoryDB object at 0x7f09c5ceda20>, msg_publisher = <pocs.utils.messaging.PanMessaging object at 0x7f0999d6e128>, msg_subscriber = <pocs.utils.messaging.PanMessaging object at 0x7f0999d6e668>
cmd_publisher = <pocs.utils.messaging.PanMessaging object at 0x7f0999d6e048>, cmd_subscriber = <pocs.utils.messaging.PanMessaging object at 0x7f0999d6e6d8>
                                                                                                     
    def test_basic_arduino_io(serial_handlers, memory_db, msg_publisher, msg_subscriber, cmd_publisher,
                              cmd_subscriber):                                                       
        board = 'telemetry'                                                                        
        ser = arduino_io.open_serial_device('arduinosimulator://?board=' + board)                    
        board = board + '_board'                                                       
        aio = arduino_io.ArduinoIO(board, ser, memory_db, msg_publisher, cmd_subscriber)                                                                                                                                                                                                                                      
                                                                                       
        # Exercise ability to reconnect if disconnected.                                             
        aio.reconnect()                                                                
        aio.disconnect()                                                                             
                                                                                                                                                                                                                                                                                                                              
        got_reading = aio.read_and_record()                                                          
        if got_reading is False:                                                       
            got_reading = aio.read_and_record()                                                      
            assert got_reading is True                                                 
        else:                                                                                                                                                                                                                                                                                                                             assert got_reading is True                                                                                                                                                                                                                                                                                        
                                                                                                                                 
        # Check that the reading was sent as a message.                                   
        topic, msg_obj = msg_subscriber.receive_message(blocking=False)                                     
        assert topic == board                                                                                                                                                                                                                                                                                                 
[...]

PR #682, opened 2018-10-06
Increase coverage of arduino related code

jamessynge:
Add ability to shrink size of read chunks and size of read queue,
which will cause some chunks to be dropped, exercising the retry
logic.

Mark some code to be skipped in coverage calculations.

Partially addresses issue #653.

### DIFF
--- a/pocs/sensors/arduino_io.py
+++ b/pocs/sensors/arduino_io.py
@@ -75,7 +75,7 @@ def detect_board_on_port(port, logger=None):
                 return data['name']
             logger.warning('Unable to find board name in reading: {}', reading)
             return None
-        except Exception as e:
+        except Exception as e:  # pragma: no cover
             logger.error('Exception while auto-detecting port {}: {}', port, e)
     finally:
         if serial_reader:
--- a/pocs/serial_handlers/protocol_arduinosimulator.py
+++ b/pocs/serial_handlers/protocol_arduinosimulator.py
@@ -30,14 +30,17 @@ class ArduinoSimulator:
     at a rate similar to 9600 baud, the rate used by our Arduino sketches.
     """
 
-    def __init__(self, message, relay_queue, json_queue, stop, logger):
+    def __init__(self, message, relay_queue, json_queue, chunk_size, stop, logger):
         """
         Args:
             message: The message to be sent (millis and report_num will be added).
-            relay_queue: The queue.Queue instance from which replay command bytes are read
-                and acted upon. Elements are of type bytes.
-            json_queue: The queue.Queue instance to which json messages (serialized to bytes)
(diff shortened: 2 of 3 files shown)


pocs/sensors/arduino_io.py
```diff
@@ -75,7 +75,7 @@ def detect_board_on_port(port, logger=None):
                 return data['name']
             logger.warning('Unable to find board name in reading: {}', reading)
             return None
-        except Exception as e:
+        except Exception as e:  # pragma: no cover
```
wtgee: No way we can trigger it?

pocs/serial_handlers/protocol_arduinosimulator.py
```diff
@@ -30,14 +30,17 @@ class ArduinoSimulator:
...
-                and acted upon. Elements are of type bytes.
-            json_queue: The queue.Queue instance to which json messages (serialized to bytes)
-                are written at ~9600 baud. Elements are of type bytes.
+            relay_queue: The queue.Queue instance from which relay command
+                bytes are read and acted upon. Elements are of type bytes.
+            json_queue: The queue.Queue instance to which json messages
+                (serialized to bytes) are written at ~9600 baud. Elements
+                are of type bytes (i.e. each element is a sequence of bytes of
+                length up to chunk_size).
+            chunk_size: The number of bytes to write to json_queue at a time.
```
wtgee: No default for this?

pocs/serial_handlers/protocol_arduinosimulator.py
```diff
@@ -173,17 +178,15 @@ def __init__(self, *args, **kwargs):
...
+        self.relay_queue = queue.Queue(maxsize=1)
+        self.json_queue = queue.Queue(maxsize=1)
         self.json_bytes = bytearray()
         self.stop = threading.Event()
         self.stop.set()
         self.device_simulator = None
 
     def __del__(self):
-        if self.simulator_thread:
+        if self.simulator_thread:  # pragma: no cover
```
wtgee: This seems like a lot to not cover

pocs/sensors/arduino_io.py
jamessynge: No, not that I was able to find.

pocs/serial_handlers/protocol_arduinosimulator.py
jamessynge: Yes, but not here, instead down in _create_simulator.

pocs/serial_handlers/protocol_arduinosimulator.py
jamessynge: It basically is covering a misuse of the simulator, i.e. of test code.

PR #684, opened 2018-10-07
Updates to current_time

wtgee:
* **Behavior change** Increment by one second after reading value. Helps with test problems due to colliding filenames coming from the camera simulator. Also slightly more "realistic"
* Docs updates

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -148,6 +148,10 @@ class PocsShell(Cmd):
         """Setup and initialize a POCS instance."""
         simulator = listify(arg[0].split())
 
+        if 'POCSTIME' in os.environ:
+            print_warning("Clearing POCSTIME variable")
+            del os.environ['POCSTIME']
+
         if simulator is None:
             simulator = []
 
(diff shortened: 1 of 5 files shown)


bin/pocs_shell
```diff
@@ -148,6 +148,10 @@ class PocsShell(Cmd):
         """Setup and initialize a POCS instance."""
         simulator = listify(arg[0].split())
 
+        if 'POCSTIME' in os.environ:
```
jamessynge: I'm curious what motivated this addition? Not asking for a change.

pocs/tests/test_observatory.py
```diff
@@ -184,14 +184,14 @@ def test_standard_headers(observatory):
         'latitude': 19.54,
         'longitude': -155.58,
         'moon_fraction': 0.7880103086091879,
-        'moon_separation': 156.1607340087774,
+        'moon_separation': 148.34401,
```
jamessynge: These changes seem unrelated to the description.

pocs/utils/__init__.py
```diff
@@ -49,6 +63,8 @@ def current_time(flatten=False, datetime=False, pretty=False):
 
     if pocs_time is not None and pocs_time > '':
         _time = Time(os.getenv('POCSTIME'))
```
jamessynge: Why repeat the call to getenv, rather than use the `pocs_time` variable?

PR #685, opened 2018-10-08
Fix bug in ArduinoIO.handle_commands

jamessynge:
handle_commands was not retrying the read from the cmd_subscriber.
Added a sleep() method to CountdownTimer to help with waiting for
some portion of the time left in a countdown timer.

Add more testing of ArduinoIO.

### DIFF
--- a/pocs/utils/__init__.py
+++ b/pocs/utils/__init__.py
@@ -109,7 +109,7 @@ class CountdownTimer(object):
         assert duration >= 0, "Duration must be non-negative."
         self.is_non_blocking = (duration == 0)
 
-        self.duration = duration
+        self.duration = float(duration)
         self.restart()
 
     def expired(self):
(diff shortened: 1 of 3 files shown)


pocs/utils/__init__.py
```diff
@@ -135,12 +135,30 @@ def time_left(self):
...
                 return self.duration
             else:
-                return max(0, delta)
+                return max(0.0, delta)
 
     def restart(self):
         """Restart the timed duration."""
         self.target_time = time.monotonic() + self.duration
 
+    def sleep(self, max_sleep=None):
```
wtgee: For the usage that you have here, with very small times, I don't think this is a problem but it concerns me that this could be blocking on long countdown, such as used in exposures. The answer of course is to not use this method then but I can see it being temping. 

Maybe a note in the docstring about the (obvious) fact that it is blocking. A callback would be nice that's probably overkill.

jamessynge: Created issue #686 to track this.

PR #687, opened 2018-10-11
Observation cleanup cleanup

wtgee:
* Restores manifest log for uploads
* Puts cleaning/uploading of observation into separate subprocess outside of Observatory (currently this is still blocking per observation but could be changed - also still called from Observatory).
* Adding config key for service account

Remaining items from #577 for #599. Also restores some items from #654.


pocs/utils/google/storage.py
```diff
@@ -315,8 +315,15 @@ def upload_observation_to_bucket(pan_id,
     Returns:
         str: A string path used to search for files.
     """
-    assert os.path.exists(dir_name)
-    assert re.match(r'PAN\d\d\d', pan_id) is not None
+    if os.path.exists(dir_name) is False:
+        raise OSError("Directory does not exist, cannot upload: {}".format(dir_name))
+
+    if re.match(r'PAN\d\d\d', pan_id) is None:
```
jamessynge: This pattern will also match PAN0000000000. If you want to limit to 3 digits, you could use `r'PAN\d\d\d$'`.

pocs/utils/google/storage.py
```diff
@@ -315,8 +315,15 @@ def upload_observation_to_bucket(pan_id,
...
+        raise Exception("Invalid PANID. Must be of the form 'PANXXX'. Got: {!r}".format(pan_id))
```
jamessynge: How about: `Must be of the form 'PANnnn'.`

scripts/upload_files.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash -e
+
+usage() {
+  echo -n "##################################################
```
jamessynge: I'm guessing that the ########### lines at the top and bottom are different lengths.

scripts/upload_files.sh
```diff
@@ -0,0 +1,55 @@
...
+# Upload files to Google Cloud Storage bucket.
+# 
+# This script will call the underlying gsutil utility, using
+# the parallel options (-m) by default. This will also write
+# a manifest log into the directory which can be used for upload
+# valildation.
```
jamessynge: typo: valildation -> validation

scripts/upload_files.sh
```diff
@@ -0,0 +1,55 @@
...
+#
+# Note:
+# The SEARCH_PATH and DESTINATION are often similar. Normal invocation
+# is done via $POCS/scripts/upload_image_dir.py. See also the utils in
+# $POCS/pocs/utils/google/storage.py.
+#
+# Note:
+# The SEARCH_PATH can be any pattern, e.g:
+#     - 'PAN006'  # All images for unit
+#     - 'Hd189733'  # All images for field
```
jamessynge: Should that be `*/Hd189733`?

upload_files.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash -e
```
jamessynge: Is this file a mistake? There is another one in scripts/

PR #689, opened 2018-10-13
Add timeout support to PanMessaging.receive_message

jamessynge:
Add testing of conftest.py messaging fixtures to
test_messaging.py.

Delay message_forwarder fixture to allow forwarder to start.
Will later work on reducing that delay or eliminating the
need for a delay.

Remove unused imports.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,6 +9,7 @@
 import os
 import pytest
 import subprocess
+import time
 
 from pocs import hardware
 from pocs.utils.database import PanDB
@@ -17,17 +18,18 @@ from pocs.utils.messaging import PanMessaging
 
 # Global variable set to a bool by can_connect_to_mongo().
 _can_connect_to_mongo = None
+_all_databases = ['mongo', 'file', 'memory']
 
 
 def pytest_addoption(parser):
     hw_names = ",".join(hardware.get_all_names()) + ' (or all for all hardware)'
+    db_names = ",".join(_all_databases) + ' (or all for all databases)'
     group = parser.getgroup("PANOPTES pytest options")
     group.addoption(
         "--with-hardware",
         nargs='+',
         default=[],
-        help=("A comma separated list of hardware to test. " + "List items can include: " +
-              hw_names))
+        help=("A comma separated list of hardware to test. List items can include: " + hw_names))
     group.addoption(
         "--without-hardware",
         nargs='+',
@@ -50,7 +52,8 @@ def pytest_addoption(parser):
         "--test-databases",
         nargs="+",
         default=['mongo'],
-        help="Test databases in the list. Note that travis-ci will test all of them by default.")
+        help=("Test databases in the list. List items can include: " + db_names +
+              ". Note that travis-ci will test all of them by default."))
 
 
 def pytest_collection_modifyitems(config, items):
@@ -217,7 +220,7 @@ def can_connect_to_mongo():
     return _can_connect_to_mongo
 
 
-@pytest.fixture(scope='function', params=['mongo', 'file', 'memory'])
+@pytest.fixture(scope='function', params=_all_databases)
 def db_type(request):
 
     db_list = pytest.config.option.test_databases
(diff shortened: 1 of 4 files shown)


conftest.py
```diff
@@ -50,7 +52,8 @@ def pytest_addoption(parser):
         "--test-databases",
         nargs="+",
         default=['mongo'],
-        help="Test databases in the list. Note that travis-ci will test all of them by default.")
+        help=("Test databases in the list. List items can include: " + db_names +
```
jamessynge: Sorry, I apparently couldn't resist adding an improvement to the help text while I was working in this file.

conftest.py
```diff
@@ -267,6 +270,9 @@ def message_forwarder(messaging_ports):
         args.append(str(pub))
 
     proc = subprocess.Popen(args, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    # It takes a while for the forwarder to start, so allow for that.
+    # TODO(jamessynge): Come up with a way to speed up these fixtures.
+    time.sleep(3)
```
wtgee: Were we missing messages in some of the tests?

pocs/utils/messaging.py
```diff
@@ -201,7 +206,7 @@ def send_message(self, topic, message):
         # Send the message
         self.socket.send_string(full_message, flags=zmq.NOBLOCK)
 
-    def receive_message(self, blocking=True, flags=0):
+    def receive_message(self, blocking=True, flags=0, timeout_ms=0):
```
wtgee: Can you add the param to the docstring.

pocs/utils/messaging.py
```diff
@@ -217,9 +222,15 @@ def receive_message(self, blocking=True, flags=0):
...
+        elif timeout_ms > 0:
+            # Wait until a message is available or the timeout expires.
+            self.socket.poll(timeout=timeout_ms, flags=(zmq.POLLIN | zmq.POLLOUT))
+            # Don't block at this point, because we will have waited as long
+            # as necessary.
+            flags = flags | zmq.NOBLOCK
         try:
             message = self.socket.recv_string(flags=flags)
-        except Exception:
+        except Exception as e:
```
wtgee: I'm surprised pyflakes doesn't complain about not using the `e`. Any reason it was added?

pocs/tests/test_messaging.py
```diff
@@ -85,3 +85,60 @@ def test_storage_id(pub_and_sub, config, db):
...
+    topic, msg_obj = sub.receive_message()
+    assert isinstance(msg_obj, dict)
+    assert 'message' in msg_obj
+    assert msg_obj['message'] == 'a string'
+    assert 'timestamp' in msg_obj
+
+    # import pdb;pdb.set_trace()
+
+
+def test_msg_pub_sub(msg_publisher, msg_subscriber):
```
wtgee: Seems like these two tests could be one but with a param of params. That said, this probably is more obvious.

conftest.py
jamessynge: Only for tests that I'd written, and was trying to add in another PR that I'll send soon for coverage of the arduino i/o code. I'm going to try to improve the msg forwarding support in the tests in general a bit later.

pocs/utils/messaging.py
jamessynge: Good question. Probably because I was using the debugger. I'll double check pyflakes on this.

jamessynge:
Adding this in support of testing. In particular, want to be able to call receive_message but not block indefinitely if something is wrong.

pocs/utils/messaging.py
jamessynge: Done.

pocs/tests/test_messaging.py
jamessynge: Ack.

jamessynge:
The `create_forwarder_sockets` code currently runs in another process, and isn't captured by coverage AFAICT.

Issue #420, opened 2018-01-27
Disentangle reading from multiple arduinos


jamessynge:
ArduinoSerialMonitor supports reading from both the telemetry and camera boards, but effectively requires them to produce output at exactly the same rate else a data loss will occur.

I'm planning to create pocs/sensors/arduino_io.py, based on peas/sensors.py, but the primary class will support a single serial connected device.

jamessynge:
Related to #326, we should consider how to find the available relays on each board. One option would be for the json output to tell us. For example:
`... "relays": {"camera": 1, "fan": 0, ...}, ...`
Then we can send to the board a name that it produced:
`camera=0\n`

jamessynge:
I've created ArduinoIO and SensorRecorder (PR #428), designed for working in a multithreaded Python app. However, I've just learned that debugging such an app isn't possible using Python's native debugger, pdb. While I could debug it using log messages, my inclination is to instead head in a direction that Wilfred mentioned: use separate processes for each Arduino. Specifically:

- Each process works with a specific device (e.g. `/dev/ttyACM1`).
- PanDB is used to save each reading into a history collection and a current collection, named with the name of the board.
- PanMessaging is used for two purposes:
  - Sending readings to the rest of the system on channel environment.
  - Receiving relay change commands (format to be determined, though probably using a channel name that includes the name of the board).
- The process stops if the device stops working; this allows a supervising process to re-do auto-detection periodically and find the board attached to a new device. While it would be useful for a message to be sent indicating that there is an issue, zeromq doesn't guarantee delivery, so we might want to rely on other means for detecting that the connection has been lost.

wtgee:
Think this sounds like a good move and would also be easy to implement for the Pi.

jamessynge:
PR coming in a few minutes. The ArduinoIO object listens for commands to channel `<board>:commands` (e.g. `telemetry_board:commands`), with the following incoming messages:

`{"command": "shutdown"}` - Stops the main loop.

`{"command": "write_line", "line": "some line of text"}` - Sends the line of text to the Arduino (serial port)

I plan to extend this (once we agree on the protocol) to support changing relays by name (already supported by the telemetry board sketch). That makes it easier to replace the Arduino with a Raspberry Pi or other maker board. That might mean that we should change the name from ArduinoIO to SerialSensorIO, etc.

wtgee:
Linking #125

jamessynge:
Resuming work on this, with this plan:

[...]

PR #690, opened 2018-10-13
Add DelaySigTerm to pocs/utils.

jamessynge:
Used to prevent SIGTERM from interrupting a "critical section",
which in this case is any block of code run as follows:

```
with DelaySigTerm():
    code-to-protect-from-sigterm
```

This is a part of addressing issue #420.

### DIFF
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -130,8 +130,9 @@ def test_delay_of_sigterm_with_nosignal():
 def test_delay_of_sigterm_with_handled_signal():
     """Confirm that another type of signal can be handled.
 
-    In this test we'll send SIGCHLD, which should call the
-    handler the test installs immediately.
+    In this test we'll send SIGCHLD, which should immediately call the
+    signal_handler the test installs, demonstrating that only SIGTERM
+    is affected by this DelaySigTerm.
     """
     test_signal = signal.SIGCHLD
 
@@ -208,6 +209,9 @@ def test_delay_of_sigterm_with_raised_exception():
             # Send the test signal. It should immediately
             # call our handler.
             os.kill(os.getpid(), test_signal)
+            # Should not reach this point because signal_handler() should
+            # be called because we called:
+            #     signal.signal(other-handler, signal_handler)
             after_signal = True
             assert False, "Should not get here!"
     except UserWarning:
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -1,11 +1,12 @@
 import os
 import pytest
-
+import signal
 import time
 from datetime import datetime as dt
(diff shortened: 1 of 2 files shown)


jamessynge:
This will be used in some of the ArduinoIO code to handle shutting down safely when SIGTERM is received.

pocs/tests/utils/test_utils.py
```diff
@@ -115,3 +116,149 @@ def test_countdown_timer():
...
+
+        after_with = True
+        assert signal.getsignal(signal.SIGTERM) == orig_sigterm_handler
+    finally:
+        assert before_signal
+        assert signal_handled
+        assert after_signal
+        assert after_with
+        assert signal.getsignal(signal.SIGTERM) == orig_sigterm_handler
+        signal.signal(test_signal, old_test_signal_handler)
```
wtgee: Just curious, do you need to explicitly restore this even though the function will go out of scope?

pocs/tests/utils/test_utils.py
```diff
@@ -115,3 +116,149 @@ def test_countdown_timer():
...
+        # Install our handler.
+        signal.signal(test_signal, signal_handler)
+
+        with DelaySigTerm():
+            assert signal.getsignal(signal.SIGTERM) != orig_sigterm_handler
+            before_signal = True
+            # Send the test signal. It should immediately
+            # call our handler.
+            os.kill(os.getpid(), test_signal)
+            after_signal = True
```
wtgee: How about:

```python
after_signal = True  # Should not get here
assert after_signal
```
just to make it clear that it shouldn't get to either of these statements.

pocs/tests/utils/test_utils.py
```diff
@@ -115,3 +116,149 @@ def test_countdown_timer():
...
+    with DelaySigTerm():
+        assert signal.getsignal(signal.SIGTERM) != orig_sigterm_handler
+
+    assert signal.getsignal(signal.SIGTERM) == orig_sigterm_handler
+
+
+def test_delay_of_sigterm_with_handled_signal():
+    """Confirm that another type of signal can be handled.
+
+    In this test we'll send SIGCHLD, which should call the
```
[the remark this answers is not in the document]
jamessynge: Done.

pocs/tests/utils/test_utils.py
jamessynge: So far as I know you do.

pocs/tests/utils/test_utils.py
jamessynge: Added a bunch of comments explaining, rather than the proposed code.

jamessynge:
This is for the process that interacts with the Arduino, so production code rather than testing code. And yes, it could presumably be used in additional places.

PR #691, opened 2018-10-14
Make all created Threads daemons

jamessynge:
Some non-main threads weren't daemons, and were preventing pytest
from exiting when a test failed because the thread still existed.

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -1,11 +1,10 @@
+import copy
+import os
 import re
 import shutil
 import subprocess
+import threading
 import yaml
-import os
-import copy
-from threading import Event
-from threading import Thread
 
 from astropy.io import fits
 from astropy.time import Time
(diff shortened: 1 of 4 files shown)


pocs/camera/camera.py
```diff
@@ -214,8 +213,10 @@ def take_observation(self, observation, headers=None, filename=None, *args, **kw
                 observation.exposure_list[image_id] = file_path
 
         # Process the exposure once readout is complete
-        t = Thread(target=self.process_exposure, args=(
-            metadata, observation_event, exposure_event))
+        t = threading.Thread(
+            target=self.process_exposure,
+            args=(metadata, observation_event, exposure_event),
+            daemon=True)
```
wtgee: I think for the camera and some of these others I was following the advice given in the note [here](https://docs.python.org/3/library/threading.html#thread-objects). I can't think of any unintended consequences of daemonizing this with the camera but seems like it would be hard to test until we have one fail.

jamessynge:
Yes, it avoids some of those.

PR #693, opened 2018-10-15
Replace command collect_status with status in pocs_shell

jamessynge:
New command prints status once, then returns control to the user.

Fix initialization of cameras to honor simulator specified to
setup_pocs command.

Fix a couple of typos.

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -217,6 +217,7 @@ Hardware names: {}   (or all for all hardware)'''.format(
         if self.msg_subscriber is None:
             self.do_start_messaging()
         status = self.pocs.status()
+        print()
         pprint(status)
         print()
 
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -156,12 +156,12 @@ class PocsShell(Cmd):
             simulator = []
 
         try:
-            cameras = create_cameras_from_config()
+            cameras = create_cameras_from_config(simulator=simulator)
             observatory = Observatory(simulator=simulator, cameras=cameras)
             self.pocs = POCS(observatory, messaging=True)
             self.pocs.initialize()
-        except error.PanError:
-            pass
+        except error.PanError as e:
+            print_warning('Problem setting up POCS: {}'.format(e))
 
     def help_setup_pocs(self):
         print('''Setup and initialize a POCS instance.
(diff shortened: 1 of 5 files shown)


pocs/camera/__init__.py
```diff
@@ -64,6 +65,8 @@ def create_cameras_from_config(config=None, logger=None, **kwargs):
     if not logger:
         logger = logger_module.get_root_logger()
 
+    logger.info(f'kwargs: {kwargs!r}')
```
wtgee: Maybe just a `debug` level on this.

pocs/camera/__init__.py
```diff
@@ -79,7 +82,9 @@ def kwargs_or_config(item, default=None):
 
     logger.debug("Camera config: {}".format(camera_info))
 
-    a_simulator = 'camera' in kwargs_or_config('simulator', default=list())
+    simulator_names = hardware.get_simulator_names(config=config, kwargs=kwargs)
+    logger.info(f'simulator_names = {", ".join(simulator_names)}')
```
wtgee: `info` seems a little much here too when coming from the camera since it is showing all values (not just camera ones).  Maybe an info if `a_simulator` is True? (although I thought we had that)

bin/pocs_shell
```diff
@@ -209,32 +209,16 @@ Hardware names: {}   (or all for all hardware)'''.format(
...
-                    status = self.pocs.status()
-                    pprint(status)
-                    print_info('*' * 80)
-                    self.pocs.db.insert_current('system', status, store_permanently=False)
-                    time.sleep(2)
-                except KeyboardInterrupt:
-                    print_warning('Stopping collection')
-                    _keep_going = False
-        else:
+    def do_status(self, *arg):
```
wtgee: I think the default `do_emptyline` should call this (which is what happens in the mostly unused `DriftShell`)

pocs/camera/__init__.py
jamessynge: Dropped that, I should have removed it before creating the PR.

pocs/camera/__init__.py
jamessynge: Switched to debug.

bin/pocs_shell
jamessynge: Having the handler for empty line produce a lot of output (or really, to have it do much of anything) seems odd to me.

jamessynge:
I notice that there is a line in the Arduino simulator that is sometimes getting hit, sometimes not, so still causing variation in coverage. Will look into that later.

PR #694, opened 2018-10-15
Quick Camera fix

wtgee:
* Bug fix from #693
* Skip gphoto2 camera if it doesn't respond to request for serialnumber

### DIFF
--- a/pocs/camera/__init__.py
+++ b/pocs/camera/__init__.py
@@ -159,12 +159,9 @@ def create_cameras_from_config(config=None, logger=None, **kwargs):
                                 filter_type=camera_filter,
                                 focuser=camera_focuser,
                                 readout_time=camera_readout)
-        except ImportError:
-            # Raise errors if bad module
-            raise error.CameraNotFound(msg=camera_model)
-        except error.CameraNotFound as e:
+        except Exception as e:
             # Warn if bad camera but keep trying other cameras
-            logger.warning(e)
+            logger.error(msg="Cannot find camera type: {} {}".format(camera_model, e))
         else:
             is_primary = ''
             if camera_info.get('primary', '') == cam.uid:
--- a/pocs/camera/__init__.py
+++ b/pocs/camera/__init__.py
@@ -81,7 +81,7 @@ def create_cameras_from_config(config=None, logger=None, **kwargs):
     logger.debug("Camera config: {}".format(camera_info))
 
     simulator_names = hardware.get_simulator_names(config=config, kwargs=kwargs)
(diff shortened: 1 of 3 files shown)


pocs/camera/__init__.py
```diff
@@ -81,7 +81,7 @@ def kwargs_or_config(item, default=None):
     logger.debug("Camera config: {}".format(camera_info))
 
     simulator_names = hardware.get_simulator_names(config=config, kwargs=kwargs)
-    logger.debug(f'simulator_names = {", ".join(simulator_names)}')
+    logger.debug('simulator_names = {}'.format(", ".join(simulator_names)))
```
jamessynge: Did you intend to revert this?

pocs/camera/__init__.py
```diff
@@ -151,9 +151,6 @@ def kwargs_or_config(item, default=None):
         try:
             module = load_module('pocs.camera.{}'.format(camera_model))
             logger.debug('Camera module: {}'.format(module))
-        except ImportError:
```
jamessynge: This seems like a very different type of problem from "the camera is connected but unresponsive" or "the camera is disconnected". ImportError seems to say that the config is bad. Why is that not worth raising for?

pocs/camera/__init__.py
wtgee: Yes, it doesn't work with the `join` in the f-string and caused a bug. Really it should be it's own PR.

pocs/camera/__init__.py
wtgee: I originally had it with both types of exceptions raised but we want the same behavior in either case,  namely to give a warning but still try the other cameras. The `logger.error` will display separate messages depending on why it failed. Maybe just a more generic message instead of "Cannot find camera type"?

pocs/camera/__init__.py
wtgee: Hm, odd. I'll try to recreate the problem on PAN001.

I think it was coming through as some mixed type, e.g.:
```
In [1]: lst = ['a', None]

In [2]: f'lst = {", ".join(lst)}'
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-296f01098e95> in <module>()
----> 1 f'lst = {", ".join(lst)}'

TypeError: sequence item 1: expected str instance, NoneType found
```

But that fails outside the f-string anyway.

wtgee: Oh, hm. PAN001 is on python 3.5.

wtgee: I'll remove from this PR.

PR #699, opened 2018-10-18
Use DelaySigTerm to cleanly stop ArduinoIO

jamessynge:
Expose an ArduinoIO.stop_running property, use
from arduino_recorder.py CLI script to stop when
SIGTERM is received.

Added a callback arg to DelaySigTerm, allowing
arduino_recorder.py to be notified when
 SIGTERM is received.

FYI, the reason that I didn't use DelaySigTerm
directly in ArduinoIO.handle_reading is that it
doesn't work in threads other than main, and that
makes testing and uses of ArduinoIO other than
in arduino_recorder.py more difficult.
Nonetheless, it makes DelaySigTerm less applicable
than originally intended.

Fix minor comment error in logger.py.

Continues work on issue #420.

### DIFF
--- a/pocs/sensors/arduino_io.py
+++ b/pocs/sensors/arduino_io.py
@@ -9,6 +9,7 @@ import copy
 import serial
 import time
 from serial import serialutil
+import threading
 import traceback
 
 from pocs.utils.error import ArduinoDataError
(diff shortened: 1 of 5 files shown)


jamessynge:
I agree, the names suck here. I too found it confusing. I'm going to go ahead and merge, but will add fixing this to the list of things to do.

PR #702, opened 2018-10-18
Implement PanMongoDb.permanently_erase_database for tests

jamessynge:
Update PanMongoDb.clear_current to use non-deprecated method.

### DIFF
--- a/pocs/utils/database.py
+++ b/pocs/utils/database.py
@@ -161,20 +161,7 @@ class PanDB(object):
         if not isinstance(db_type, str) and db_type:
             raise ValueError('db_type, a string, must be provided and not empty')
 
-        # Pre-defined list of collections that are valid.
-        collection_names = [
-            'camera_board',
-            'config',
-            'current',
-            'drift_align',
-            'environment',
-            'mount',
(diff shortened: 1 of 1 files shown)


jamessynge:
If you search for permanently_erase_database, you'll see that it is only used in the test fixtures, and has those extra arguments `really` and `dangerous` that are designed to prevent this from being used by mistake. For example:
```
@pytest.fixture(scope='function', params=_all_databases)
def db_type(request):

    db_list = pytest.config.option.test_databases
    if request.param not in db_list and 'all' not in db_list:
        pytest.skip("Skipping {} DB, set --test-all-databases=True".format(request.param))

    # If testing mongo, make sure we can connect, otherwise skip.
    if request.param == 'mongo' and not can_connect_to_mongo():
        pytest.skip("Can't connect to {} DB, skipping".format(request.param))
    PanDB.permanently_erase_database(
        request.param, 'panoptes_testing', really='Yes', dangerous='Totally')
    return request.param
```

jamessynge:
Added issue #705 to track improving testing of the classes in this file.

Issue #700, opened 2018-10-18
Don't start message forwarder in POCS


jamessynge:
I extracted the message forwarder into a separate process months ago, but haven't yet removed the attempt to start another forwarder inside of the `POCS` class.

- [x] Modify messaging fixtures to be compatible with POCS.
- [x] Modify config fixture to take messaging ports from messaging_ports fixture.
- [ ] Modify POCS to not start forwarders in _setup_messaging.
- [ ] Modify test_pocs.py to be more robust... there are tests that fail intermittently, apparently related to waiting for messages.

wtgee:
Closing as stale. Some of this has been done, other parts have changed. Reopen with more information if necessary.

PR #706, opened 2018-10-19
Use non-production ports in the config test fixture

jamessynge:
Copy the ports from the messaging_ports fixture into the config fixture.

Part of #700.

### DIFF
--- a/pocs/tests/conftest.py
+++ b/pocs/tests/conftest.py
@@ -10,6 +10,7 @@ import pytest
 
 import pocs.base
 from pocs.utils.config import load_config
+from pocs.utils.logger import get_root_logger
 
 # Global variable with the default config; we read it once, copy it each time it is needed.
 _one_time_config = None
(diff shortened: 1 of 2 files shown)


pocs/tests/conftest.py
```diff
@@ -22,19 +23,38 @@ def images_dir(tmpdir_factory):
...
         _one_time_config['scheduler']['fields_file'] = 'simulator.yaml'
 
-    _one_time_config['directories']['images'] = images_dir
+    # Make a copy before we modify based on test fixtures.
+    result = copy.deepcopy(_one_time_config)
 
-    return copy.deepcopy(_one_time_config)
+    # We allow for each test to have its own images directory, and thus
+    # to not conflict with each other.
+    result['directories']['images'] = images_dir
```
wtgee: Ah, good, thanks. (even though not part of PR)

pocs/tests/conftest.py
```diff
@@ -22,19 +23,38 @@ def images_dir(tmpdir_factory):
...
+
+    # For now (October 2018), POCS assumes that the pub and sub ports are
+    # sequential. Make sure that is what the test fixtures have in them.
+    # TODO(jamessynge): Remove this once pocs.yaml (or messaging.yaml?) explicitly
+    # lists the ports to be used.
+    assert messaging_ports['cmd_ports'][0] == (messaging_ports['cmd_ports'][1] - 1)
```
wtgee: Does this make more sense in the `messaging_ports` fixture?

jamessynge: Yes, except that the point here is to ensure that the messaging_ports has provided the correct pairs. The fix to do so was in another PR, hence that is why I put it here, and also because I wanted the enforcement closer to where the problem arises... still not here, but by the time we're in POCS we don't see that there are pairs of ports.

pocs/tests/conftest.py
jamessynge: OK, fine, I did do a little straightening up and commenting while I was at it! ;-)

pocs/tests/conftest.py
jamessynge: FYI, I plan to get the the point where pocs.yaml actually has the pair of ports listed, rather than implicit, which will eliminate the need for these asserts.

jamessynge:
Hey, we're reached 80% coverage again!!

PR #708, opened 2018-10-21
Improve robustness of log formatter

jamessynge:
Attempt to undo the special casing that the standard
logging package applies to a dict as the second of
two args to a logging method (i.e. format_str, dict_value).

Annoyingly complicated. Makes me wonder if we should
simply require use of legacy formatting. Oh well, wasted
the effort to fix it, so let's try it out.

Add tests of log formatting code.

### DIFF
--- a/pocs/utils/logger.py
+++ b/pocs/utils/logger.py
@@ -1,11 +1,15 @@
-import json
-import os
-import sys
-import time
+import collections
 import datetime
+import json
 import logging
 import logging.config
+import os
+import re
+import string
+import sys
 from tempfile import gettempdir
+import time
+from warnings import warn
 
 from pocs.utils.config import load_config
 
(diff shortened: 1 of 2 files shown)


pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+def field_name_to_key(field_name):
+    """Given a field_name from Formatter.parse(), extract the argument key."""
+    assert isinstance(field_name, str)
+    assert len(field_name)
+    m = re.match(r'^([^.[]+)', field_name)
+    if not m:
+        return None
+    arg_name = m.group(1)
+    if arg_name.isdigit():
+        pos = int(arg_name)
```
wtgee: This looks unused.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
 all_loggers = {}
 
 
+def field_name_to_key(field_name):
+    """Given a field_name from Formatter.parse(), extract the argument key."""
```
wtgee: Docstring needs param defined.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+    arg_name = m.group(1)
+    if arg_name.isdigit():
+        pos = int(arg_name)
+        return int(arg_name)
+    else:
+        return arg_name
+
+
+def format_references_keys(fmt, args):
+    assert isinstance(args, dict)
```
wtgee: Docstring.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+                key = field_name_to_key(field_name)
+                if isinstance(key, str) and key in args:
+                    return True
+    except Exception:
+        pass
+    return False
+
+
+def format_has_legacy_style(fmt):
+    fmt = fmt.replace('%%', '')
```
wtgee: Simple but still needs a docstring.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+    return '%' in fmt
+
+
+formatting_methods = dict(
```
wtgee: Some comments describing this would be nice.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+        return None
+    arg_name = m.group(1)
+    if arg_name.isdigit():
+        pos = int(arg_name)
+        return int(arg_name)
+    else:
+        return arg_name
+
+
+def format_references_keys(fmt, args):
```
wtgee: Maybe `format_has_reference_keys` to be more consistent with method below and to clarify usage?

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+        # Looks old school.
+        method_names.append('legacy_direct')
+
+    # First try with the selected method names.
+
+    for method_name in method_names:
+        try:
+            method = formatting_methods[method_name]
+            msg = method(fmt, args)
+            if debug:
```
wtgee: Do you want to leave these debug statements here? It's a little clunky but I understand could be useful.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+                print(f'Method {method_name!r} worked for {fmt!r}')
+            return msg
+        except Exception:
+            if debug:
+                print(f'Method {method_name!r} Failed for {fmt!r}')
+
+    # Didn't succeed. Select fallback methods, again based on content.
+    fallback_names = []
```
wtgee: If `method_names` is a list can't we just put the fallback methods on the end of the list and only run through the loop sequence once?

pocs/utils/logger.py
```diff
@@ -72,9 +199,9 @@ def get_root_logger(profile='panoptes', log_config=None):
         # missing from the log_config. It is hard to understand how
         # this could occur given that none of the callers of
         # get_root_logger pass in their own log_config.
-        if 'formatters' not in log_config and sys.stdout.isatty():
-            import pdb
-            pdb.set_trace()
+        if 'formatters' not in log_config:  # pragma: no cover
+            warn('formatters is missing from log_config!')
```
wtgee: So since  #672 we can use things from `pocs.utils.error` within the other util classes. This might be a good place to raise a custom exception rather than the `warn`.

pocs/utils/logger.py
```diff
@@ -15,30 +19,153 @@
...
+        except Exception:
+            if debug:
+                print(f'Method {method_name!r} Failed for {fmt!r}')
+
+    warn(f'Unable to format log.')
+    warn(f'Log message (format string): {fmt!r}')
+    warn('Log args type: %s' % type(args))
+    try:
+        warn(f'Log args: {args!r}')
+    except Exception:
```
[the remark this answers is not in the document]
jamessynge: Agreed, but for now just marked as no cover.

pocs/utils/logger.py
jamessynge: Added a TODO.

PR #709, opened 2018-10-21
Improve error handling in pocs/utils/images

jamessynge:
Produce more explicit errors about why images can't be
linked in attempt to address an issue seen on travis.

Improve test coverage of images code.

Move fixtures for test images to POCS/conftest.py.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -324,26 +324,26 @@ def save_environ():
     os.environ = old_env
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
 def data_dir():
     return os.path.join(os.getenv('POCS'), 'pocs', 'tests', 'data')
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
 def unsolved_fits_file(data_dir):
     return os.path.join(data_dir, 'unsolved.fits')
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
 def solved_fits_file(data_dir):
     return os.path.join(data_dir, 'solved.fits.fz')
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
 def tiny_fits_file(data_dir):
     return os.path.join(data_dir, 'tiny.fits')
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
 def noheader_fits_file(data_dir):
     return os.path.join(data_dir, 'noheader.fits')
--- a/conftest.py
+++ b/conftest.py
@@ -6,6 +6,7 @@
 # In addition, there are fixtures defined here that are available to
 # all tests, not just those in pocs/tests.
 
+import copy
 import os
 import pytest
 import subprocess
@@ -314,3 +315,35 @@ def cmd_subscriber(message_forwarder):
     subscriber = PanMessaging.create_subscriber(port)
     yield subscriber
     subscriber.close()
+
+
+@pytest.fixture(scope='function')
+def save_environ():
+    old_env = copy.deepcopy(os.environ)
+    yield
+    os.environ = old_env
+
+
+@pytest.fixture
+def data_dir():
+    return os.path.join(os.getenv('POCS'), 'pocs', 'tests', 'data')
+
+
+@pytest.fixture
+def unsolved_fits_file(data_dir):
+    return os.path.join(data_dir, 'unsolved.fits')
+
+
+@pytest.fixture
+def solved_fits_file(data_dir):
+    return os.path.join(data_dir, 'solved.fits.fz')
+
+
+@pytest.fixture
+def tiny_fits_file(data_dir):
+    return os.path.join(data_dir, 'tiny.fits')
+
+
+@pytest.fixture
+def noheader_fits_file(data_dir):
+    return os.path.join(data_dir, 'noheader.fits')
--- a/pocs/tests/test_images.py
+++ b/pocs/tests/test_images.py
@@ -1,5 +1,7 @@
 import os
 import pytest
+import shutil
+import tempfile
 
 from pocs.images import Image
 from pocs.images import OffsetError
@@ -8,6 +10,15 @@ from pocs.utils.error import Timeout
 
 from astropy import units as u
 from astropy.coordinates import SkyCoord
+
+
+def copy_file_to_dir(to_dir, file):
+    assert os.path.isfile(file)
+    result = os.path.join(to_dir, os.path.basename(file))
+    assert not os.path.exists(result)
+    shutil.copy(file, to_dir)
+    assert os.path.exists(result)
+    return result
 
 
 def test_fits_exists(unsolved_fits_file):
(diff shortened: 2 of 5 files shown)


conftest.py
```diff
@@ -314,3 +315,35 @@ def cmd_subscriber(message_forwarder):
...
+
+
+@pytest.fixture(scope='function')
+def save_environ():
+    old_env = copy.deepcopy(os.environ)
+    yield
+    os.environ = old_env
+
+
+@pytest.fixture
```
wtgee: We could probably `session` scope all of these path name fixtures. Not sure if that actually gains us anything.

pocs/tests/test_images.py
```diff
@@ -132,14 +112,16 @@ def test_pointing_error(solved_fits_file):
     assert (perr.magnitude.to(u.degree).value - 1.9445870862060288) < 1e-5
 
 
-# def test_compute_offset_arcsec(solved_fits_file, unsolved_fits_file):
-#     img0 = Image(solved_fits_file)
-#     img1 = Image(unsolved_fits_file)
+def test_compute_offset_arcsec(solved_fits_file, unsolved_fits_file):
+    img0 = Image(solved_fits_file)
+    img1 = Image(solved_fits_file)
```
wtgee: We should probably have two FITS files in a sequence that we can use that actually generates an offset. I thought that is what the `unsolved_fits_file` was and at one point I think creating an `Image` would solve by default, but I'm wondering if this was commented out because that doesn't happen any more (by default). We could use `unsolved_fits_file` as `img1` and do an `img1.solve_field()` and then have an offset, but that might take more time then we want. Maybe could just check the offset in a test that is already solving?

pocs/tests/test_images.py
```diff
@@ -132,14 +112,16 @@ def test_pointing_error(solved_fits_file):
...
 
-#     offset_info = img0.compute_offset(img1)
+    offset_info = img0.compute_offset(img1)
+    print('offset_info:', offset_info)
```
wtgee: This doesn't need to be printed.

pocs/tests/utils/test_image_utils.py
```diff
@@ -21,6 +39,49 @@ def test_crop_data():
     assert cropped03.sum() == 36.
 
 
+def test_make_pretty_image(solved_fits_file, tiny_fits_file, save_environ):
+    # First make a dir and put the files in it
+    with tempfile.TemporaryDirectory() as tmpdir:
+        fz_file = os.path.join(tmpdir, os.path.basename(solved_fits_file))
+        fits_file = os.path.join(tmpdir, os.path.basename(tiny_fits_file))
+        # TODO Add a small CR2 file to our sample image files.
```
wtgee: I recall that it was actually really hard to find a decently small CR2 file, which is why one was never placed there. There isn't really a tool to crop CR2 that still produces as CR2 as a result. Could maybe dig around a bit more to find one.

pocs/tests/utils/test_image_utils.py
```diff
@@ -21,6 +39,49 @@ def test_crop_data():
...
+
+        # Can't operate on non-existent files.
+        with pytest.warns(UserWarning):
+            assert not img_utils.make_pretty_image(fz_file)
+        with pytest.warns(UserWarning):
```
wtgee: This is sort of the same test as above.

pocs/utils/images/__init__.py
```diff
@@ -83,27 +94,28 @@ def make_pretty_image(fname, title=None, timeout=15, link_latest=False, **kwargs
...
+        return pretty_path
+
     # Symlink latest.jpg to the image; first remove the symlink if it already exists.
-    if link_latest and os.path.exists(pretty_path):
-        latest_path = os.path.join(
-            os.getenv('PANDIR'),
-            'images',
-            'latest.jpg'
-        )
+    images_dir = make_images_dir()
```
wtgee: Having this called from `make_pretty_image` seems a bit odd given it's name. Might make more sense to have a `get_images_dir` that just silently handles making the directory.

pocs/utils/images/__init__.py
```diff
@@ -25,6 +25,17 @@
 palette.set_bad('g', 1.0)
 
 
+def make_images_dir():
+    images_dir = os.path.join(os.getenv('PANDIR'), 'images')
+    try:
+        os.makedirs(images_dir, exist_ok=True)
+        return images_dir
+    except Exception as e:
+        warn(f'Unable to create the images directory: {images_dir}')
```
wtgee: Could raise a custom exception here.

pocs/utils/images/__init__.py
```diff
@@ -25,6 +25,17 @@
 palette.set_bad('g', 1.0)
 
 
+def make_images_dir():
```
wtgee: Docstring.

pocs/tests/utils/test_image_utils.py
```diff
@@ -7,6 +9,22 @@
...
+def test_make_images_dir(save_environ):
+    assert img_utils.make_images_dir()
+
+    # Invalid parent directory for 'images'.
+    os.environ['PANDIR'] = '/dev/null/'
+    with pytest.warns(UserWarning):
+        assert img_utils.make_images_dir() is None
+
+    # Valid parents for 'images' that need to be created.
+    with tempfile.TemporaryDirectory() as tmpdir:
```
wtgee: We have a fixture for a [temp images dir](https://github.com/panoptes/POCS/blob/42e717a2099e5714757e0d5e503919d865535d7f/pocs/tests/conftest.py#L20), can that be used? It already appends an `images` to the end but could still test the `make_images_dir` functionality.

pocs/tests/utils/test_image_utils.py
```diff
@@ -21,6 +39,49 @@ def test_crop_data():
...
+        with pytest.warns(UserWarning):
+            assert not img_utils.make_pretty_image(fz_file)
+        with pytest.warns(UserWarning):
+            assert not img_utils.make_pretty_image(fits_file)
+
+        # Copy the files.
+        shutil.copy(solved_fits_file, tmpdir)
+        shutil.copy(tiny_fits_file, tmpdir)
+
+        # Not a valid file type (can't automatically handle fits.fz files).
```
wtgee: This shouldn't be true and if it is we need to fix it, not test around it.

pocs/tests/utils/test_image_utils.py
```diff
@@ -21,6 +39,49 @@ def test_crop_data():
     assert cropped03.sum() == 36.
 
 
+def test_make_pretty_image(solved_fits_file, tiny_fits_file, save_environ):
+    # First make a dir and put the files in it
+    with tempfile.TemporaryDirectory() as tmpdir:
```
wtgee: Same comment about `images_dir` fixture.

pocs/tests/utils/test_image_utils.py
```diff
@@ -21,6 +39,49 @@ def test_crop_data():
...
+        with pytest.warns(UserWarning):
+            assert not img_utils.make_pretty_image(fz_file)
+
+        # Can handle the fits file, and creating the images dir for linking
+        # the latest image.
+        imgdir = os.path.join(tmpdir, 'images')
+        assert not os.path.isdir(imgdir)
+        os.environ['PANDIR'] = tmpdir
+
+        pretty = img_utils.make_pretty_image(fits_file, link_latest=True)
```
wtgee: I want to say that the issue arises when there is a stale symlink (perhaps with different permissions?) so I'm worried this won't fully cover the problem but it's definitely an improvement.

pocs/tests/utils/test_image_utils.py
jamessynge: Agreed, I wasn't able to find a cropped CR2 file. There are some that are nearer to 5MB than 20MB available, but not small test images. What I found surprising was that github had very few such files, and the programs that I found that can read CR2 files (e.g. libraw) apparently don't have publicly visible tests as they don't have publicly visible sample data to test against.

Issue #697, opened 2018-10-17
Add AC sensor signal into safety decision of POCS


jamessynge:
@wtgee Per our discussion with Nem today.

jamessynge:
Specifically, POCS.is_safe should add a new entry to the is_safe_values dictionary for AC. The environment data for the telemetry board already has the info we're looking for; it _simply_ needs to be extracted and compared against a suitable threshold. Unfortunately, that threshold is likely to be unit specific.

FWIW, I'd really like to change the design so that we get a binary signal indicating if AC is present or not, without a comparison required. Using the Mean Well brand power supply would provide that. Similarly, just plugging in a 5V wall wart (e.g. the type used to power an Arduino UNO) could be used to close or open a relay, and we could use a digital IO pin on the Arduino to detect that the AC to 5 VC DC conversion is happening. It isn't quite the same as having the power supply tell us that it has 5V, but it is easy to create out of standard parts.

wtgee:
> FWIW, I'd really like to change the design so that we get a binary signal indicating if AC is present or not, without a comparison required

I'd like this. It would be great if we could just read a boolean flag out of the arduino.

I'll get to this work this week. Thanks for writing up Issue.

PR #710, opened 2018-10-22
AC Power Check

wtgee:
Start to implement the safety check for the AC power as per #697.

The AC power reading will work similar to the weather in that will both check for a simulator (`power`) as well as an entry in the database.

Note that this creates a new top-level collection called `power`. For now this will just have a `main` (`mains`?) entry but I imagine will have the other sensed lines in the future. This code will expect a `power.main` entry in the `current` collection.

Also includes some slight clean up for some of the other safety checks.

### DIFF
--- a/pocs/utils/database.py
+++ b/pocs/utils/database.py
@@ -498,10 +498,10 @@ class PanMemoryDB(AbstractPanDB):
             obj = json_util.loads(obj)
         return obj
 
-    def clear_current(self, type):
+    def clear_current(self, entry_type):
         try:
-            del self.current['type']
-        except KeyError:
+            del self.current[entry_type]
+        except KeyError as e:
             pass
 
     @classmethod


pocs/core.py
```diff
@@ -403,6 +406,60 @@ def has_free_space(self, required_space=0.25 * u.gigabyte):
...
+            if 'power' in self.config['simulator']:
+                self.logger.debug("AC power simulator always safe")
+                return True
+
+        # Get current power readings from database
+        try:
+            record = self.db.get_current('power')
+            has_power = bool(record['data'].get('main', False))
+
+            timestamp = record['date'].replace(tzinfo=None)  # current_time is timezone naive
```
jamessynge: I'm noticing a pattern here. Perhaps we need to fix current_time or should add support for an 'age' field into pocs/utils/database.py?

pocs/core.py
```diff
@@ -403,6 +406,60 @@ def has_free_space(self, required_space=0.25 * u.gigabyte):
...
+            age = (current_time().datetime - timestamp).total_seconds()
+
+            self.logger.debug("Power Safety: {} [{:.0f} sec old - {:%Y-%m-%d %H:%M:%S}]",
```
jamessynge: That format is worth of an f-string! ;-)

pocs/hardware.py
```diff
@@ -1,6 +1,14 @@
 """Information about hardware supported by Panoptes."""
 
-ALL_NAMES = sorted(['camera', 'dome', 'mount', 'night', 'weather', 'theskyx'])
+ALL_NAMES = sorted([
+    'camera',
+    'dome',
+    'mount',
+    'night',
```
jamessynge: Along with my suggestion that we add a `power_failure` simulator, we could also add `day` and `storm` simulators.

We probably need different variables here to hold different kinds of names: equipment names vs. condition names. `day` and `night`, or `has_ac` and `ac_failure` would represent conditions, while `camera`, `dome`, etc. represent equipment.

pocs/tests/test_pocs.py
```diff
@@ -324,6 +324,38 @@ def test_unsafe_park(pocs):
     pocs.power_down()
 
 
+def test_no_ac_power(pocs):
+    # Simulator makes AC power safe
+    assert pocs.has_ac_power() is True
+
+    # Remove 'power' from simulator
+    pocs.config['simulator'] = ['camera', 'mount', 'weather']
```
jamessynge: Why not `pocs.config['simulator'].remove('power')`?

pocs/tests/test_pocs.py
```diff
@@ -324,6 +324,38 @@ def test_unsafe_park(pocs):
...
+    pocs.initialize()
+
+    # With simulator removed the power should fail
```
jamessynge: Could also insert into the test db a current record for power showing a failure (or success). That way more of the code is evaluated.

pocs/core.py
```diff
@@ -298,12 +298,20 @@ def is_safe(self, no_warning=False):
 
         is_safe_values = dict()
 
+        # Check if AC power connected and return immediately if not
```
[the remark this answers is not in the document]
wtgee: The idea was that if power is out you don't want to waste time/power checking the weather or anything so that it is conserving battery.

pocs/core.py
```diff
@@ -328,15 +336,14 @@ def is_dark(self):
             bool: Is night at location
 
         """
-        # See if dark
+        # See if dark - we check this first because we want to know
+        # the sun position even if using a simulator.
```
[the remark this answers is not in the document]
wtgee: Correct, we don't report the sun angle anywhere except for when doing the dark check. I've thought about changing that in the past. Here i just took the opportunity to update the comment.

pocs/core.py
wtgee: The timezone info we probably should figure out a bit better w.r.t. `current_time`. As I recall it was not as trivial as i wanted it to be.

pocs/core.py
wtgee: Actually this is a place where it doesn't help because then the string is longer than the line-length.

pocs/core.py
```diff
@@ -403,6 +406,60 @@ def has_free_space(self, required_space=0.25 * u.gigabyte):
...
+                defaults to 90 seconds.
+
+        Returns:
+            bool: True if system AC power is present.
+        """
+        # Always assume False
+        self.logger.debug("Checking for AC power")
+        has_power = False
+
+        # TODO(wtgee): figure out if we really want to simulate no power
```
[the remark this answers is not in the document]
wtgee: See below.

pocs/hardware.py
wtgee: I think we should discuss this more and figure out what we want. I agree there should be some logical distinction for the simulator types. For this PR I will leave the simple `power` simulator.

pocs/tests/test_pocs.py
wtgee: Yes, that's much better. Thanks!

pocs/tests/test_pocs.py
wtgee: I enter a range of values in the loop that follows this. Were you thinking of something different?

pocs/core.py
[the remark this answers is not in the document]
jamessynge: Well, that is true, but the same might be said of weather too. Unless we have evidence otherwise, I think these will all happen fairly fast so we don't need to worry about the short circuit.

OTOH, I'd like to head to a situation in which there is no db look up or call at all. We accumulate reports from various sensors and can periodically turn that into a health report, including dealing with stale reports. But, that is for later.

pocs/core.py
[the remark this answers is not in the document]
jamessynge: Thanks, appreciate that clarity.

pocs/core.py
jamessynge: I think you're saying that record['date'] is of type date.date (or an astropy equivalent), and has tzinfo in it, but that current_time() returns local time with no tzinfo in it. Correct?

pocs/core.py
jamessynge: Sigh

pocs/hardware.py
jamessynge: Agreed on both fronts.

pocs/core.py
wtgee: Yes, correct. Created #716.

PR #712, opened 2018-10-24
Test caching of astrometry and using libcfitsio packages

jamessynge:

PR #713, opened 2018-10-27
Extract astrometry installation into a script.

jamessynge:
This is a variant of my earlier #712. In addition to those changes,
I'm attempting to deal with astrometry.net being down by using github
as a backup location for getting astrometry. Not sure if it will work.


jamessynge:
Closing in favor of other changes.

PR #718, opened 2018-11-01
Docker for build dependencies only - WIP

jamessynge:
Incomplete, but almost there.


jamessynge:
Please take a look and let me know what you think... or try it out yourself.
I plan to remove the $POCS/* contents from the image, to add useful readme content, and maybe to add a script to automate the build process (i.e. inserting the appropriate strings into the Dockerfile and piping that the `docker build`).

scripts/install/create-core-panoptes-directories.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -e
+#
+# Setup directories needed by PANOPTES. Other scripts, such as
+# install-dependencies.sh, expect to find these directories.
+# We do NOT parameterize these paths (i.e. don't use environment
```
wtgee: Why not just allow this from `$PANDIR`?

ci/full-dependencies/Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+ENV SOLVE_FIELD=/usr/bin/solve-field
+
+################################################################################
+# Several update steps before our custom commands. By rarely changing these
+# commands, we allow the caching that 'docker build' performs to save us some
+# time when iterating on the Panoptes specific steps (i.e. instead of executing
+# these, the cached image layer is used).
+
+# Update the information we know about package versions.
+RUN apt-get update --fix-missing
```
wtgee: My understanding is that in general you want to minimize the number of `RUN`, `COPY`, and `ADD` steps as they each create new layers and thus inflate the size. I realize we might be caching them but it still causes the overall container size to be larger.

Unless this is being used as part of a multi-stage build? But if that is the case it seems like we should stage it up even more.

ci/full-dependencies/Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+#
+#         git ls-tree HEAD | cut -d$'\t' -f2 | xargs -i echo "COPY {} {}"
+# 
+# This command produces a single line version, which I've hand wrapped below.
+#
+#         echo COPY $(git ls-tree HEAD | cut -d$'\t' -f2) '$POCS/'
+
+# First we copy the files and directories least likely to change while
+# iterating on the building of the Docker image.
+COPY .codecov.yml .coveragerc .gitignore .gitmodules .pycodestyle.cfg \
```
wtgee: I don't understand why the selective copying? Is this also for caching purposes?

scripts/install/create-core-panoptes-directories.sh
```diff
@@ -0,0 +1,44 @@
...
+  (set -x ; ${MKDIR_ROOT} -p /var/panoptes)
+else
+  echo "/var/panoptes already exists."
+fi
+
+echo "Ensuring that all files in /var/panoptes are owned by ${PANUSER}."
+(set -x ; ${CHOWN} -R "${PANUSER}:${PANUSER}" /var/panoptes)
+(set -x ; ${CHMOD} -R 755 /var/panoptes)
+
+for SUBDIR in POCS PAWS logs astrometry/data
```
wtgee: These directories all seem a) hard-coded b) buried in the script.

scripts/install/install-astrometry.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -ex
```
wtgee: How does this file related to #712 and #713?

scripts/install/install-astrometry.sh
```diff
@@ -0,0 +1,47 @@
...
+
+ASTROMETRY_VERSION="${ASTROMETRY_VERSION:-0.72}"
+ASTROMETRY_DIR="astrometry.net-${ASTROMETRY_VERSION}"
+
+TAR_NAME="astrometry.net-${ASTROMETRY_VERSION}.tar.gz"
+TAR_URL="http://astrometry.net/downloads/${TAR_NAME}"
```
wtgee:
```suggestion
TAR_URL="https://github.com/dstndstn/astrometry.net/releases/download/${ASTROMETRY_VERSION}/${TAR_NAME}"
```

scripts/install/install-astrometry.sh
```diff
@@ -0,0 +1,47 @@
...
+  exit
+fi
+
+cd $PANDIR
+echo "Downloading ${TAR_URL} ..."
+$DO_WGET "${TAR_URL}" || /bin/true
+if [ -f "${TAR_NAME}" ] ; then
+  echo "Unpacking ${TAR_NAME} ..."
+  tar zxf "${TAR_NAME}"
+else
```
wtgee: If the zip is trying to work around the missing `astrometry.net` domain then we should just default to using the tarball from github. I don't see any reason to prefer the `astrometry.net` link.

scripts/install/install-astrometry.sh
```diff
@@ -0,0 +1,47 @@
...
+  echo "Unpacking ${ZIP_NAME} ..."
+  unzip "${ZIP_NAME}"
+fi
+
+echo "Building astrometry in ${ASTROMETRY_DIR}..."
+cd "${ASTROMETRY_DIR}"
+
+make
+make py
+find . -type f -executable -ls
```
wtgee: Debug?

scripts/install/install-dependencies.sh
```diff
@@ -669,6 +679,11 @@ function test_installed_astrometry_version() {
 # then builds and installs into ${ASTROMETRY_DIR}. Skips as much as
 # possible if able to determine that the version hasn't changed.
 function install_astrometry() {
+  ASTROMETRY_VERSION="${ASTROMETRY_VERSION}" "${THIS_DIR}/install-astrometry.sh"
+  return
+
+
+
```
wtgee: Do you still need everything below if the script is already installing?

wtgee: > IFF we are going to move in the direction of docker

I still don't understand if this question regarding docker is for testing (on travis) or if it applies more generally to an install method. Or even more sorta-generally as an image that could be used, e.g. for compute instance images, travis, etc.

Breaking it up would favor either approach as you could have different files (or I think there are build "flavors" or something) for different targets but build them from the same intermediary stages.

scripts/install/install-astrometry.sh
wtgee: Yeah, NP.

scripts/install/create-core-panoptes-directories.sh
wtgee: It just seems to hardcode the directory so I was wondering the reason.

jamessynge:
I'm going to close this, and will send other smaller PRs when ready.

PR #720, opened 2018-11-08
Fix install-dependencies.sh bug

jamessynge:
Was writing to PANOPTES_ENV_SH when it should have written to SHELL_RC.

Use /etc/os-release if available instead of lsb_release (which is deprecated).

### DIFF
--- a/scripts/install/install-dependencies.sh
+++ b/scripts/install/install-dependencies.sh
@@ -15,7 +15,7 @@ if [[ -z "${PANDIR}" ]] ; then
     PANDIR=/var/panoptes
     GUESSED_A_VAR=1
   else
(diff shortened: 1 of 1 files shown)


scripts/install/install-dependencies.sh
```diff
@@ -394,7 +396,12 @@ function install_mongodb() {
   local MONGO_VERSION=""
   local MONGO_SOURCE_PATH=""
   local LSB_RELEASE=""
-  if [[ -n "$(safe_which lsb_release)" ]] ; then
+  # lsb_release is deprecated, removed from many debian distributions,
```
wtgee: I didn't realize this, thanks.

jamessynge:
FYI, I added some other minor fixes to the same file, not having realized that you'd approved.

PR #721, opened 2018-11-09
Add scripts for working with a cache for apt-get

jamessynge:
run-apt-cacher-ng-in-docker.sh starts a docker container
with a caching proxy for apt-get downloads.

configure-apt-cache.sh records under /etc/apt/apt.conf.d/
the info about the location (URL) of the caching proxy.

These support building and rebuilding of docker images
without repeatedly downloading the same linux packages
over and over again.

### DIFF
--- a/scripts/install/configure-apt-cache.sh
+++ b/scripts/install/configure-apt-cache.sh
@@ -1,0 +1,60 @@
+#!/bin/bash -e
+
+# Tell apt where to find the cache for packages, if one has been setup.
+# We look for the APT_PROXY_PORT environment variable to be set to the port
+# number, else for the first argument to be set to the port number.
+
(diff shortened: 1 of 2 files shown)


jamessynge:
While I'm focusing for the short run on CI applications of docker, I'm *very* open to applying it to managing the software configuration on units, too. We clearly have an issue w.r.t. complexity when setting up a unit, and that is going to be a barrier. OTOH, docker brings its own new realm of complexity, so we have some trade-offs to evaluate.

So, I'm writing PRs like this with some degree of generality in mind, and as a result am placing scripts that apply to installation in scripts/install, whether they are currently for docker image building or for installing on the host OS.

Seem OK?

wtgee:
> Seem OK?

Yup, that's fine, thanks for update. Just trying to get a sense for when doing the reviews.

PR #692, opened 2018-10-14
Improve coverage of ArduinoIO and simulator

jamessynge:
ArduinoSimulator now adds commands it receives
to the next output message that it produces.
Send command bytes to the simulator one at a time,
again to better simulate the reality that the
Arduino communication is serial... and to improve
code coverage.
Add "# pragma: no cover" to various unreachable
lines in the Arduino simulator code.

Test that ArduinoIO listens for messages on the
command topic and forwards the contained line
to the Arduino.

Use the new timeout_ms param of PanMessaging.receive_message
in ArduinoIO.handle_commands.

Add some comments, TODOs, and clean up some
code style issues.

PR #725, opened 2018-11-10
Increase coverage of Observatory dome functions

jamessynge:
There has been some variability of coverage in Observatory.open_dome,
so this change deliberately repeats completed operations to try to get
better coverage.


jamessynge:
The changes in #692 should have had zero effect on the dome test as the dome and arduino simulators don't share any code. Yes, they're both pyserial pseudo devices, don't share code IIRC.

PR #726, opened 2018-11-11
Add install-apt-packages.sh

jamessynge:
scripts/install/install-apt-packages.sh will work in docker build,
docker run or on the debian/ubuntu host. Will use an APT caching
proxy if one is provided via command line or environment variable.
Will use sudo if not running as root, else just execute the apt
commands directly.

Add scripts/install/install-helper-functions.sh with functions
from install-dependencies.sh and new functions to support using
an APT caching proxy.

Use install-helper-functions.sh from configure-apt-cache.sh
and install-dependencies.sh.

Use install-apt-packages.sh from install-dependencies.sh.

Removed some dead code from install-dependencies.

Tidied up install shell scripts following much of the advice
from shellcheck, but also disabled some shellcheck complaints.

### DIFF
--- a/scripts/install/configure-apt-cache.sh
+++ b/scripts/install/configure-apt-cache.sh
@@ -4,7 +4,7 @@
 # if one has been setup. Pass in the port as the first (and only) argument
 # to this script, else set the APT_PROXY_PORT environment variable to the
 # port number. Records the URL of the proxy "permanently" in
-# /etc/apt/apt.conf.d/01proxy; root root access is required to write to
+# /etc/apt/apt.conf.d/01proxy; root access is required to write to
 # that file.
 #
 # Using this script is appropriate if it isn't practical to pass the
--- a/scripts/install/configure-apt-cache.sh
+++ b/scripts/install/configure-apt-cache.sh
@@ -1,52 +1,45 @@
 #!/bin/bash -e
 
-# Tell apt where to find the cache for packages, if one has been setup.
-# We look for the APT_PROXY_PORT environment variable to be set to the port
-# number, else for the first argument to be set to the port number.
-
-# Requires root access in order to write to /etc/apt/apt.conf.d/01proxy
-
-# TODO(james): Determine if we can replace this change of the config file
-# with a --option flag on apt-get, ala:
+# Tell apt where to find the local (on-host) caching proxy for APT packages,
+# if one has been setup. Pass in the port as the first (and only) argument
+# to this script, else set the APT_PROXY_PORT environment variable to the
+# port number. Records the URL of the proxy "permanently" in
+# /etc/apt/apt.conf.d/01proxy; root root access is required to write to
+# that file.
+#
+# Using this script is appropriate if it isn't practical to pass the
+# URL of the proxy on the apt-get command line, such as this way:
 #
 #   apt-get -o "Acquire::HTTP::Proxy=${APT_PROXY_URL}" install ...
 #
-# This would avoid the need for sudo outside of docker build, and would
-# avoid docker build permanently recording the proxy in docker image.
-
+# The script install-apt-packages.sh uses this latter approach.
+#
 # To learn more, see:
 # https://github.com/sameersbn/docker-apt-cacher-ng
 # https://gist.github.com/dergachev/8441335
 # https://chandrusoft.wordpress.com/2012/09/30/disable-proxy-when-using-apt-get/
 
-APT_PROXY_PORT="${APT_PROXY_PORT:-${1}}"
-if [ -z "${APT_PROXY_PORT}" ] ; then
-  echo "No apt cache proxy declared via APT_PROXY_PORT or script argument."
-  exit 0
-fi
-echo "APT_PROXY_PORT: ${APT_PROXY_PORT}"
+THIS_DIR="$(dirname "$(readlink -f "${0}")")"
+# shellcheck source=/var/panoptes/POCS/scripts/install/install-helper-functions.sh
+source "${THIS_DIR}/install-helper-functions.sh"
 
-# Need to figure out our IP address. We'll try multiple techniques.
-HOST_IP=$(awk '/^[a-z]+[0-9]+\t00000000/ { printf("%d.%d.%d.%d\n", "0x" substr($3, 7, 2), "0x" substr($3, 5, 2), "0x" substr($3, 3, 2), "0x" substr($3, 1, 2)) }' < /proc/net/route)
-if [ -z "${HOST_IP}" -a -x /sbin/ip ] ; then
-  HOST_IP=$(/sbin/ip route get 8.8.8.8 | awk -F"src " 'NR==1{split($2,a," ");print a[1]}')
-fi
-if [ -z "${HOST_IP}" ] ; then
-  echo "Unable to determine host IP address!"
+# shellcheck disable=SC2086
+APT_PROXY_URL="$(get_apt_proxy_url ${1})"
+if [ -z "${APT_PROXY_URL}" ] ; then
+  echo "Unable to find operating APT proxy!" 1>&2
   exit 1
 fi
 
-APT_PROXY_URL="http://${HOST_IP}:${APT_PROXY_PORT}"
 APT_PROXY_FILE=/etc/apt/apt.conf.d/01proxy
 
-if [ 0 == $(id -u) ]; then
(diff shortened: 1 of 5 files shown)


scripts/install/configure-apt-cache.sh
```diff
@@ -1,52 +1,45 @@
...
-
-# Requires root access in order to write to /etc/apt/apt.conf.d/01proxy
-
-# TODO(james): Determine if we can replace this change of the config file
-# with a --option flag on apt-get, ala:
+# Tell apt where to find the local (on-host) caching proxy for APT packages,
+# if one has been setup. Pass in the port as the first (and only) argument
+# to this script, else set the APT_PROXY_PORT environment variable to the
+# port number. Records the URL of the proxy "permanently" in
+# /etc/apt/apt.conf.d/01proxy; root root access is required to write to
```
wtgee: double word `root root`

scripts/install/install-apt-packages.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Installs APT packages (e.g. debian/ubuntu packages installed for all users).
+# Requires being logged in as root or the ability to successfully execute sudo.
+
+THIS_DIR="$(dirname "$(readlink -f "${0}")")"
+# shellcheck source=/var/panoptes/POCS/scripts/install/install-helper-functions.sh
+source "${THIS_DIR}/install-helper-functions.sh"
+
+# Suppress prompting for input during package processing, unless DEBIAN_FRONTEND
+# is already set. The tzdata
```
wtgee: Trailing sentence.

scripts/install/install-apt-packages.sh
```diff
@@ -0,0 +1,50 @@
...
+if [ -z "${DEBIAN_FRONTEND}" ] ; then
+  export DEBIAN_FRONTEND=noninteractive
+fi
+
+# Generate the basic apt-get install command, minus the list of packages.
+declare -a apt_get_install=(apt-get install --no-install-recommends --yes)
+# shellcheck disable=SC2119
+apt_proxy_url="$(get_apt_proxy_url)"
+if [ -n "${apt_proxy_url}" ] ; then
+  apt_get_install+=(-o "Acquire::HTTP::Proxy=${apt_proxy_url}")
```
wtgee: Should we use the HTTPS by default?

scripts/install/install-apt-packages.sh
```diff
@@ -0,0 +1,50 @@
...
+  for APT_PKGS_FILE in "$@"
+  do
+    # Remove all the comments from the package list and install the packages whose
+    # names are left.
+    APT_PKGS="$(cut '-d#' -f1 "${APT_PKGS_FILE}" | sort | uniq)"
+    echo
+    echo_running_sudo "apt-get install for the files in ${APT_PKGS_FILE}"
+    echo
+    # shellcheck disable=SC2086
+    my_sudo "${apt_get_install[@]}" ${APT_PKGS}
```
wtgee: The `apt_get_install[@]` is an odd syntax to me, can you comment on what that is doing?

scripts/install/configure-apt-cache.sh
```diff
@@ -1,52 +1,45 @@
...
 # https://gist.github.com/dergachev/8441335
 # https://chandrusoft.wordpress.com/2012/09/30/disable-proxy-when-using-apt-get/
 
-APT_PROXY_PORT="${APT_PROXY_PORT:-${1}}"
-if [ -z "${APT_PROXY_PORT}" ] ; then
-  echo "No apt cache proxy declared via APT_PROXY_PORT or script argument."
-  exit 0
-fi
-echo "APT_PROXY_PORT: ${APT_PROXY_PORT}"
+THIS_DIR="$(dirname "$(readlink -f "${0}")")"
```
wtgee: These nested double quotes don't seem right but I'm assuming this works since it is in all the files here.

scripts/install/install-dependencies.sh
```diff
@@ -563,6 +481,7 @@ function install_conda() {
   # Where CONDA_INSTALL_DIR is where Anaconda or miniconda was installed.
   # We do the first step here. Later we'll activate the PANOPTES environment.
 
+  # shellcheck disable=SC1090
```
wtgee: What check is this disabling? Just curious since there are a few of them.

scripts/install/install-helper-functions.sh
```diff
@@ -0,0 +1,148 @@
...
+  fi
+}
+
+#-------------------------------------------------------------------------------
+# Misc helper functions.
+
+# Execute the args (a command) as root, either using sudo or directly if already
+# running as root.
+function my_sudo() {
+  if [ "$(id -u -n)" == "root" ] ; then
```
wtgee: If one is root then does running it with sudo not work? Just wondering why we can't always use sudo regardless of user. For Docker I just install sudo to make it work with default root user.

scripts/install/install-helper-functions.sh
```diff
@@ -0,0 +1,148 @@
...
+}
+
+# Does the first arg start with the second arg?
+function beginswith() { case "${1}" in "${2}"*) true;; *) false;; esac; }
+
+#-------------------------------------------------------------------------------
+# Helpers for finding the APT caching proxy.
+
+# 
+function proc_net_route_addr() {
```
wtgee: There's no way to query the docker command for this info?

scripts/install/install-helper-functions.sh
```diff
@@ -0,0 +1,148 @@
...
+  local -r -a words=( $(grep -E -i '^[a-z][a-z0-9]+\s00000000\s[0-9a-f][0-9a-f][0-9a-f][0-9a-f]' /proc/net/route ) )
+  local -r gateway="${words[2]}"
+  local -r a="0x${gateway:6:2}"
+  local -r b="0x${gateway:4:2}"
+  local -r c="0x${gateway:2:2}"
+  local -r d="0x${gateway:0:2}"
+  printf "%d.%d.%d.%d\n" "${a}" "${b}" "${c}" "${d}"
+}
+
+# Print the first IP on the route to 8.8.8.8, i.e. to Google's public DNS server?
```
wtgee: Don't think this is supposed to be a question.

scripts/install/configure-apt-cache.sh
jamessynge: Fixed.

scripts/install/configure-apt-cache.sh
jamessynge: That's the magic of using $() rather than backticks!

scripts/install/install-apt-packages.sh
jamessynge: Fixed.

scripts/install/install-apt-packages.sh
jamessynge: That would require adding a trusted cert to the proxy setup. This is only for the communication between apt-get and the proxy... IIUC. I'm hoping that the communication thereafter is over TLS, but don't know.

scripts/install/install-apt-packages.sh
jamessynge: Agreed, I had to learn this. Done.

scripts/install/install-dependencies.sh
jamessynge: Shellcheck supports checks that span multiple files, but only if it knows where that other file is, and it is complaining that it can't figure out the value of CONDA_SH and so doesn't know where conda is located.

Since we don't write the conda code, I'm disabling that feature here.

scripts/install/install-helper-functions.sh
jamessynge: Inside of docker sudo is not normally available. And it is recommended to NOT install sudo or su.

scripts/install/install-helper-functions.sh
jamessynge: Not that I know of. I think it is different for each container, and the docker command is not available inside the container. My research about the problem let to code very much like this.

scripts/install/install-helper-functions.sh
jamessynge: Right!

jamessynge:
I'm adding a minimal README.md to the install directory. Much work is needed if that is to be a place where we describe the install process, vs. a google doc.

PR #728, opened 2018-11-15
Use a power simulator as a default for now

wtgee:
Needed until the arduino side of the work is done for #697

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -149,9 +149,8 @@ class PocsShell(Cmd):
         simulator = listify(arg[0].split())
 
         # TODO(wtgee) Incorporate real power readings
-        simulator.append('power')
-
-        simulator = list(set(simulator))  # Make unique
+        if 'power' not in simulator:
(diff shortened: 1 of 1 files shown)


bin/pocs_shell
```diff
@@ -148,6 +148,11 @@ class PocsShell(Cmd):
         """Setup and initialize a POCS instance."""
         simulator = listify(arg[0].split())
 
+        # TODO(wtgee) Incorporate real power readings
```
jamessynge: This is OK for now, but I was wondering about adding an entry to the config for this, That makes it easy for "optional" features such as this to be enabled or disabled on a per site basis. Thoughts?

bin/pocs_shell
```diff
@@ -148,6 +148,11 @@ class PocsShell(Cmd):
...
+        simulator.append('power')
```
jamessynge: This could just be:
```
if 'power' not in simulator:
    simulator.append('power')
```

PR #730, opened 2018-11-17
Add ability to test both host and container for the APT caching proxy.

jamessynge:
Improve comments in install-helper-functions.sh.

### DIFF
--- a/scripts/install/install-helper-functions.sh
+++ b/scripts/install/install-helper-functions.sh
@@ -55,12 +55,15 @@ function safe_which() {
 function beginswith() { case "${1}" in "${2}"*) true;; *) false;; esac; }
 
 #-------------------------------------------------------------------------------
-# Helpers for finding the APT caching proxy.
+# Helpers for finding the IP gateway of this device.
 
-# 
-function proc_net_route_addr() {
+# Get the address of the gateway stored in /proc/net/route. Available inside
+# a docker container or in Ubuntu 18.04. When executed inside a container,
+# the address is that of the host; when executed on the host, the address is
+# that of the IP gateway on the lan segment (e.g. the adjacent router).
+function proc_net_route_gateway() {
   if [ ! -e /proc/net/route ] ; then
-    return
+    return 0
   fi
   # This can be done with awk, but that is then much harder to debug.
   # /proc/net/route has contents like this:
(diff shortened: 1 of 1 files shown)


scripts/install/install-helper-functions.sh
```diff
@@ -55,12 +55,15 @@ function safe_which() {
 function beginswith() { case "${1}" in "${2}"*) true;; *) false;; esac; }
 
 #-------------------------------------------------------------------------------
-# Helpers for finding the APT caching proxy.
+# Helpers for finding the IP gateway of this device.
 
-# 
-function proc_net_route_addr() {
+# Get the address of the gateway stored in /proc/net/route. Available inside
```
wtgee: Is this used on the build or when running? I'm thinking about the different docker networking stack on OSX and wondering if this will work. Seems like it's running inside the image so should be okay but just wondering. Also might just want to get it working first and then worry about OSX.

scripts/install/install-helper-functions.sh
```diff
@@ -55,12 +55,15 @@ function safe_which() {
...
+# a docker container or in Ubuntu 18.04. When executed inside a container,
```
wtgee: "...inside an image,..." (is that right?)

scripts/install/install-helper-functions.sh
```diff
@@ -78,15 +81,65 @@ function proc_net_route_addr() {
...
+function sbin_ip_host() {
+  local cmd
+  if [ -x "$(safe_which ip)" ] ; then
+    cmd='ip'
+  elif [ -x /sbin/ip ] ; then
+    cmd='/sbin/ip'
+  else
+    return 0
   fi
+  "${cmd}" route get 8.8.8.8 | awk -F"src " 'NR==1{split($2,a," ");print a[1]}'
```
wtgee: Seems like the only difference from `sbig_ip_gateway` is the work `src` here instead of `via`. Does it make sense to combine these two?

scripts/install/install-helper-functions.sh
jamessynge: I've not got OSX handy... maybe I should borrow an old laptop. Once I merge, perhaps you could test these functions on your mac?

scripts/install/install-helper-functions.sh
jamessynge: The image is the thing created by a builder and loaded into a container to execute (i.e. a filesystem snapshot). The container is where code executes (including during building, AFAICT).

scripts/install/install-helper-functions.sh
jamessynge: Agreed, that is only difference. Extracted out.

scripts/install/install-helper-functions.sh
wtgee: I don't have one either. Let's worry about it afterward.

scripts/install/install-helper-functions.sh
wtgee: Oh, I thought these were copied into the machine and run on the image. Thanks.

PR #735, opened 2018-11-23
Dithering utils

wtgee:
These are pulled from https://github.com/AstroHuntsman/huntsman-pocs

Besides some slight code cleanup the only difference is to specify a default `pattern_offset` of 30 arcminutes.

Note that these are placed in the utils folder and are not yet a part of normal operations.

### DIFF
--- a/pocs/utils/dither.py
+++ b/pocs/utils/dither.py
@@ -1,0 +1,95 @@
+import numpy as np
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+from astropy.coordinates import SkyOffsetFrame
+from astropy.coordinates import ICRS
+
+# Pattern for dice 9 3x3 grid (sequence of (RA offset, dec offset) pairs)
+dice9 = ((0, 0),
+         (0, 1),
+         (1, 1),
+         (1, 0),
+         (1, -1),
+         (0, -1),
+         (-1, -1),
+         (-1, 0),
+         (-1, 1))
+
+
+# Pattern for dice 5 grid (sequence of (RA offset, dec offset) pairs)
+dice5 = ((0, 0),
+         (1, 1),
+         (1, -1),
+         (-1, -1),
+         (-1, 1))
+
+
+def get_dither_positions(base_position,
+                         n_positions,
+                         pattern=None,
+                         pattern_offset=30 * u.arcminute,
+                         random_offset=None):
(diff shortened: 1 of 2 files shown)


pocs/utils/dither.py
```diff
@@ -0,0 +1,95 @@
...
+            raise ValueError(f"Base position '{base_position}' cannot be converted to a SkyCoord")
+
+    # Use provided pattern if given.
+    if pattern:
+        if not isinstance(pattern_offset, u.Quantity):
+            pattern_offset = pattern_offset * u.arcsec
+
+        # Get n_positions from the pattern
+        ra_offsets = [pattern[count % len(pattern)][0] for count in range(n_positions)]
+        dec_offsets = [pattern[count % len(pattern)][1] for count in range(n_positions)]
```
AnthonyHorton: For the sake of readability perhaps the `[count % len(pattern) for count in range(n_positions)]` should be extracted into a descriptively named utility function? I find these two lines a bit opaque as they stand, and I wrote them!

wtgee:
@AnthonyHorton I've added the plotting options back although changed slightly. Mostly I'm not using pyplot but instead generating an actual figure. As such I just return the figure and it is up to the user to display/save.

I also simplified it a bit so the only param to the plotting function is the generated positions. This means it does not show the given offsets as part of the legend or title.  Since the figure is returned to the user these could be added in after the fact if desired.

I've also added testing via `pytest-mpl`

wtgee:
I've wasted too much time on the dithering plot but did finally get the image upload of the failed images to work. Here's the output (baseline, created, diff). Needless to say I'm pretty frustrated. Note that this has the `remove_text=True` option so it should be ignoring the text but I guess that is only the title. Apparently there is some minor difference in the font style that is rending a 13% rms on the image diff. 😠 

baseline-test_plot_dither

test_plot_dither

test_plot_dither-failed-diff

.travis.yml
```diff
@@ -91,3 +91,5 @@ cache:
     - $PANDIR/astrometry/
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+after_failure:
+  - bash scripts/testing/failed-images-upload.sh ${POCS}/test_images/
```
wtgee: @jamessynge when you get a chance I would appreciate your thoughts on this.

PR #737, opened 2018-11-23
Don't overshoot the closed position on the first shutter.

jamessynge:
The feedback mechanism on the Astrohaven dome is rather limited:
it tells you whether a sensor on the lower shutter is tripped, but
nothing about the upper shutter. Since we were finding the dome
wasn't closing all the way reliably, overshooting seemed to
insure that the dome would close all the way. What we didn't
realize is that the variability in the straps and the sensor
could cause the overshoot to bend the lower shutter up and
over the pulley block on top of the fixed dome base, after
which we need to disassemble the dome to fix it. SIGH.

This was happening just for the first side to be closed (the
A side) as the other side couldn't close too far because it
would hit the top shutter of the first side. So, I've modified
the code so that it will stop immediately when it gets the
"closed" feedback from the first side, but then on the second
side will send one extra "close" command once the limit is
reached. Why? To avoid finding that the dome is almost closed,
but with a slight gap between the shutters.

Read SerialData params for a dome from a 'serial' dict.
As with AbstractSerialMount, this modifies AbstractSerialDome to
read the serial device configuration from an entry in the pocs*.yaml
file called 'serial' inside of the 'dome' entry.

Remove the default baudrate for the dome serial device from the
code and require that it be in the config.

### DIFF
--- a/pocs/dome/astrohaven.py
+++ b/pocs/dome/astrohaven.py
@@ -7,37 +7,76 @@ from pocs.dome import abstract_serial_dome
 
 
 class Protocol:
+    # yapf: disable
     # Status codes, produced when not responding to an input. They are oriented towards
     # reporting whether the two shutters are fully closed.
     BOTH_CLOSED = '0'  # Both A and B shutters are fully closed.
-
     A_IS_CLOSED = '1'  # Only shutter A is fully closed.
     B_IS_CLOSED = '2'  # Only shutter B is fully closed.
-
-    BOTH_OPEN = '3'    # Really means both NOT fully closed.
+    BOTH_OPEN   = '3'  # Really means both NOT fully closed.
 
     # Status codes produced by the dome when not responding to a movement command.
     STABLE_STATES = (BOTH_CLOSED, BOTH_OPEN, B_IS_CLOSED, A_IS_CLOSED)
 
     # Limit responses, when the limit has been reached on a direction of movement.
-    A_OPEN_LIMIT = 'x'  # Response to asking for A to open, and being at open limit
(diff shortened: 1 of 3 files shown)


pocs/dome/astrohaven.py
```diff
@@ -87,10 +126,10 @@ def is_closed(self):
         return v == Protocol.BOTH_CLOSED
 
     def close(self):
```
wtgee: The PR description could probably be placed in the docstring here.

pocs/dome/astrohaven.py
```diff
@@ -157,6 +191,11 @@ def _full_move(self, send, target_feedback, feedback_countdown=1):
             target_feedback: The response code to compare to the response from the dome;
                 this is a string of one ASCII character. See Protocol above for the codes;
                 while the dome is moving, it echoes the command code sent.
+            feedback_countdown: Number of times the target_feedback needs to be received
```
wtgee: Seems like most of this would be better placed in the `close` method rather than in a private method. But maybe you wanted to bury the details...

pocs/dome/astrohaven.py
wtgee: But I see some of it is below in `_full_move`

wtgee:
@jamessynge since you are no longer in the dome can you let me know what you would like to happen with this PR? Thanks!

jamessynge:
I'll try picking this up and completing it so that I can leave the AstroHaven dome code in good condition.

wtgee:
Closing as stale but will keep branch available.

PR #744, opened 2018-11-30
Fix FLI serial_number

wtgee:
A wayward comma was causing this to come across as a tuple.


wtgee:
> Looks like this slipped through the review in #620. The perils of code we can't test without hardware!

Yup, was my fault. I had originally placed all those variables in a dictionary and then decided to put them each on their own line and somehow missed this comma.  I didn't even notice this was the problem while observing last night so currently there is a hack-y work-around on each of the FLI-Pis.

PR #746, opened 2018-12-07
Make google bucket transfer script more generic

wtgee:
Change the helper script to be more generic so that transfers can go in either direction.

No change of behavior, just a rename of variables to make it clear that
the underlying `cp` works either way.

Added default value for MANIFEST_FILE.

I'm using this in panoptes/PIAA to download observations, where there is no specific unit_id so it cannot be incorporated into the `PanStorage` class.


scripts/transfer_files.sh
```diff
@@ -3,54 +3,56 @@
 usage() {
   echo -n "
 #############################################################
-# Upload files to Google Cloud Storage bucket.
+# Transer files to/from a Google Cloud Storage bucket.
```
jamessynge: c/Transer/Transfer/

scripts/transfer_files.sh
```diff
@@ -3,54 +3,56 @@
 usage() {
   echo -n "
 #############################################################
```
jamessynge: Why are these lines prefixed as if they were comments? It doesn't seem necessary.

scripts/transfer_files.sh
```diff
@@ -3,54 +3,56 @@
...
- $ $(basename $0) SEARCH_PATH DESTINATION MANIFEST_PATH
+ $ $(basename $0) FROM_PATH TO_PATH MANIFEST_FILE
  
  Options:
-  SEARCH_PATH     The string path to be searched.
-  DESTINATION     Remote upload destination.
-  MANIFEST_PATH   Path to manifest file for tracking uploads.
+  FROM_PATH     File(s) to be moved.
+  TO_PATH       Where file(s) are to be placed.
+  MANIFEST_FILE Path to manifest file for tracking uploads. Optional, defaults
```
jamessynge: If a manifest file already exists, will it be overwritten?

scripts/transfer_files.sh
```diff
@@ -3,54 +3,56 @@
...
-# a manifest log into the directory which can be used for upload
+# a manifest log into the directory which can be used for transfer
 # validation.
 #
 # For more information, see documentation at:
 # https://cloud.google.com/storage/docs/gsutil/commands/cp
 #
 # Note:
-# The SEARCH_PATH and DESTINATION are often similar. Normal invocation
+# The FROM_PATH and TO_PATH are often similar. Normal invocation
```
jamessynge: Maybe include an example or two, one of upload, the other of download?

scripts/transfer_files.sh
wtgee: That's a good question. I've just copied and pasted over from, e.g. `take_bias.sh`, `take_pic.sh` but it does seem to be a little odd. I thought I originally copied that from one of your bash scripts but I can't see what that would have been.  In the `take_pic.sh` and `take_bias.sh` it reads like a docstring with examples below the comments.  Here the final `NOTE:` line breaks that pattern.

scripts/transfer_files.sh
wtgee: I've been trying to determine if it makes more sense to use the `rsync` function instead of the `cp` but can't seem to find any indication of which would be preferred. With the `rsync` we wouldn't have the manifest file, which might be nice. Currently they are accidentally included in the upload so they also exist in the bucket (unless that is how it is supposed to work).

Thoughts?

jamessynge:
Did you want any further review in this?

wtgee:
> Did you want any further review in this?

Thanks for prompt. I guess I was just planning on fixing up the few items mentioned above as there is not really any functionality change. I'll take care of this a bit later today.

PR #752, opened 2018-12-09
Properly check for mongod server instance before providing client.

### LINKED ISSUE #751: conf_test can_connect_to_mongo broken
I'm on a fresh install and trying to run tests and forgot to install mongo. The underlying `pocs.utils.database.PanDB(db_type='mongo', connect=True)` will create and return an object despite the lack of mongo installed on the system. Tests hang forever trying to connect.

wtgee:
Update `get_shared_mongo_client` to properly check if server is actually available.

See [second Note](https://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient) in api docs for MongoClient.

**Behavior Change:** This now causes `PanMongoDB` to raise a `ConnectionError` if server is not available.

Closes #751.

### DIFF
--- a/pocs/utils/database.py
+++ b/pocs/utils/database.py
@@ -7,6 +7,7 @@ from warnings import warn
 from uuid import uuid4
 from glob import glob
 from bson.objectid import ObjectId
+from pymongo.errors import ConnectionFailure
 
 from pocs.utils import current_time
 from pocs.utils import serializers as json_util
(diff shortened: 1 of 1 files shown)


wtgee:
I've added a pragam to skip the coverage check on ConnectionFailure. I suppose ideally we should mock (the lack of) mongod or do something smarter.

PR #757, opened 2018-12-10
Fixing up the serial doctest.

wtgee:
I've made this separate from a more generic PR to clean up the doctests because some of this behavior confuses me so I want it to be discussed with @jamessynge.

I can't get this example to work without calling `ResetBuffers`, which seems a little odd to me. Perhaps it should be called automatically somewhere?

Also, the `device_listener.write()` will fail if given a byte-string as the underlying rs232 is calling an `encode` automatically. Could switch [`rs232.SerialData.write`](https://github.com/panoptes/POCS/blob/develop/pocs/utils/rs232.py#L211) to detect string type and skip, which would probably make it more generic.

That might make this test a little more consistent as it currently writes a normal string but returns (via `ReadFakeDevice`) a byte-string.

### DIFF
--- a/pocs/utils/rs232.py
+++ b/pocs/utils/rs232.py
@@ -54,8 +54,6 @@ class SerialData(object):
 
         # Register our serial simulators
         >>> serial.protocol_handler_packages.append('pocs.tests.serial_handlers')
-
-        # Create a fake device
         >>> from pocs.tests.serial_handlers.protocol_buffers import SetRBufferValue as WriteFakeDevice
         >>> from pocs.tests.serial_handlers.protocol_buffers import GetWBufferValue as ReadFakeDevice
         >>> from pocs.tests.serial_handlers.protocol_buffers import ResetBuffers
@@ -65,6 +63,9 @@ class SerialData(object):
 
         # Connect to our fake buffered device
         >>> device_listener = SerialData(port='buffers://')
+
+        # Note: A manual reset is currently required because implementation is not complete.
+        # See https://github.com/panoptes/POCS/issues/758 for details.
         >>> ResetBuffers()
         >>> device_listener.is_connected
         True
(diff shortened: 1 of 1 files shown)


jamessynge:
protocol_buffers.py and neighbors contains my first attempt at using the ability that pyserial offers to provide custom implementations. As a result it doesn't support concurrent tests in a single process, and requires tests to reset buffers explicitly. I learned a lot later, which is why the Arduino and Astrohaven simulators are so much more capable, including capable of concurrency.

rs232.SerialData has methods .read and .write that work in terms of strings, and methods .read_bytes and .write_bytes that work in terms of arrays of bytes. Unfortunately .write fails to specify when it calls encode that the encoding should be ASCII.

I'd keep this change as minimal as possible, and assign a bug to me to improve the test handlers. Thoughts?

pocs/utils/rs232.py
```diff
@@ -51,33 +51,38 @@ class SerialData(object):
...
-        >>> from pocs.tests.serial_handlers.protocol_buffers import GetWBuffer as ReadFakeDevice
+        >>> from pocs.tests.serial_handlers.protocol_buffers import GetWBufferValue as ReadFakeDevice
+        >>> from pocs.tests.serial_handlers.protocol_buffers import ResetBuffers
 
         # Import our serial utils
         >>> from pocs.utils.rs232 import SerialData
 
         # Connect to our fake buffered device
         >>> device_listener = SerialData(port='buffers://')
+        >>> ResetBuffers()
```
jamessynge: Perhaps add a note that only a single SerialData instance should open 'buffers://' at any one time because it is a very limited test fixture.

wtgee:
> I'd keep this change as minimal as possible, and assign a bug to me to improve the test handlers. Thoughts?

Sounds good, will do. My goal was mostly to get the doctests integrated into pytest (and travis) automatically and this was just giving me some slight hassles in terms of cleaning it up. Really that is only because I am unfamiliar with how it works. The real answer should probably be that I learn what is going on. Thanks for explanation.

PR #761, opened 2018-12-11
Doctest fixes - WIP

wtgee:
A largish PR to clean up issues with sphinx building the docs.

PR #763, opened 2018-12-11
Make the `_all_loggers` a global variable.

wtgee:
@jamessynge take a look at this and tell me if you see any repercussions.  This was the real culprit behind all the sphinx build problems in #761. Might also explain some of our sometime issues with logging and the lack of proper formatter.

### DIFF
--- a/pocs/utils/logger.py
+++ b/pocs/utils/logger.py
@@ -16,7 +16,7 @@ from pocs.utils.config import load_config
 # We don't want to create multiple root loggers that are "identical",
 # so track the loggers in a dict keyed by a tuple of:
 #    (profile, json_serialized_logger_config).
-all_loggers = {}
+_all_loggers = {}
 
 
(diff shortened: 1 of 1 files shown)


wtgee:
Actually it's still unclear to me if this is the problem for the sphinx build. It seemed to fix it while testing but then I created this PR in a separate branch and when I got back to #761 (and cherry-pick this commit) it doesn't work. Same thing happened with cherry-picking #762.

But still curious as to whether this should be a global or not.

jamessynge:
I think this is a no-op. A global declaration is really only needed if you want to assign to a global variable.

wtgee:
> I think this is a no-op. A global declaration is really only needed if you want to assign to a global variable.

I didn't know if it was needed to assign a key to a global but I suppose not. And, as mentioned, didn't seem to be the fix anyway. Thanks for looking, will close.

PR #766, opened 2018-12-11
Support max_nexp

### LINKED ISSUE #749: Targets File - Support max_nexp
Targets currently have a `min_nexp` property but there are some cases, such as with the pretty images of a comet, where we don't necessarily want to dwell on it forever but want to specify.

wtgee:
Add the ability to limit the maximum number of exposures taken during an observation. Note that this does not prevent the Observation from being rescheduled.

**Behavior Change** This changes the default `exp_set_size=1` so that essentially any combination of exposures is allowed by default.

Closes #749.

### DIFF
--- a/pocs/scheduler/observation.py
+++ b/pocs/scheduler/observation.py
@@ -10,7 +10,7 @@ class Observation(PanBase):
 
     @u.quantity_input(exp_time=u.second)
     def __init__(self, field, exp_time=120 * u.second, min_nexp=60, max_nexp=None,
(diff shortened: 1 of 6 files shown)


wtgee:
Thanks for the approval. There are a few kinks to work out with this regarding how the observations reset so I'm not sure it's ready to merge yet. I'll try to get back to it this weekend.

wtgee:
It's not clear we want to use this versus just creating a "follow-up" scheduler mode. This would also need further checks to insure it works properly. Closing for now.

Issue #768, opened 2019-01-14
Add support for filter wheels


AnthonyHorton:
Proposal to add support for optical filter exchange mechanisms (e.g. filter wheels). POCS supports several types of monochrome camera (SBIG, FLI), which are typically used with some sort of filter wheel in order to perform multiband imaging. Currently these filter wheels are unsupported, though.

In the first instance it is proposed to create an `AbstractFilterWheel` base class, create a functional `FilterWheel` class for the SBIG filter wheels, and modify the `AbstractCameraClass` to make it possible to associate a filter wheel with a camera, in the same way as it is currently done with focusers.

In order to make full use of the ability to change filters under computer control more extensive changes would be needed, e.g. `Observation`s should optionally include the name(s) of the desired filter for the exposures. These changes can go into `huntsman-pocs` for now, at least.

AnthonyHorton:
#769 (#776) implement basic support for SBIG filter wheels, along with a base class and simulator. The modifications to `Observation` and `Camera` to allow automated use of filter wheels are still to be done, though.

AnthonyHorton:
Closing this support for filterwheels has been added, and opening a new Issue (#901) specifically about support for an additional brand of filterwheel.

PR #769, opened 2019-01-30
Add support for filter wheels

AnthonyHorton:
Addresses #768 by adding support for motorised optical filter wheels. The scope of this PR only includes support for the SBIG filter wheels from Diffraction Limited but with the ground work in place it would not be difficult to add support for other filter wheels.

The design follows the pattern established with the `Camera` and `Focuser` classes. There is an `AbstractFilterWheel` base class as well as implementations of the `FilterWheel` class for both the SBIG filter wheels and simulated hardware. Properties, method names and behaviours have been chosen to be consistent with the `Camera` and `Focuser` classes.

~~Flagging as WIP only because there are a few unrelated changes in here (line length fixes & other minor tweaks) that I need to split out into their own PRs. Otherwise~~ I think this is basically ready to go.

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -412,9 +412,9 @@ class AbstractCamera(PanBase):
         header.set('CAM-MOD', self.model, 'Camera model')
 
         if self.focuser:
-            header = self.focuser._fits_header(header)
+            header = self.focuser._add_fits_keywords(header)
         if self.filterwheel:
-            header = self.filterwheel._fits_header(header)
+            header = self.filterwheel._add_fits_keywords(header)
 
         return header
 
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -18,6 +18,7 @@ from pocs.utils import load_module
 from pocs.utils import images as img_utils
 from pocs.utils.images import fits as fits_utils
 from pocs.focuser import AbstractFocuser
+from pocs.filterwheel import AbstractFilterWheel
 
 
 class AbstractCamera(PanBase):
@@ -26,7 +27,9 @@ class AbstractCamera(PanBase):
 
     Attributes:
         filter_type (str): Type of filter attached to camera, default RGGB.
-        focuser (`pocs.cameras.focuser.Focuser`|None): Focuser for the camera, default None.
+        focuser (`pocs.focuser.*.Focuser`|None): Focuser for the camera, default None.
+        filter_wheel (`pocs.filterwheel.*.FilterWheel`|None): Filter wheel for the camera, default
+            None.
         is_primary (bool): If this camera is the primary camera for the system, default False.
         model (str): The model of camera, such as 'gphoto2', 'sbig', etc. Default 'simulator'.
         name (str): Name of the camera, default 'Generic Camera'.
@@ -40,6 +43,7 @@ class AbstractCamera(PanBase):
                  port=None,
                  primary=False,
                  focuser=None,
+                 filterwheel=None,
                  *args, **kwargs):
         super().__init__(*args, **kwargs)
 
@@ -57,29 +61,14 @@ class AbstractCamera(PanBase):
         self._file_extension = kwargs.get('file_extension', 'fits')
         self._current_observation = None
 
-        if focuser:
-            if isinstance(focuser, AbstractFocuser):
-                self.logger.debug("Focuser received: {}".format(focuser))
-                self.focuser = focuser
-                self.focuser.camera = self
-            elif isinstance(focuser, dict):
-                try:
-                    module = load_module('pocs.focuser.{}'.format(focuser['model']))
-                except AttributeError as err:
-                    self.logger.critical("Couldn't import Focuser module {}!".format(module))
-                    raise err
-                else:
-                    focuser_kwargs = copy.copy(focuser)
-                    focuser_kwargs.update({'camera': self, 'config': self.config})
-                    self.focuser = module.Focuser(**focuser_kwargs)
-            else:
-                # Should have been passed either a Focuser instance or a dict with Focuser
-                # configuration. Got something else...
-                self.logger.error(
-                    "Expected either a Focuser instance or dict, got {}".format(focuser))
-                self.focuser = None
-        else:
-            self.focuser = None
+        self._create_submodule(subcomponent=focuser,
+                               name='focuser',
+                               class_name='Focuser',
+                               base_class=AbstractFocuser)
+        self._create_submodule(subcomponent=filter_wheel,
+                               name='filterwheel',
+                               class_name='FilterWheel',
+                               base_class=AbstractFilterWheel)
 
         self.logger.debug('Camera created: {}'.format(name))
 
@@ -495,6 +484,36 @@ class AbstractCamera(PanBase):
         self.logger.debug("Updating FITS headers: {}".format(file_path))
         fits_utils.update_headers(file_path, info)
         return file_path
+
+    def _create_subcomponent(self, subcomponent, name, class_name, base_class):
+        """
+
+        """
+        if subcomponent:
+            if isinstance(subcomponent, base_class):
+                self.logger.debug("{} received: {}".format(class_name, subcomponent))
+                setattr(self, name, subcomponent)
+                getattr(self, name).camera = self
+            elif isinstance(subcomponent, dict):
+                module_name = 'pocs.{}.{}'.format(name, subcomponent['model'])
+                try:
+                    module = load_module(module_name)
+                except AttributeError as err:
+                    self.logger.critical("Couldn't import {} module {}!".format(
+                        class_name, module_name))
+                    raise err
+                else:
+                    subcomponent_kwargs = copy.copy(subcomponent)
+                    subcomponent_kwargs.update({'camera': self, 'config': self.config})
+                    setattr(self, name, getattr(module, class_name)(**subcomponent_kwargs))
+            else:
+                # Should have been passed either a instance of base_class or dict with subcomponent
+                # configuration. Got something else...
+                self.logger.error(
+                    "Expected either a {} instance or dict, got {}".format(class_name, focuser))
+                setattr(self, name, None)
+        else:
+            setattr(self, name, None)
 
     def __str__(self):
         name = self.name
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -522,8 +522,8 @@ class AbstractCamera(PanBase):
             else:
                 # Should have been passed either a instance of base_class or dict with subcomponent
                 # configuration. Got something else...
-                self.logger.error(
-                    "Expected either a {} instance or dict, got {}".format(class_name, focuser))
+                self.logger.error("Expected either a {} instance or dict, got {}".format(
+                    class_name, subcomponent))
                 setattr(self, name, None)
         else:
             setattr(self, name, None)
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -61,14 +61,14 @@ class AbstractCamera(PanBase):
         self._file_extension = kwargs.get('file_extension', 'fits')
         self._current_observation = None
 
-        self._create_submodule(subcomponent=focuser,
-                               name='focuser',
-                               class_name='Focuser',
-                               base_class=AbstractFocuser)
-        self._create_submodule(subcomponent=filter_wheel,
-                               name='filterwheel',
-                               class_name='FilterWheel',
-                               base_class=AbstractFilterWheel)
+        self._create_subcomponent(subcomponent=focuser,
+                                  name='focuser',
+                                  class_name='Focuser',
+                                  base_class=AbstractFocuser)
+        self._create_subcomponent(subcomponent=filterwheel,
+                                  name='filterwheel',
+                                  class_name='FilterWheel',
+                                  base_class=AbstractFilterWheel)
 
         self.logger.debug('Camera created: {}'.format(name))
 
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -62,11 +62,11 @@ class AbstractCamera(PanBase):
         self._current_observation = None
 
         self._create_subcomponent(subcomponent=focuser,
-                                  name='focuser',
+                                  sub_name='focuser',
                                   class_name='Focuser',
                                   base_class=AbstractFocuser)
         self._create_subcomponent(subcomponent=filterwheel,
-                                  name='filterwheel',
+                                  sub_name='filterwheel',
                                   class_name='FilterWheel',
                                   base_class=AbstractFilterWheel)
 
@@ -512,7 +512,7 @@ class AbstractCamera(PanBase):
                 subcomponent, e.g. `pocs.focuser`
             class_name (str): name of the subcomponent class, e.g. 'Focuser'
             base_class (class): the base class for the subcomponent, e.g.
-                `pocs.focuser.AbtractFocuser1, used to check whether subcomponent is an instance.
+                `pocs.focuser.AbtractFocuser`, used to check whether subcomponent is an instance.
         """
         if subcomponent:
             if isinstance(subcomponent, base_class):
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -70,7 +70,7 @@ class AbstractCamera(PanBase):
                                   class_name='FilterWheel',
                                   base_class=AbstractFilterWheel)
 
-        self.logger.debug('Camera created: {}'.format(name))
+        self.logger.debug('Camera created: {}'.format(self))
 
 ##################################################################################################
 # Properties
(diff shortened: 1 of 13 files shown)


AnthonyHorton:
The Arduino tests are causing trouble again. `test_create_simulator_small_read_buffer` failed, and in the process spat out so many log messages that Travis is refusing to display the whole log.

```
>       report_num = reading['report_num']
E       KeyError: 'report_num'
pocs/tests/test_arduino_io.py:210: KeyError
```

wtgee:
> Unrelated changes now pulled out into #770, #771 & #772. Once those are merged this should be ready for review.

I'll wait for the merge of the others (and maybe remove the WIP if appropriate).

pocs/camera/camera.py
```diff
@@ -498,17 +498,29 @@ def _process_fits(self, file_path, info):
...
 
+        Args:
+            subcomponent (instance of class_name | dict): the subcomponent object, or the keyword
+                arguments required to create it.
+            sub_name (str): name of the subcomponent, e.g. 'focuser'. Will be used as the attribute
+                name, and must also match the name corresponding POCS submodule for this
+                subcomponent, e.g. `pocs.focuser`
+            class_name (str): name of the subcomponent class, e.g. 'Focuser'
+            base_class (class): the base class for the subcomponent, e.g.
+                `pocs.focuser.AbtractFocuser1, used to check whether subcomponent is an instance.
```
wtgee: Trailing `1` instead of backtick

pocs/camera/sbigudrv.py
```diff
@@ -317,14 +319,244 @@ def take_exposure(self,
...
+                a Quantity with time units. If a numeric type without units is given seconds will be
+                assumed.
+
+        Raises:
+            RuntimeError: raised if the driver returns an error or if the final status and position
+                are not as expected.
+            pocs.utils.error.Timeout: raised if the move does not end within the period of time
+                specified by the timeout argument.
         """
+        if timeout is not None:
```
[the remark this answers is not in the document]
AnthonyHorton: Oh, that looks handy!

pocs/filterwheel/sbig.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _move_to(self, position, move_event):
+        self._SBIGDriver.cfw_goto(handle=self._handle,
+                                  position=position,
+                                  cfw_event=move_event,
+                                  timeout=self._timeout)
+
+    def _fits_header(self, header):
+        header = super()._fits_header(header)
+        header.set('FW-FW', self.firmware_version, 'Filter wheel firmware version')
```
wtgee: Hm...that's an awkward keyword. ;)

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+
+class AbstractFilterWheel(PanBase):
+    """
+    Base class for all filter wheels
+
+    Args:
+        name (str, optional): name of the filter wheel
+        model (str, optional): model of the filter wheel
+        camera (pocs.camera.*.Camera, optional): camera that this filter wheel is associated with.
+        filter_names (list of str): names of the filters installed at each filter wheel position
```
wtgee: I can imagine a scenario where I want to pass a `dict` of names/positions but might not be necessary.

pocs/filterwheel/simulator.py
```diff
@@ -0,0 +1,119 @@
...
+        camera (pocs.camera.*.Camera, optional): camera that this filter wheel is associated with.
+        filter_names (list of str): names of the filters installed at each filter wheel position
+        timeout (u.Quantity, optional): maximum time to wait for a move to complete. Should be
+            a Quantity with time units. If a numeric type without units is given seconds will be
+            assumed. Default is 10 seconds.
+        serial_number (str): serial number of the filter wheel
+        move_time (astropy.units.Quantity, optional): time to move the filter wheel by one position,
+            optional, default 1 second.
+        move_bidirectional (bool, optional): if True will simulate filter wheel which can rotate in
+            either direction, if False (default) will similate a filter wheel that only moves in
```
wtgee: `simulate`

Also, is this only a feature of the simulator? I didn't see anything about this in the actual code.

pocs/filterwheel/simulator.py
```diff
@@ -0,0 +1,119 @@
...
+            msg = "Attempt to move filter wheel when already moving"
+            self.logger.error(msg)
+            raise RuntimeError(msg)
+
+        move_distance = position - self.position
+        if move_distance < 0:
+            if not self._move_birectional:
+                move_distance += self._n_positions
+            else:
+                move_distance = -move_distance
```
wtgee: Can you replace the above with a simple `move_distance = position % self._n_positions`?

pocs/filterwheel/simulator.py
```diff
@@ -0,0 +1,119 @@
...
+        move_duration = move_distance * self._move_time
+
+        move = threading.Timer(interval=move_duration,
+                               function=self._complete_move,
+                               args=(position, move_event))
+        move.start()
```
wtgee: Just to be safe it seems like this line should go below the next two lines, since they set the condition the `move` timer changes. If for some reason the time got set to zero move time this could feasibly get stuck.

pocs/tests/test_filterwheel.py
```diff
@@ -0,0 +1,148 @@
...
+def test_init(filterwheel):
+    assert isinstance(filterwheel, SimFilterWheel)
+    assert filterwheel.is_connected
+
+
+def test_camera_init():
+    sim_camera = SimCamera(filterwheel={'model': 'simulator',
+                                        'filter_names': ['one', 'deux', 'drei', 'quattro']})
+    assert isinstance(sim_camera.filterwheel, SimFilterWheel)
+    assert sim_camera.filterwheel.is_connected
```
wtgee: Maybe we want to check these values in the case when there is no filterwheel as well?

pocs/camera/sbigudrv.py
```diff
@@ -317,14 +319,244 @@ def take_exposure(self,
...
+            start_time = time.time()
+            if isinstance(timeout, u.Quantity):
+                timeout = timeout.to(u.second).value
+
+        try:
+            query = self.cfw_query(handle, model)
+            while query['status'] == 'BUSY':
+                if timeout is not None and (time.time() - start_time) > timeout:
+                    msg = "Timeout waiting for filter wheel on {} move to complete".format(handle)
```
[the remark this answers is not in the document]
AnthonyHorton: Might be handy. It would be in the preceding log messages, but only if logging at DEBUG level. I'll put it in here too.

pocs/tests/test_filterwheel.py
```diff
@@ -0,0 +1,148 @@
...
+    with pytest.raises(AttributeError):
+        filterwheel.filter_names = ["Unsharp mask", "Gaussian blur"]
+
+# Movement
+
+
+def test_move_number(filterwheel):
+    assert filterwheel.position == 1
+    e = filterwheel.move_to(2)
+    assert math.isnan(filterwheel.position)  # position is NaN while between filters
```
wtgee: Should this move a larger position so there is no chance of some race-condition (not that there probably is anyway)

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+        """ Number of positions in the filter wheel """
+        return self._n_positions
+
+    @property
+    def position(self):
+        """ Current integer position of the filter wheel """
+        raise NotImplementedError
+
+    @position.setter
+    def position(self, position):
```
wtgee: The setter on the position, while convenient, seems somewhat dangerous to me. I'm not suggesting we do away with it but seems like you could easily accidentally set this.

pocs/tests/test_filterwheel.py
```diff
@@ -0,0 +1,148 @@
...
+    e.wait()
+    assert filterwheel.position == 2
+    e = filterwheel.move_to(3, blocking=True)
+    assert e.is_set()
+    assert filterwheel.position == 3
+    filterwheel.position = 4  # Move by assignment to position property blocks until complete
+    assert filterwheel.position == 4
+
+
+def test_move_bad_number(filterwheel):
```
wtgee: Can you test at `n_positions` as well for both edge cases?

pocs/tests/test_filterwheel.py
```diff
@@ -0,0 +1,148 @@
...
+    with pytest.raises(ValueError):
+        filterwheel.move_to(0, blocking=True)  # No zero based numbering here!
+    with pytest.raises(ValueError):
+        filterwheel.position = 99
+
+
+def test_move_name(filterwheel, caplog):
```
wtgee: Wha?! How did I not know about `caplog` fixture?! That's nice.

pocs/camera/sbigudrv.py
```diff
@@ -317,14 +319,244 @@ def take_exposure(self,
...
+        cfw_info_params = CFWParams(CFWModelSelect[model],
+                                    CFWCommand.GET_INFO,
+                                    CFWGetInfoSelect.FIRMWARE_VERSION)
+        cfw_info_results = CFWResults()
+        with self._command_lock:
+            self._set_handle(handle)
+            self._send_command('CC_CFW', cfw_info_params, cfw_info_results)
+
+        results = self._cfw_parse_results(cfw_info_results)
+        results = {'model': CFWModelSelect(cfw_info_results.cfwModel).name,
```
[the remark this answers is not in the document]
AnthonyHorton: It does. The previous line shouldn't be there any more. Discovered in testing that the `GET_INFO` command doesn't return all the same values as the `QUERY` command, despite having fields in the results struct for them, so there's no point in calling `_cfw_parse_results()`

pocs/camera/sbigudrv.py
```diff
@@ -317,14 +319,244 @@ def take_exposure(self,
...
+                asks the driver to autodetect the model.
+
+        Returns:
+            dict: dictionary containing the 'model', 'position', 'status' and 'error' values
+                returned by the driver.
+
+        Raises:
+            RuntimeError: raised if the driver returns an error
+        """
+        cfw_query_params = CFWParams(CFWModelSelect[model],
```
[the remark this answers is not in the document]
AnthonyHorton: Fair point. I've pulled out the common stuff into a 6 line helper function.

pocs/camera/sbigudrv.py
```diff
@@ -1289,3 +1533,109 @@ class GetDriverInfoResults0(ctypes.Structure):
     _fields_ = [('version', ctypes.c_ushort),
                 ('name', ctypes.c_char * 64),
                 ('maxRequest', ctypes.c_ushort)]
+
+
+#################################################################################
+# Filter wheel related
```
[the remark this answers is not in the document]
AnthonyHorton: The following `Structure`s and `Enum`s are used in this file, and only in this file, so I think it's reasonable for them to be in here.

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+    @property
+    def is_connected(self):
+        """ Is the filterwheel available """
+        return self._connected
+
+    @property
+    def camera(self):
+        """
+        Reference to the Camera object that the FilterWheel is assigned to, if any. A filter wheel
+        should only ever be assigned to one or zero Cameras!
```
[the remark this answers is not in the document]
AnthonyHorton: The setter for the `FilterWheel` class `camera` property will log a `WARNING` and ignore you if you try to set the `camera` on a filterwheel that already has a `camera`, if that's what you mean.

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+    def position(self, position):
+        self.move_to(position, blocking=True)
+
+    @property
+    def current_filter(self):
+        """ Name of the filter in the current position """
+        try:
+            filter_name = self.filter_names[self.position - 1]  # 1 based numbering
+        except (IndexError, TypeError):
+            # Some filter wheels sometimes cannot return their current position
```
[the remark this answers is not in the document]
AnthonyHorton: The expectation is that this almost always means that the filter wheel is in between filter positions, hopefully on its way to where it was just asked to move to.

Under those circumstances there's not much that an internal software position tracker could add, and `UNKNOWN` is probably the most appropriate response to the question of 'which filter is currently in position'.

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+            for i, filter_name in enumerate(self.filter_names):
+                if filter_name.startswith(position):
+                    if int_position is None:
+                        int_position = i + 1  # 1 based numbering for filter wheel positions
+                    else:
+                        # Already matched at least once
+                        msg = "More than one filter name matches '{}', using '{}'".format(
+                            position, self.filter_names[int_position - 1])
+                        self.logger.warning(msg)
+                        break
```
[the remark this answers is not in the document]
AnthonyHorton: Purely so that I can log a warning about multiple matches. Ambigious filter wheel `move_to()` commands probably indicate a mistake somewhere, so shouldn't pass entirely silently.

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+        raise NotImplementError
+
+    def move_to(self, position, blocking=False):
+        """
+        Move the filter wheel to the given position.
+
+        The position can be expressed either as an integer, or as (part of) one of the names from
+        the filter_names list. To allow filter names of the form '<filter band>_<serial number>'
+        to be selected by band only position can be a substring from the start of one
+        of the names in the filter_names list, provided that this produces only one match.
```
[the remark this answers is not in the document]
AnthonyHorton: OK.

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+                raise ValueError(msg)
+
+        if int_position < 1 or int_position > self.n_positions:
+            msg = "Position must be between 1 and {}, got {}".format(self.n_positions, int_position)
+            self.logger.error(msg)
+            raise ValueError(msg)
+
+        return int_position
+
+    def _fits_header(self, header):
```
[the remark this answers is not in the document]
AnthonyHorton: Good point. Will do.

pocs/filterwheel/filterwheel.py
```diff
@@ -0,0 +1,224 @@
...
+        header.set('FW-NAME', self.name, 'Filter wheel name')
+        header.set('FW-MOD', self.model, 'Filter wheel model')
```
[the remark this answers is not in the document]
AnthonyHorton: Chosen for consistency with `CAM-MOD` and `FOC-MOD` that I already have from the `Camera` and `Focuser` classes.

pocs/filterwheel/simulator.py
AnthonyHorton: Yes, it's used to determine how long a simulated move should take, based on whether the filter wheel can move directly to towards the requested position or whether it has to go the long way around because it only rotates in one direction.

Outside of the simulator this is an implementation detail of the filter wheel itself, so the POCS code doesn't need to know about it.

Filter wheels that only rotate in one direction are pretty common, as it improves repeatability. The SBIG filter wheel that I've been testing with is like that, but the only reason I could tell was by watching it.

pocs/filterwheel/simulator.py
AnthonyHorton: Sort of...
```
move_distance = position - self.position
if self._move_bidirectional:
    move_distance = abs(move_distance)
else:
    move_distance = move_distance & self._n_positions
```
is equivalent, 1 line shorter, and arguably clearer.

pocs/filterwheel/simulator.py
AnthonyHorton: OK. Setting the `Timer` with zero interval does seem to cause trouble, which is one of the reasons I added some code to the higher level `move_to()` method in the base class to prevent the `_move_to()` method from being called for a 'move' to the position that the filter wheel is already in.

pocs/filterwheel/filterwheel.py
AnthonyHorton: I'm not seeing a need right now. Can add that option later, if it does turn out to be necessary.

pocs/tests/test_filterwheel.py
AnthonyHorton: Well, a camera with no filterwheel will have a `filterwheel` attribute (thanks to the `_create_subcomponent()` call) but the value will be `None`. Adding a test for that.

pocs/filterwheel/filterwheel.py
AnthonyHorton: It's really just there as a convenience during interactive use (e.g. testing in an IPython shell). The intent is that you almost always use the `.move_to()` method directly so that you can collect the events and move multiple filter wheels in parallel with each other. Perhaps the documentation should say that.

pocs/tests/test_filterwheel.py
AnthonyHorton: It is. I was forced into finding out how to do this because exceptions raised in a `Thread` don't make it out to the calling thread, so the log messages were the only way I could test that the timeout was happening in `test_move_timeout`.

pocs/tests/test_filterwheel.py
AnthonyHorton: OK

PR #771, opened 2019-01-30
Camera refactoring

AnthonyHorton:
Minor refactoring of Camera class, prompted by work on #769 (and mostly pulled out of that PR)

- modifications to SBIG & FLI `Camera` initialisation to allow serial number to be specified in `serial_number` keyword argument instead of having to use `port` for something that isn't a port, & to reduce appearance of placeholder UID ('XXXXXX') in the logs.
- convert the `Focuser` attribute creation code to a generic subcomponent creation method of `Camera`
- move calls to `Focuser` FITS header method to base class
- clean up of `__str__` methods

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -58,7 +58,7 @@ class AbstractCamera(PanBase):
         self._current_observation = None
 
         self._create_subcomponent(subcomponent=focuser,
-                                  name='focuser',
+                                  sub_name='focuser',
                                   class_name='Focuser',
                                   base_class=AbstractFocuser)
 
@@ -480,17 +480,29 @@ class AbstractCamera(PanBase):
         fits_utils.update_headers(file_path, info)
         return file_path
 
-    def _create_subcomponent(self, subcomponent, name, class_name, base_class):
+    def _create_subcomponent(self, subcomponent, sub_name, class_name, base_class):
         """
+        Creates a subcomponent as an attribute of the camera. Can do this from either an instance
+        of the appropriate subcomponent class, or from a dictionary of keyword arguments for the
+        subcomponent class' constructor.
 
+        Args:
+            subcomponent (instance of class_name | dict): the subcomponent object, or the keyword
+                arguments required to create it.
+            sub_name (str): name of the subcomponent, e.g. 'focuser'. Will be used as the attribute
+                name, and must also match the name corresponding POCS submodule for this
+                subcomponent, e.g. `pocs.focuser`
+            class_name (str): name of the subcomponent class, e.g. 'Focuser'
+            base_class (class): the base class for the subcomponent, e.g.
+                `pocs.focuser.AbtractFocuser1, used to check whether subcomponent is an instance.
         """
         if subcomponent:
             if isinstance(subcomponent, base_class):
                 self.logger.debug("{} received: {}".format(class_name, subcomponent))
-                setattr(self, name, subcomponent)
-                getattr(self, name).camera = self
+                setattr(self, sub_name, subcomponent)
+                getattr(self, sub_name).camera = self
             elif isinstance(subcomponent, dict):
-                module_name = 'pocs.{}.{}'.format(name, subcomponent['model'])
+                module_name = 'pocs.{}.{}'.format(sub_name, subcomponent['model'])
                 try:
                     module = load_module(module_name)
                 except AttributeError as err:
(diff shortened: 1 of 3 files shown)


pocs/camera/camera.py
```diff
@@ -496,14 +480,44 @@ def _process_fits(self, file_path, info):
         fits_utils.update_headers(file_path, info)
         return file_path
 
+    def _create_subcomponent(self, subcomponent, name, class_name, base_class):
```
wtgee: It's nice that this is split into a separate method as it is easier to read. Do you imagine it will be used more than once or by different subcomponents within one instance?

pocs/camera/camera.py
```diff
@@ -496,14 +480,44 @@ def _process_fits(self, file_path, info):
...
+        """
+
```
wtgee: Maybe a quick docstring about what this does.

pocs/camera/camera.py
```diff
@@ -496,14 +480,44 @@ def _process_fits(self, file_path, info):
...
+                setattr(self, name, None)
+        else:
+            setattr(self, name, None)
+
     def __str__(self):
         name = self.name
         if self.is_primary:
             name += ' [Primary]'
 
         s = "{} ({}) on {}".format(name, self.uid, self.port)
```
wtgee: With the default `name=None` this string might look weird.

pocs/camera/fli.py
```diff
@@ -69,21 +72,23 @@ def __init__(self,
...
             except KeyError:
-                message = 'Could not find {} ({})!'.format(self.name, self.port)
+                message = 'Could not find {} ({})!'.format(name, kwargs['serial_number'])
                 self.logger.error(message)
                 warn(message)
                 return
-            self.logger.debug('Found {} ({}) on {}'.format(self.name, self.port, device_node))
-            self.port = device_node
+            self.logger.debug('Found {} ({}) on {}'.format(
+                name, kwargs['serial_numner'], device_node))
```
wtgee: `serial_numner` -> `serial_number`

pocs/camera/camera.py
AnthonyHorton: Different subcomponents. The motivation for doing this was the introduction of filter wheel subcomponents in #769. I wanted to reuse the code for focuser subcomponent creation to create the filter wheel subcomponent so I needed to convert it into a generic method.

pocs/camera/camera.py
AnthonyHorton: In `AbstractCamera` the default for the `name` attribute is 'Generic Camera', and the SBIG and FLI `Camera` classes replace that with their own defaults ('SBIG Camera' and 'FLI Camera').  Might end up looking a bit odd with Canon cameras though, which don't change the default from 'Generic Camera'.

pocs/camera/fli.py
AnthonyHorton: Oops! This is the sort of thing that slips though when you're modifying hardware dependent code without the hardware on hand to test it.

pocs/camera/camera.py

pocs/camera/camera.py
wtgee: D'oh, yeah, sorry, I had a brain lapse and was thinking the `setattr` from above would overwrite that, but they are using the variable `name`, not the "name".

AnthonyHorton: Ah, yes. I should probably make the variable name in the method about something more specific than `name` to avoid that sort of confusion.

AnthonyHorton:
Kind of in two minds about retaining the ability to use `port` when you mean `serial_number` for the SBIG and FLI camera classes. The only reason to keep this at this point is so that existing Huntsman Telescope config files don't break.

Similarly the rather confusing terminology around filters, copied from the Canon cameras (where it makes sense). The camera's attribute/property is `filter_type`, which is a fine name when what you want to know whether you've got an `RGGB` or `CRGB` Bayer filter, or maybe a monochrome camera, but is an odd name for astronomy bandpass filters.  Maybe change to a more general `filter` property, but retain `filter_type` as an alias to avoid breaking stuff?

pocs/camera/camera.py
AnthonyHorton: Renamed to `name` to `sub_name` in order to be a bit less ambiguous.

wtgee:
> Similarly the rather confusing terminology around filters, copied from the Canon cameras (where it makes sense). The camera's attribute/property is `filter_type`, which is a fine name when what you want to know whether you've got an `RGGB` or `CRGB` Bayer filter, or maybe a monochrome camera, but is an odd name for a specific astronomy bandpass filter (including serial number). Maybe change to a more general `filter` property, but retain `filter_type` as an alias to avoid breaking stuff?

Maxim DL uses `BAYERPAT` to indicate the actual bayer pattern, which could then be used with a `FILTER=Bayer`.

I'm okay with the change either way. We are not really using that keyword in PANOPTES at all right now.

AnthonyHorton:
For now I think I'll leave things as they are. The important is that something sensible ends up against the `FILTER` keyword of the FITS files, and that is already happening. The only issue is that the terminology in the config files can be a bit confusing.

PR #773, opened 2019-01-31
Handle the weather data when PanDB is used with type of `file`.

wtgee:
Note: The name of the json file must be passed to the script with the `-f` parameter (probably in the crontab)


wtgee:
Ah! Thanks. I sort of forgot I already fixed this problem. :)

PR #776, opened 2019-02-01
Prevent filter wheel moving during a camera exposure

### LINKED ISSUE #775: Filterwheel should not work while exposing.
Need a safety check so that filterwheel is inactive during exposing.

AnthonyHorton:
Fixes #775 by preventing a filter wheel from moving if the associated camera is currently taking an exposure. In the process added an `is_exposing` property to the `AbstractCamera` base class, with implementations for the SBIG, FLI and simulator `Camera` classes.

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -167,6 +167,11 @@ class AbstractCamera(PanBase):
         else:
             return self._filter_type
 
+    @property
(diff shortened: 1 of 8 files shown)


AnthonyHorton:
Occurs to me that we should also prevent the camera starting an exposure while its filter wheel is moving...

PR #778, opened 2019-02-04
POCS in the Dock(er)

### LINKED ISSUE #734: Evaluate Docker as Developer and/or User option
Docker has recently moved to requiring users to register and log in to the docker store just to use the containers. This means if we go the Docker route that all PANOPTES users would first be required to register first, which doesn't exactly thrill me.

Note that Docker is under an Apache 2.0 license which technically means we could just download it once and host it ourselves and provide it to our users that way. Still, not entirely my favorite option.

Creating this issue just to be aware and discuss.

@jamessynge not sure what the current status of the Docker work is but would appreciate thoughts.

https://github.com/docker/docker.github.io/issues/6910

wtgee:
This presents a number of changes for building and running POCS and friends
in a docker environment. As of this PR, starting the environment (via docker-compose)
will launch four separate services (container images):

* messaging-hub
* weather
* paws
* pocs

As of now, it will also map the following directories to the host:

* $PANDIR/json_store
* $PANDIR/logs
* $PANDIR/images
* $PANDIR/POCS

It also appropriately sets up the internal network that allows the services to communicate
with each other over zeromq and exposes port 8080 on the host machine for reaching
PAWS.

The environment can be started with a helper script:

```bash
$POCS/scripts/run_pocs_docker.sh [help|jupyterlab]
```

As of now there is one additonal option that can be passed (`jupyterlab`) which
will start a jupyterlab environment within the running pocs service, exposing port
8888 on the host for access.

The POCS repo is installed in development mode (i.e. `pip install -e .`) so that changes
to the `$POCS` dir on the host will be reflected in the running service.

To do still:
* Environment sensors a la the weather container.
* Convenience script to make it easier for non-technical users to install docker, authenticate,
pull images, etc.
* Lots more I'm sure.

Note: this runs all of the docker images in `privileged` mode, meaning they have access to the host.

Also:
* Dockerfile to create a `pocs-base` image that contains the POCS repository
installed on an ubuntu 18.04 base image. This includes an install of all the
dependencies, including astrometry.net. Some of the apt-packages have been
cleaned up or changed as part of this PR.
* docker-compose.yaml files for easily starting all of the containers and helpers.
* Bash script for starting environments.
* Helper scripts for starting services. The POCS script (`$POCS/scripts/docker/run_pocs.sh`) still needs some work.
* Small README updates (more likely to come).
* `$PANUSER` is root as everything is run inside the images.
* The default python3 in ubuntu is used rather than dealing with anaconda. This helps keeps the image size smaller.

Closes #734 

> Note: The images are currently built and stored in our Google Container Registry rather than on Dockerhub.  The base is [gcr.io/panoptes-survey](http://gcr.io/panoptes-survey).

There is a coming PR for https://github.com/panoptes/PAWS that this PR relies on.


README.md
```diff
@@ -125,8 +125,7 @@ are some helper scripts to make this easier (from [here](https://cloud.google.co
 
 ```
 gcloud components install docker-credential-gcr
-docker-credential-gcr configure-docker
-docker-credential-gcr gcr-login
+gcloud auth configure-docker
 ```
```
programatt: one thing to note here is that you can also use an IAM service account and login with docker directly if there's no other need for having gcloud installed on the devices. https://cloud.google.com/container-registry/docs/advanced-authentication basically you create a service account and grant it read only permissions to the GCR storage bucket for the project and then they can login with docker cli directly using the service account json key like this `docker login -u _json_key -p "$(cat keyfile.json)" https://us.gcr.io` (assuming the key is in ./keyfile.json)

wtgee: Thanks @programatt, sorry for going so long. I've spent some time deep in Docker land in the last few weeks and think it will work well, thanks for all the tips.

We do use `gcloud` and `gsutil` pretty extensively and rely on a GCP network connection, so we do need them to authenticate at some point. We are using service account keys in a number of places and they work well.

I think I know have it so the images are public so anyone should be able to pull them, which is what I'm going for.

programatt: @wtgee awesome! Glad to hear it seems to be going well for this project.

wtgee:
Closing in favor of #859

PR #779, opened 2019-02-06
Add support for ZWO ASI cameras

AnthonyHorton:
Basically finished, marking as WIP because there are some unrelated changes to pull out into separate PRs.


pocs/camera/libfli.py
```diff
@@ -52,23 +52,18 @@ def __init__(self, library_path=False, *args, **kwargs):
         will be used to try to locate it.
 
         Args:
-            library_path (string, optional): shared library path,
-                e.g. '/usr/local/lib/libfli.so'
+            library_path (str, optional): path to the library, e.g. '/usr/local/lib/libfli.so'
```
wtgee:
```suggestion
            library_path (str, optional): path to the library, e.g. '/usr/local/lib/libfli.so'.
```

pocs/camera/sbigudrv.py
```diff
@@ -41,28 +42,22 @@ def __init__(self, library_path=False, retries=1, *args, **kwargs):
         will be used to locate it.
 
         Args:
-            library_path (string, optional): shared library path,
-                e.g. '/usr/local/lib/libsbigudrv.so'
+            library_path (str, optional): path to the library e.g. '/usr/local/lib/libsbigudrv.so'
```
wtgee:
```suggestion
            library_path (str, optional): path to the library e.g. '/usr/local/lib/libsbigudrv.so'.
```

pocs/camera/camera.py
```diff
@@ -197,10 +197,13 @@ def take_observation(self, observation, headers=None, filename=None, *args, **kw
         exp_time, file_path, image_id, metadata = self._setup_observation(observation,
                                                                           headers,
                                                                           filename,
-                                                                          *args,
                                                                           **kwargs)
 
-        exposure_event = self.take_exposure(seconds=exp_time, filename=file_path, *args, **kwargs)
+        try:
```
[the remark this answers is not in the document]
AnthonyHorton: Much neater syntax, too. As an aside, not entirely sure _why_ I had to start popping the `exp_time` entry from `kwargs`. I started getting errors from the simulator  `take_observation` test (which uses the `exp_time` keyword argument to override the `Observation` `exp_time` attribute) after my refactoring of `take_exposure` because `take_exposure` was being passed multiple values for `seconds`.  The `pop` fixes the problem, but feel that if I understood the root cause better there'd be a more elegant solution.

pocs/camera/camera.py
```diff
@@ -219,8 +222,51 @@ def take_observation(self, observation, headers=None, filename=None, *args, **kw
...
-    def take_exposure(self, *args, **kwargs):
-        raise NotImplementedError
+    def take_exposure(self,
+                      seconds=1.0 * u.second,
+                      filename=None,
+                      dark=False,
+                      blocking=False,
+                      *args,
+                      **kwargs):
+        """
```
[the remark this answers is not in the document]
AnthonyHorton: It is ugly, but OK.

pocs/camera/camera.py
```diff
@@ -219,8 +222,51 @@ def take_observation(self, observation, headers=None, filename=None, *args, **kw
...
+        Take an exposure for given number of seconds and saves to provided filename.
+
+        Args:
+            seconds (u.second, optional): Length of exposure
```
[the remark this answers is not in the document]

pocs/camera/camera.py
```diff
@@ -219,8 +222,51 @@ def take_observation(self, observation, headers=None, filename=None, *args, **kw
...
+            filename (str, optional): Image is saved to this filename
+            dark (bool, optional): Exposure is a dark frame (don't open shutter), default False
```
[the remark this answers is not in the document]
AnthonyHorton: Not entirely no effect, I think.  After my  tinkering with the `take_exposure` methods I think it will, now, effect the value of the `IMAGETYP' keyword in the FITS header. Before the keyword was just absent on cameras that don't support taking dark frames internally but it occurred to me that you might still be taking a dark by other means (e.g. put the lens cap on your DSLR lens, selected a filter wheel position with an opaque dummy filter in it) in which case you definitely would want that keyword to appear, with the value 'Dark Frame', if `dark==True`.

pocs/camera/fli.py
```diff
@@ -45,20 +46,22 @@ def __init__(self,
         # Create an instance of the FLI Driver interface
         self._FLIDriver = libfli.FLIDriver(library_path)
 
+        logger = get_root_logger()
```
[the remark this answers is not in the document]
AnthonyHorton: It's just because I don't want to call `super().__init__()` until later, once I have the correct values in `self._serial_number` and `self.port`, so that the log entries generated by `super().__init__()` don't use the placeholder default values for those.  Until I make that call `self.logger` doesn't exist so to enable logging in the port finding section I need to use `get_root_logger()`.

A bit of a faff just to avoid 'XXXXXX' appearing in the logs, but it works.

pocs/utils/__init__.py
```diff
@@ -387,3 +387,23 @@ def __exit__(self, *exc):
             # Suppress any exception caught while the context was running.
             return True
         return False
+
+
+def get_quantity_value(quantity, unit=None):
+    """ Given an Quantity attempts to convert to the given unit then returns the numerical value.
```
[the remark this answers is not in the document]
AnthonyHorton: Yes, that's right, I found myself writing basically this same block of code many times to check if something is a `Quantity`, convert units and extract the `value` when I needed to a simple numeric type (e.g. `ctypes` stuff, FITS headers).

I basically always use the optional unit argument, even if I'm fairly confident that the `Quantity` should already be in the right units. It's an extra check against errors.

pocs/camera/camera.py
```diff
@@ -219,8 +222,51 @@ def take_observation(self, observation, headers=None, filename=None, *args, **kw
 
         return observation_event
 
-    def take_exposure(self, *args, **kwargs):
-        raise NotImplementedError
+    def take_exposure(self,
```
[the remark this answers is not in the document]
AnthonyHorton: Yes, when writing the `take_exposure` method for the ZWO Camera I noticed just how much identical code there already was in SBIG, FLI and Simulator `take_exposure` methods so decided I shouldn't just create a 4th copy!

AnthonyHorton:
@wtgee Thanks for diving in to this review so quickly. Feeling guilty about all the not-really-related-to-the-purpose-of-the-PR refactoring in here. Guess you'd like me to split some of that out into separate bite-size PRs?

AnthonyHorton:
@wtgee Did some more work on the ZWO `Camera` class, bringing it up to roughly the same level of functionality as the SBIG and FLI cameras, with the additional of settable gain and image type.

It all seems to work to the extent that I can test it, though as I only have an uncooled camera on my desk right now I can't test the cooling related functionality yet.

I think this is now ready for when you get a chance to complete the review.

pocs/camera/camera.py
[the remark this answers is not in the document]
AnthonyHorton: Found and fixed the typo that started the problem, so have removed this `pop` again now.

wtgee:
> @wtgee Did some more work on the ZWO `Camera` class, bringing it up to roughly the same level of functionality as the SBIG and FLI cameras, with the addition of settable gain and image type.
> 
> It all seems to work to the extent that I can test it, though as I only have an uncooled camera on my desk right now I can't test the cooling related functionality yet.

@joshwalawender you mentioned you have a cooled ZWO available, any chance you could test this out within the next couple of days? If not, we could merge the PR under the assumption that it's at least moving us in the right direction and then fix anything related to cooling what's someone gets a chance to hook up a cooled version.

joshwalawender:
I do have a cooled ZWO available, but I'm now so far behind on POCS developments that I don't have a configured and working POCS install.  I'm happy to do some testing, but I'll need some help getting back up to speed.  I'd say you should merge and we can find any bugs discovered in cooling later.

wtgee:
> I do have a cooled ZWO available, but I'm now so far behind on POCS developments that I don't have a configured and working POCS install. I'm happy to do some testing, but I'll need some help getting back up to speed. I'd say you should merge and we can find any bugs discovered in cooling later.

Hopefully I'll have that Docker PR finished up by the end of the day and maybe you could test two birds with one stone. ;) We'll proceed with merging in this PR once review is done and I'll loop you in on trying to test out the cooled version.

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+                msg = "No ZWO ASI camera devices found"
+                logger.error(msg)
+                warn(msg)
+                return
+
+            # Get the IDs
+            for camera_index in range(n_cameras):
+                # Can get IDs without opening cameras by parsing the name string
+                self._info = Camera._ASIDriver.get_camera_property(camera_index)
+                model, _, id = self._info['name'].partition('(')
```
wtgee: Could just `split` instead of `partition` and then not worry about the `_` middle value? Or can there potentially be multiple `(` in the name?

wtgee: Also, `id` is a built-in function so this variable should change to something else just to be sure.

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            warn(msg)
+            return
+
+        super().__init__(name, *args, **kwargs)
+        self.connect()
+        assert self.is_connected, error.PanError("Could not connect to {}".format(self))
+
+        Camera._assigned_ids.append(self.uid)
+
+        self._setter_try('ccd_set_point', 'set_point', set_point)
```
wtgee: I see `_setter_try` used 3 times. In which of these would you be expecting the set to fail and if you know that in advance why bother calling it?

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+                Camera._ASIDriver.set_control_value(self._camera_ID, control_type, min_value)
+                raise error.PanError(msg)
+        else:
+            if not self._control_info[control_type]['is_auto_supported']:
+                msg = "{} cannot set {} to AUTO".format(self.model, control_name)
+                raise PanError(msg)
+
+        Camera._ASIDriver.set_control_value(self._camera_ID, control_type, value)
+
+    def _setter_try(self, attr, keyword, value):
```
wtgee: The `attr` and `keyword` seem a bit redundant. You use it for `ccd_set_point` and `set_point` but that seems like it's not totally necessary. Is it worth having the extra param just to have that difference in the log output?

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            # connect() will have set this based on camera info, but that doesn't know about filters
+            # upstream of the CCD. Can be set manually here, or handled by a filterwheel attribute.
+            self._filter_type = filter_type
+
+        self._setter_try('gain', 'gain', gain)
+
+        if image_type:
+            self._setter_try('image_type', 'image_type', image_type)
+        else:
+            # Take monochrome 12 bit raw images by default, if we can
```
wtgee: You say 12 bit and then look for `RAW16`, is that correct?

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            if 'RAW16' in self._info['supported_video_format']:
+                self.image_type = 'RAW16'
+
+        self.logger.info('{} initialised'.format(self))
+
+    def __del__(self):
+        """ Attempt some clean up """
+        with suppress(AttributeError):
+            id = self.uid
```
wtgee:
```suggestion
            uid = self.uid
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            Camera._assigned_ids.remove(id)
```
wtgee:
```suggestion
            Camera._assigned_ids.remove(uid)
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            self.logger.debug('Removed {} from assigned IDs list'.format(id))
```
wtgee:
```suggestion
            self.logger.debug('Removed {} from assigned IDs list'.format(uid))
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            Camera._ASIDriver.close_camera(camera_ID)
+            self.logger.debug("Closed ZWO camera {}".format(camera_ID))
+
+    # Properties
+
+    @AbstractCamera.uid.getter
+    def uid(self):
+        """Return unique identifier for camera.
+
+        Need to overide this because the base class only returns the 1st
```
wtgee:
```suggestion
        Need to override this because the base class only returns the 1st
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+        6 characters of the serial number, which is not a unique identifier
+        for most of the camera types
```
wtgee:
```suggestion
        for most of the camera types.
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+        """
+        return self._serial_number
+
+    @property
+    def image_type(self):
+        """ Current camera image type, one of 'RAW8', 'RAW16', 'Y8', 'RGB24' """
+        roi_format = Camera._ASIDriver.get_roi_format(self._camera_ID)
+        return roi_format['image_type']
```
wtgee: Is there a default `image_type`? Regardless, the setter can fail below if someone tries to set an unsupported type, which would then cause this property to throw a `KeyError` if accessed. Not sure if that is the intent.

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+        return header
+
+    def _refresh_info(self):
+        self._info = Camera._ASIDriver.get_camera_property(self._camera_index)
+
+    def _control_getter(self, control_type):
+        if control_type in self._control_info.keys():
+            return Camera._ASIDriver.get_control_value(self._camera_ID, control_type)
+        else:
+            raise NotImplementedError("'{}' has no '{}' parameter".format(self.model, control_type))
```
wtgee: Maybe an `AttributeError` here instead?

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def _control_setter(self, control_type, value):
+        if control_type not in self._control_info.keys():
+            raise NotImplementedError("{} has no '{}' parameter".format(self.model, control_type))
```
wtgee: And maybe `AttributeError` here as well?

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+
+        control_name = self._control_info[control_type]['name']
+        if not self._control_info[control_type]['is_writable']:
+            raise NotImplementError("{} cannot set {} parameter'".format(
```
wtgee:
```suggestion
            raise NotImplementedError("{} cannot set {} parameter'".format(
```

But maybe `AttributeError` again.

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+                except RuntimeError as err:
+                    raise error.PanError('Error getting image data from {}: {}'.format(self, err))
+                else:
+                    fits_utils.write_fits(image_data, header, filename, self.logger, exposure_event)
+            elif exposure_status == 'FAILED':
+                raise error.PanError("Exposure failed on {}".format(self))
+            elif exposure_status == 'IDLE':
+                raise error.PanError("Exposure missing on {}".format(self))
+            else:
+                raise error.PanError("Unepected exposure status on {}: '{}'".format(
```
wtgee:
```suggestion
                raise error.PanError("Unexpected exposure status on {}: '{}'".format(
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+        Connect to ZWO ASI camera.
+
+        Gets 'camera_ID' (needed for all driver commands), camera properties and details
+        of available camera commands/parameters.
+        """
+        self.logger.debug("Connecting to {}".format(self))
+        self._refresh_info()
+        self._camera_ID = self._info['camera_ID']
+        self.model, _, _ = self._info['name'].partition('(')
+        if self._info['is_colour_camera']:
```
wtgee: I don't see `is_colour_camera` anywhere else. And should probably be `is_color_camera` for consistency (sorry! :smile: )

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            self._filter_type = self._info['bayer_pattern']
+        else:
+            self._filter_type = 'M'
```
wtgee:
```suggestion
            self._filter_type = 'M'  # Monochrome
```

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+        header.set('YPIXSZ', get_quantity_value(self._info['pixel_size'], u.um), 'Microns')
+        header.set('EGAIN', get_quantity_value(self.egain, u.electron / u.adu), 'Electrons/ADU')
+        return header
+
+    def _refresh_info(self):
+        self._info = Camera._ASIDriver.get_camera_property(self._camera_index)
+
+    def _control_getter(self, control_type):
+        if control_type in self._control_info.keys():
+            return Camera._ASIDriver.get_control_value(self._camera_ID, control_type)
```
wtgee: Everywhere you use this above you then return the first element of a list. Does it make sense to just do that here?

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+            self.logger.error(msg)
+            raise ValueError(msg)
+        roi_format = self._ASIDriver.get_roi_format(self._camera_ID)
+        roi_format['image_type'] = new_image_type
+        Camera._ASIDriver.set_roi_format(self._camera_ID, **roi_format)
+
+    @property
+    def ccd_temp(self):
+        """ Current temperature of the camera's image sensor """
+        return self._control_getter('TEMPERATURE')[0]
```
wtgee: See note in `_control_getter` about the `[0]`.

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+    @property
+    def ccd_cooling_power(self):
+        """ Current power level of the camera's image sensor cooling system (as a percentage). """
+        return self._control_getter('COOLER_POWER_PERC')[0]
+
+    @property
+    def gain(self):
+        """ Current value of the camera's gain setting in internal units.
+
+        See `egain` for the corresponding eletrons / ADU value.
```
wtgee:
```suggestion
        See `egain` for the corresponding electrons / ADU value.
```

pocs/camera/libasi.py
```diff
@@ -0,0 +1,597 @@
...
+                         'is_trigger_camera': bool(camera_info.is_trigger_camera)}
+        return pythonic_info
+
+    def _parse_bins(self, supported_bins):
+        bins = tuple(int(bin) for bin in supported_bins if bin != 0)
+        return bins
+
+    def _parse_formats(self, supported_formats):
+        formats = []
+        for i in range(8):
```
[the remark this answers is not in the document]
AnthonyHorton: It could change, but only if the API of the SDK changes. The `supported_video_formats` field of the camera info C `struct` is an array of 8 `int`s. There's no real reason not to use `len(supported_formats)` here, though.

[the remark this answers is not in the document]
AnthonyHorton: Better yet, `for supported_format in supported_formats:`

pocs/camera/libasi.py
```diff
@@ -0,0 +1,597 @@
...
+
+    def _parse_return_value(self, value, control_type):
+        """ Helper function to apply appropiate type conversion and/or units to value """
+        try:
+            int_value = value.value  # If not done already extract Python int from ctypes.c_long
+        except AttributeError:
+            int_value = value  # If from a ctypes struct value will already be a Python int
+
+        # Apply control type specific units and/or data types
+        if control_type == 'EXPOSURE':
```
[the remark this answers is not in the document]
AnthonyHorton: Hmmm. Agree it's ugly, but not obvious to me how to do it more elegantly, because of the mix of casting, scale factors, unit conversions, enum lookup, etc.  A dictionary of lambda functions would do it, but wouldn't really simplify things. I could group all the bools together in one `elif control_type in (<list of control types>): nice_value = bool(int_value)` though, which might help a bit.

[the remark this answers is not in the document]
AnthonyHorton: The order follows the order of the control types in the `enum` that lists them all. Now you raise it I'm not sure there's any reason to stick with that though.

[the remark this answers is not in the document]
AnthonyHorton: Actually there are only 4 distinct cases, so with the help of a dictionary and a tuple I've reduces it to a 4 block if, elif, elif, else. Looks nicer now.

pocs/camera/zwo.py
```diff
@@ -0,0 +1,446 @@
...
+    def __init__(self,
+                 name='ZWO ASI Camera',
+                 set_point=None,
+                 filter_type=None,
+                 library_path=None,
+                 timeout=0.5 * u.second,
+                 gain=None,
+                 image_type=None,
+                 *args, **kwargs):
+        # ZWO ASI cameras don't have a 'port', they only have a non-deterministic integer
```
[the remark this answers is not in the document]
AnthonyHorton: Yes, the `AbstractCamera` docstring is outdated in the sense that it doesn't have any of the stuff that's only implemented in the derived classes.

pocs/camera/zwo.py
AnthonyHorton: The string ID part of the `name` field, enclosed in `()`, is set by the user, so for a sufficiently pathological user it may contain additional `(`...  I was careful to make zero assumptions about what is in between the first `(` and final `)`.

I could use `split` with `maxsplit=1` to only split on the first `(`  but the appeal of `partition` is that it offers a slightly neater way of handling the case where the camera hasn't had the string ID set, i.e. there is no `(` in the `name` field.

With `partition` I get an empty string back for the string ID and I can just use `if not string_ID:`. If I was using `split` I'd have to either check the length of the result or, better, catch `IndexError` when trying to access the second item of the list.

pocs/camera/zwo.py
AnthonyHorton: Hmm, it is of marginal usefulness. All it really does is convert `NotImplementedError` exceptions to warnings in the case where the user tried to set `set_point`, `gain` or `image_type` when initialising a camera the doesn't support setting those.  Our expert users probably _should_ have a fairly good idea of the cameras that they're configuring, so throwing a fatal `NotImplementedError` if they make a mistake in `pocs.yaml` is perhaps reasonable.

pocs/camera/zwo.py
AnthonyHorton: Yes, in a sense. The analogue to digital converter inside the camera produces (at most) 12 bit data, but that gets zero-padded to 16 bits on output to align with conventional 8 bit byte boundaries

It gets even more confusing for the low bit depth option. The raw data format for that is 8 bits per pixel but the camera specs talk about 10 bit mode, so the 2 least significant bits are getting thrown away?

pocs/camera/zwo.py
AnthonyHorton: Yes, `image_type` is an internal parameter of the camera. It has a default value and it will always have a value (internally it's just an integer at a certain memory location).  A failed attempt to set it to an unsupported value will just leave it set to whatever the current value is.

pocs/camera/zwo.py
AnthonyHorton: I used `NotImplementedError` here (even though this isn't really what that's for) because there is already a lot of code (e.g. large swathes of `pocs/tests/test_camera.py`) that catches `NotImplementedError`s from the `Camera` property getters and setters as a way of determining what capabilities the `Camera` has. Maybe I should use a new custom exception (e.g. NotSupportedError) that inherits from NotImplementedError (and PanError)?

wtgee: > Maybe I should use a new custom exception (e.g. `NotSupportedError`) that inherits from `NotImplementedError` (and `PanError`)?

If not too much work, I do like this idea.

pocs/camera/zwo.py
AnthonyHorton: *monocle drops out*

pocs/camera/zwo.py
AnthonyHorton: Maybe... `ASIDriver.get_control_value()` returns both the current value and a boolean indicating whether the value is currently being automatically adjusted by the camera.  For most of the camera parameters auto is not an option so this is usually `False`, and even when auto is on you may not care.  But there probably are some situations where you would want to know so I've left the second value in this private method in case I need it at some point.

pocs/camera/zwo.py
AnthonyHorton: OK, added a couple of simple custom exceptions. I should use `PanError` based exceptions more. There are a lot of places in my code where I use builtin exceptions and have to log the error separately.

pocs/camera/zwo.py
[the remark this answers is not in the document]
AnthonyHorton: Docstring updated.

PR #780, opened 2019-02-09
Fixing missing custom error

wtgee:
Missed during PR.


AnthonyHorton:
That's another reminder to move at much code as possible from the hardware specific classes (where the simulator can't test it) to the base class (where the simulator can).

PR #781, opened 2019-02-13
Implement is_exposing property for ZWO

AnthonyHorton:
Implements missing `is_exposing` property for the ZWO `Camera` class. This property was added to the simulator, SBIG and FLI `Camera` classes while the ZWO class was still under development so the ZWO class didn't get one at that time.

Needed for #775.

### DIFF
--- a/pocs/camera/zwo.py
+++ b/pocs/camera/zwo.py
@@ -221,6 +221,11 @@ class Camera(AbstractCamera):
         return self._info['e_per_adu']
 
     @property
+    def is_exposing(self):
+        """ True if an exposure is currently under way, otherwise False """
+        return Camera._ASIDriver.get_exposure_status(self._camera_ID) == "WORKING"
+
+    @property
     def properties(self):
         """ A collection of camera properties as read from the camera """
         return self._info


AnthonyHorton:
They're pretty quick. I am torn between caching status in a member variable and relying on the `_take_exposure()` and `_readout()` methods to keep it updated, or asking the hardware every time. The former is arguably more elegant and efficient, the latter is definitely more robust (e.g. will give the correct answer even when other code is doing stuff with the camera). For now I'll stick with asking the hardware.

I would like to revisit a bunch of issues like this as part of a major refactoring of the (non-Canon) camera code. There are three classes now, and they all do almost exactly the same things but in different ways. There's definitely scope for enforcing more consistency here, with benefits arising from moving more code to base classes where it can be more easily maintained and more thoroughly tested.  Would also be good preparation for possibly spinning the hardware support code out into a separate library, as discussed.

wtgee:
> They're pretty quick. I am torn between caching status in a member variable and relying on the `_take_exposure()` and `_readout()` methods to keep it updated, or asking the hardware every time. The former is arguably more elegant and efficient, the latter is definitely more robust (e.g. will give the correct answer even when other code is doing stuff with the camera). For now I'll stick with asking the hardware.

Agreed with asking the hardware assuming it's a pretty quick call. Otherwise it's just book-keeping. That said, we will most likely have to do the manual book-keeping with the DSLRs.

> 
> I would like to revisit a bunch of issues like this as part of a major refactoring of the (non-Canon) camera code. There are three classes now, and they all do almost exactly the same things but in different ways. There's definitely scope for enforcing more consistency here, with benefits arising from moving more code to base classes where it can be more easily maintained and more thoroughly tested. Would also be good preparation for possibly spinning the hardware support code out into a separate library, as discussed.

Yes, I think we will (unfortunately) have some big changes coming up. Mostly along the lines of the clean-up that will be required as part of the move to an asynchronous and message-based system.

Issue #777, opened 2019-02-01
Camera should not start exposure while filter wheel is moving


AnthonyHorton:
Inspired by #775 I realise that the camera `take_exposure()` method should, if the camera has a filterwheel, check that the filterwheel is not moving before actually starting the exposure.

AnthonyHorton:
Closed by #782

PR #782, opened 2019-02-13
Prevent camera exposures when filterwheel is moving

AnthonyHorton:
Addresses #777 by implementing an `is_moving` property for filterwheels and checking it in the `take_exposure()` method of the `Camera` classes.

### DIFF
--- a/pocs/filterwheel/sbig.py
+++ b/pocs/filterwheel/sbig.py
@@ -67,7 +67,7 @@ class FilterWheel(AbstractFilterWheel):
         """ Is the filterwheel currently moving """
         status = self._SBIGDriver.cfw_query(self._handle)
         if status['status'] == 'UNKNOWN':
-            self.logger.warning("{} returned 'UNKNOWN' status".self)
+            self.logger.warning("{} returned 'UNKNOWN' status".format(self))
         return bool(status['status'] == 'BUSY')
 
 ##################################################################################################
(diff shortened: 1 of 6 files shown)


pocs/filterwheel/sbig.py
```diff
@@ -62,6 +62,14 @@ def position(self):
             self.logger.warning("Filter wheel position unknown, returning NaN")
         return status['position']
 
+    @property
+    def is_moving(self):
+        """ Is the filterwheel currently moving """
+        status = self._SBIGDriver.cfw_query(self._handle)
+        if status['status'] == 'UNKNOWN':
+            self.logger.warning("{} returned 'UNKNOWN' status".self)
```
wtgee:
```suggestion
            self.logger.warning("{} returned 'UNKNOWN' status".format(self))
```

AnthonyHorton:
I do like the idea of an `is_ready` property, and am planning to implement it. I think I will do it as wrapper around other properties/methods though in order to allow more flexibility. Even for the simple case of filterwheels being 'ready' entails both not moving _and_ being at the correct position. For cameras it can be more than just not currently taking an exposure, e.g. image sensor temperature within tolerances of the target value, and cooler power not at maximum.

wtgee:
> I do like the idea of an `is_ready` property, and am planning to implement it. I think I will do it as wrapper around other properties/methods though in order to allow more flexibility. Even for the simple case of filterwheels being 'ready' entails both not moving _and_ being at the correct position. For cameras it can be more than just not currently taking an exposure, e.g. image sensor temperature within tolerances of the target value, and cooler power not at maximum.

:+1: yeah, seems like the `is_ready` would be good to support all around.

AnthonyHorton:
Yup. And it's a step towards the asynchronous, distributed, messaging based model, too.

PR #783, opened 2019-02-15
Simulator Camera creation no longer requires dummy camera config entries

### LINKED ISSUE #733: Simulator Camera Creation should not require dummy camera config entries
Currently the `create_cameras_from_config` will fail if the `cameras` entry is not populated in the config file. To get a simulator to work one must have at least:

```yaml
cameras:
    auto_detect: False
    devices:
    -
        model: simulator
```

Ideally the simulator should work when no config entries are present.

This manifested itself in AstroHuntsman/huntsman-pocs where the default config file does not contain a `cameras` entry but should be fixed for POCS in general.

AnthonyHorton:
Fixes #733 by creating a dummy camera config if camera simulation is enabled and the camera config section is either absent or empty.

### DIFF
--- a/pocs/tests/test_camera.py
+++ b/pocs/tests/test_camera.py
@@ -118,6 +118,13 @@ def test_create_cameras_from_empty_config():
     assert len(cameras) == 1
 
 
+def test_dont_create_cameras_from_empty_config():
+    # Can't pass a completely empty config otherwise default config will get loaded in its place.
+    really_empty_config = {'i_need_to_evaluate_to': True}
+    cameras = create_cameras_from_config(config=really_empty_config)
+    assert len(cameras) == 0
+
+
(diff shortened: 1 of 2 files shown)


pocs/tests/test_camera.py
```diff
@@ -111,6 +111,13 @@ def test_create_cameras_from_config_fail(config):
         create_cameras_from_config(config, simulator=simulator)
 
 
+def test_create_cameras_from_empty_config():
+    # create_cameras_from_cpnfig should work with no camera config, if cameras simulation is set
```
wtgee:
```suggestion
    # create_cameras_from_config should work with no camera config, if cameras simulation is set
```

AnthonyHorton:
Closes #733

Issue #631, opened 2018-09-25
exp_time vs exptime consistency


wtgee:
We currently use both `exp_time` and `exptime` as a variable name in different places. For clarity we should pick one and always use it.

Unless there is a good reason otherwise let's start using `exptime`.

AnthonyHorton:
`EXPTIME` is the most common form for FITS headers so using `exptime` in our code makes sense.  On the other hand `exp_time` is more in keeping with Python conventions for variable names. 3rd option, replacing them all with `exposure` instead! As its one word there's no question about whether it should have underscores or not.  `EXPOSURE` is also an accepted FITS header keyword, albeit in less common usage.

ImmortalRabbit:
> `EXPTIME` is the most common form for FITS headers so using `exptime` in our code makes sense. On the other hand `exp_time` is more in keeping with Python conventions for variable names. 3rd option, replacing them all with `exposure` instead! As its one word there's no question about whether it should have underscores or not. `EXPOSURE` is also an accepted FITS header keyword, albeit in less common usage.

Can I try to solve this issue? I am planning to replace EXPTIME to EXPOSUE and exptime, exp_time to exposure. Do you think that it will solve this issue?

PR #789, opened 2019-03-04
Issue #631 exp_time vs exptime consistency

ImmortalRabbit:
I have replaced variables exp_time, exptime by exposure and EXPTIME by EXPOSURE.

### DIFF
--- a/bin/pocs_shell
+++ b/bin/pocs_shell
@@ -481,7 +481,7 @@ Hardware names: {}   (or all for all hardware)'''.format(
 # Utility Methods
 ##########################################################################
 
-def polar_rotation(pocs, exp_time=30, base_dir=None, **kwargs):
+def polar_rotation(pocs, exposure=30, base_dir=None, **kwargs):
     assert base_dir is not None, print_warning("base_dir cannot be empty")
 
     mount = pocs.observatory.mount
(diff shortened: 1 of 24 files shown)


wtgee:
Thanks @ImmortalRabbit! Glad you could jump right in.

@AnthonyHorton you had made the recommendation of `EXPOSURE` (in #631), see any issues with us switching it at this point. I do feel like `EXPTIME` is more common as a FITS header.  

@joshwalawender might also have an opinion.

I'm happy to do the actual code review.

AnthonyHorton:
Yes, I'm inclined to stick with `EXPTIME` for the FITS headers. It's far more commonly used than `EXPOSURE` in that context, and I'm loathe to change FITS header keywords unless we really have to. Want to avoid breaking data reduction/analysis code that expects a particular set of keywords!

`exposure` would be fine everywhere else, but then that means we've still got a naming inconsistency, just a different one.

wtgee:
I would tend to agree, especially regarding the `EXPTIME` as a FITS keyword. The variable names are not as important but I think the `exptime` consistency would be nice.

@ImmortalRabbit sorry for the lack of clarity in #631. Can you change this to use `exptime` and `EXPTIME` accordingly?  Thanks!

Also, are you approaching this from GSoC?

ImmortalRabbit:
> I would tend to agree, especially regarding the `EXPTIME` as a FITS keyword. The variable names are not as important but I think the `exptime` consistency would be nice.
> 
> @ImmortalRabbit sorry for the lack of clarity in #631. Can you change this to use `exptime` and `EXPTIME` accordingly? Thanks!
> 
> Also, are you approaching this from GSoC?

Yes, I will change.
Yes, I am approaching this from GSoC. Sorry, for my messy pull request, just starting to learn how to contribute.

PR #790, opened 2019-03-05
variables changed: exp_time, exposure => exptime, EXPOSURE => EXPTIME.

ImmortalRabbit:
I have replaced all exp_time and exposure occurrence to exptime. This means that all commentaries have been changed because comments were referencing to exposure or exp_time as for variables to give brief explanation , and now they use exptime everywhere. Also, I have changed EXPOSURE to EXPTIME, but I have sow some EXPOSURE occurrence in libraries where I didn't change and leaved them as EXPOSURE. It's seems like EXPOSURE in libraries do not clash with EXPOSURE in project files. #631


wtgee:
Thanks again @ImmortalRabbit! We had some merges upstream right as you were doing this so there are a few conflicts. Could you merge upstream/develop and then we'll take a look.

Also, I'm assuming this replaces #789? It would have been ok to just make the changes in that branch and keep the same PR open rather than an entire new PR, but not a big deal. Thanks for the commits!

ImmortalRabbit:
> Thanks again @ImmortalRabbit! We had some merges upstream right as you were doing this so there are a few conflicts. Could you merge upstream/develop and then we'll take a look.
> 
> Also, I'm assuming this replaces #789? It would have been ok to just make the changes in that branch and keep the same PR open rather than an entire new PR, but not a big deal. Thanks for the commits! Could you close #789 if so?

Yes, I deleted #789.  I have a question about this issue: #631 In this issue it was discussed that exp_time and exptime is a same variables but POCS uses both of them which needs to be replaced. Then it was proposed to user exposure instead of exp_time and exptime. After a lit discussion of members of this project, it was decided to use exptime. Then question is : should I replace variable "exposure" to "exptime" too or not?

wtgee:
> Then question is : should I replace variable "exposure" to "exptime" too or not?

No, we don't want to do that and seeing this PR makes me also realized we probably confused the issue even more. An `exposure` and an `exptime` are different concepts and we need to keep them separate (that is, an "expousre" has an "exposure time"). It looks like you've done a global search and replace, which then kind of muddles the two together.

I think the easiest first step would be a search and replace on just `exp_time` to `exptime`. I think any other changes involving `exposure` inevitably change too many things we don't want to change, e.g. we **do not** want to change `take_exposure` to `take_exptime`.

Also I would not include the notebooks folder for now. They don't work well with search and replace changes like this and probably need to be cleaned up separately anyway.

ImmortalRabbit:
> 
Could you check PR please? I have done everything, what you have said.

wtgee:
> > 
> 
> Could you @wtgee @jamessynge check my PR please? I have done everything, what @wtgee have said.

Hi @ImmortalRabbit, sorry for not getting right back on this. It looks like this still has a number of spots where `exposure`->`exptime` so things like `take_exptime()` are still showing up. Can you maybe take another look?  Thanks!

Issue #526, opened 2018-06-29
Remove utc_offset


wtgee:
Determine if `utc_offset` is used anywhere and if not remove from config.

A basic code search shows it is not in use.

jamessynge:
FWIW, I see that utc_offset is an attribute that [astropy uses in its timezone info](http://docs.astropy.org/en/stable/search.html?q=utc_offset&check_keywords=yes&area=default). Not sure if that matters for us or not.

wtgee:
> FWIW, I see that utc_offset is an attribute that [astropy uses in its timezone info](http://docs.astropy.org/en/stable/search.html?q=utc_offset&check_keywords=yes&area=default). Not sure if that matters for us or not.

We don't appear to be using it at all.

ImmortalRabbit:
@wtgee I think this issue should be closed because my solution seems like correct and you have already merged it. Or are there any other things to do?

wtgee:
> @wtgee I think this issue should be closed because my solution seems correct and you have already merged it. Or are there any other things to do?

All good, thanks! FYI, in your PR you have "Fixes issue #526" but if you put "Fixes #526" it would have automatically closed the issue upon merge. See [here](https://help.github.com/en/articles/closing-issues-using-keywords) for more example.

ImmortalRabbit:
@wtgee thanks for suggestion, next time I will try to do not repeat this kind of mistake.

PR #791, opened 2019-03-09
Remove utc offset

ImmortalRabbit:
I have just deleted utc_offset with a simple code search. Fixes issue #526.

### DIFF
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -18,7 +18,6 @@ location:
     latitude: 19.54 # Degrees
     longitude: -155.58 # Degrees
     elevation: 3400.0 # Meters
-    utc_offset: -10.00 # Hours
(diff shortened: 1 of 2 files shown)


jamessynge:
I'm wondering if we have enough tests related to location. Many of the tests are based on the location of PAN001 (Hawaii), rather than at several locations, i.e. both east and west of prime meridian, and north and south of the equator.

wtgee:
> I'm wondering if we have enough tests related to location. Many of the tests are based on the location of PAN001 (Hawaii), rather than at several locations, i.e. both east and west of prime meridian, and north and south of the equator.

The `gmt_offset` that is still in the config is used to set up the iOptron mounts but that's about it. For everything else we assume UTC so it ideally shouldn't matter. But yes, probably best to have some separate tests. I'm going to leave that for a separate Issue/PR and go ahead and merge this.

PR #794, opened 2019-03-10
variable exp_time changed to exptime

ImmortalRabbit:
I have done same things that I did in #790 but with proper change of exp_time to exptime without changing exposure and without changing any variable in notebook directory.


wtgee:
@ImmortalRabbit this looks a lot cleaner except for some reason there are some changes to various cameras files that look like they were part of #787, which was merged right around that time. Can you make sure your `develop` was up to date and we'll see if those other changes go away.

PR #798, opened 2019-03-19
Issue 750 move common properties

ImmortalRabbit:


wtgee:
Hi @ImmortalRabbit, it looks like this wasn't updated from `develop` as it is reverting a number of the "exposure" related changes. Can you fix those up and then I'll take a look at this?

Not sure how you are starting your branches but we follow the pattern described here: https://github.com/panoptes/POCS/wiki/Feature-Development-Process. That should help keep the PR branches pretty clean and updated.

Issue #750, opened 2018-12-08
Scheduler - Move Common Properties


wtgee:
The dispatch scheduler currently computes some `common_properties` that are shared between all targets so there is not unnecessary re-computation. These should be moved up to the `BaseScheduler`.

ImmortalRabbit:
What if I will just move common_properties to BaseScheduler and will create a method that initializes values of common_properties depends on time that dispatch used. Will this method solve this issue? @wtgee

PR #799, opened 2019-03-19
common_properties moved to BaseScheduler

ImmortalRabbit:
@wtgee Sorry for recreating many times one Pull Request, here is my solution, I just have moved common_properties to BaseScheduler, that's how I basically understood this issue. #750

### DIFF
--- a/pocs/scheduler/dispatch.py
+++ b/pocs/scheduler/dispatch.py
@@ -1,7 +1,3 @@
-from astropy import units as u
-
-from astropy.coordinates import get_moon
-
 from pocs.utils import current_time
 from pocs.utils import listify
 from pocs.scheduler import BaseScheduler
--- a/pocs/scheduler/dispatch.py
+++ b/pocs/scheduler/dispatch.py
@@ -46,11 +46,7 @@ class Scheduler(BaseScheduler):
         valid_obs = {obs: 1.0 for obs in self.observations}
         best_obs = []
 
-        common_properties = {
-            'end_of_night': self.observer.tonight(time=time, horizon=-18 * u.degree)[-1],
-            'moon': get_moon(time, self.observer.location),
-            'observed_list': self.observed_list
-        }
+        self.set_common_properties(time)
 
         for constraint in listify(self.constraints):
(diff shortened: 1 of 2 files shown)


pocs/scheduler/dispatch.py
```diff
@@ -46,11 +46,7 @@ def get_observation(self, time=None, show_all=False, reread_fields_file=False):
         valid_obs = {obs: 1.0 for obs in self.observations}
         best_obs = []
 
-        common_properties = {
-            'end_of_night': self.observer.tonight(time=time, horizon=-18 * u.degree)[-1],
-            'moon': get_moon(time, self.observer.location),
```
wtgee: Can you remove the `get_moon` import from the top of this file as it is no longer used.

pocs/scheduler/scheduler.py
```diff
@@ -59,6 +60,7 @@ def __init__(self, observer, fields_list=None, fields_file=None,
             self.logger.debug("Reading initial set of fields")
             self.read_field_list()
 
+        self.common_properties = None
```
wtgee:
```suggestion
        # Items common to each observation that shouldn't be computed each time.
        self.common_properties = None
```

Add a comment describing what this is.

pocs/scheduler/scheduler.py
```diff
@@ -274,6 +276,14 @@ def read_field_list(self):
                 except Exception as e:
                     self.logger.warning("Error adding field: {}", e)
 
+    def set_common_properties(self, time):
+
+        self.common_properties = {
```
wtgee: Not really part of this PR, but since you are right here can you change this horizon to pull from the config rather than be hard-coded?

```python
horizon_limit = self.config['location'].get('observe_horizon', -18 * u.degree)      
self.common_properties = {
    'end_of_night': self.observer.tonight(time=time, horizon=horizon_limit)[-1],
    'moon': get_moon(time, self.observer.location),
    'observed_list': self.observed_list
}
```

ImmortalRabbit:
@wtgee Could you check my commit and did I get all your points?

PR #822, opened 2019-04-14
Update Camera Constructor - WIP

### LINKED ISSUE #820: create_cameras_from_config won't work with SDK cameras
**Description**

`pocs.cameras.create_cameras_from_config()` has a fixed set of parameters that it will extract from the config and pass as arguments to the `Camera` class constructor: `model`, `port`, `target_temperature`, `filter_type`, `focuser`, `filterwheel` and `readout_time`. Both `model` and `port` are required for each camera, while the rest will be replaced with default values if they are not in the config.

Since the refactoring of the camera class in #787 the SDK `Camera` classes (SBIG, FLI, ZWO) do not require a `port`, but do need the `serial_number` argument. Consequently `create_cameras_from_config` will fail if the configuration contains cameras of these types.

`create_cameras_from_config()` needs to be modified so that it only requires either `port` or `serial_number` to be in each camera's config, together with `model`.

For the sake of maintainability `create_cameras_from_config` should simply pass any additional parameters in the camera config beyond `name`, `model`, and `port`/`serial_number` unmodified to the `Camera` constructor.  This includes the `filter_type`, `focuser`, `filterwheel` and `readout_time` parameters currently being processed by `create_cameras_from_config()`. The `Camera` classes themselves already apply suitable default values to any missing arguments, doing the same in `create_cameras_from_config()` just adds redundant complexity.

**To Reproduce**
1. Create a POCS configuration YAML file with camera devices with model `sbig`, `fli` or `zwo`.
2. Call `create_cameras_from_config()`
3. The call will fail either because there is no `port` (`KeyError: No port specified and auto_detect=False`) or because `serial_number` does not get passed to the `Camera` constructor (`ValueError: Must specify serial_number for <cam_name>.`).

**Expected behavior**
All `Camera` objects should be created without error, regardless of camera type.

wtgee:
Creating the cameras from config will pass all the config entries to the camera object itself.

* Pass all the device config to the constructor.
* Cleanup of unused modules from camera classes.

## Related Issue
Closes #820.

## How Has This Been Tested?
**None yet**


wtgee:
@AnthonyHorton thoughts on how to write a test for this?

wtgee:
Note that this keeps the same behavior that currently exists, namely to through an error when a camera is not found (although that is switching from `error.PanError` to `error.CameraNotFoundError`, which makes more sense and is more specific). However, this is being one in a loop, which means if one of the camera configs is invalid the entire camera creation process will fail. Should probably just warn and continue and then throw an error at the end if there are no cameras.

wtgee:
As I suspected, this will require #813 to be finished for tests to properly pass. The PR as submitted now fixes this problem, it's just that the tests still won't pass because of #813. Marking as WIP for now. Should get to hardware testing on #813 within the next day and then be able to merge it.

pocs/camera/__init__.py
```diff
@@ -115,71 +115,73 @@ def kwargs_or_config(item, default=None):
...
                     continue
             else:
+                # Check for proper connection method.
+                model = device_config['model']
                 try:
-                    camera_port = device_config['port']
-                except KeyError:
+                    connection_method = model_requires[model]
+                    if connection_method not in device_config:
+                        raise error.CameraNotFound
```
AnthonyHorton: Is this the right error here? If I'm following correctly this would be the result of a bad configuration (device config missing the port/serial number it needs), rather than an expected camera not being present.

pocs/camera/__init__.py
```diff
@@ -115,71 +115,73 @@ def kwargs_or_config(item, default=None):
...
                     raise error.CameraNotFound(
-                        msg="No port specified and auto_detect=False")
+                        msg=f"No connection_method for {model} specified and auto_detect=False")
 
-            camera_focuser = device_config.get('focuser', None)
-            camera_filterwheel = device_config.get('filterwheel', None)
-            camera_readout = device_config.get('readout_time', 6.0)
+            device_config.setdefault('focuser', None)
+            device_config.setdefault('filterwheel', None)
+            device_config.setdefault('readout_time', 6.0)
```
AnthonyHorton: Not actually necessary to set default values for `focuser`, `filterwheel` or `readout_time` here as the `.__init__()` methods of the `Camera` classes will do appropriate things if these are not set. I'd argue against doing it here as it's redundant, and potentially confusing (user may expect the defaults from the `Camera` classes, but instead gets these values).

I'm glad to learn of the existence of the `.setdefault()` dictionary method, though!

pocs/camera/__init__.py
```diff
@@ -115,71 +115,73 @@ def kwargs_or_config(item, default=None):
...
+                                        'autofocus_seconds': 0.1,
+                                        'autofocus_size': 500}
+            device_config['filterwheel'] = {'model': 'simulator',
+                                            'filter_names': ['one', 'deux', 'drei', 'quattro'],
+                                            'move_time': 0.1 * u.second,
+                                            'timeout': 0.5 * u.second}
+            device_config['readout_time'] = 0.5
+
+        device_config.setdefault('set_point', None)
+        device_config.setdefault('filter_type', None)
```
AnthonyHorton: As above, not actually necessary to add default values for `set_point` and `filter_type` as the `Camera` classes already do sensible things when these arguments are absent.

pocs/tests/test_camera.py
```diff
@@ -117,6 +117,24 @@ def test_create_cameras_from_config_fail(config):
     with pytest.raises(error.PanError):
         create_cameras_from_config(config, simulator=simulator)
 
+    # SBIGs require a serial_number, not port
+    config['cameras']['devices'][0] = {
+        'port': '/dev/ttyFAKE',
+        'model': 'sbit'
```
AnthonyHorton: `sbig`?

pocs/tests/test_camera.py
```diff
@@ -117,6 +117,24 @@ def test_create_cameras_from_config_fail(config):
...
+    }
+
+    with pytest.raises(error.PanError):
+        create_cameras_from_config(config, simulator=simulator)
+
+    # Canon DSLRs require a port, not a serial_number
+    config['cameras']['devices'][0] = {
+        'serial_number': 'SC1234',
+        'model': 's'
```
AnthonyHorton: Tesla?

AnthonyHorton:
> @AnthonyHorton thoughts on how to write a test for this?

For once I think this can be tested reasonably well using the simulators, with a bit of tweaking.

There is a `pocs.camera.simulator.SDKCamera` class that should, I think, already insist on being passed a `serial_number`.  The basic `pocs.camera.simulator.Camera` class doesn't, currently, demand either but we could add a check for `port` to that, then fix all the tests we just broke...  Then it's a case of adding the two simulator classes to `model_requires` and testing `create_cameras_from_config` against a few good and bad simulated camera configs.

pocs/tests/test_camera.py
wtgee: I clearly had some copy/paste errors going on with this model value.

pocs/camera/__init__.py
wtgee: More to the point, I'm not entirely sure we should be raising an exception inside a for loop. This means, in the case of Huntsman, if one of the cameras has problems then all 10 would fail to initialise.

What we are actually testing here is the `create_cameras_from_config` method, which will eventually need to be overhauled for distributed cameras and for a change in the config server, rather than the actual camera creation process itself.

I'll probably end up changing this test a bit.

pocs/tests/test_camera.py
```diff
@@ -114,8 +116,30 @@ def test_create_cameras_from_config_fail(config):
...
 
-    with pytest.raises(error.PanError):
-        create_cameras_from_config(config, simulator=simulator)
+    cameras = create_cameras_from_config(config, simulator=simulator)
+    assert len(cameras) != 2
+
+    # SBIGs require a serial_number, not port
+    config['cameras']['devices'][0] = {
+        'port': '/dev/ttyFAKE',
+        'model': 'sbig'
```
AnthonyHorton: I'd like to suggest using the SDK Camera simulator here instead of the sbig model for the same reason as I suggested `simulator` over `canon_gphoto2` below. Not such a simple change in this case, though, because that simulator class doesn't follow the `pocs.camera.<model name>.Camera` convention that `create_cameras_from_config` relies on. It's currently `pocs.camera.simulator.SDKCamera`, but should probably be moved so that it can be created from a config.

wtgee: I think I have taken care of this properly now via the splitting of the files as we discussed IRL.

Issue #531, opened 2018-07-08
Create script for full setup of computer for a unit


jamessynge:
I've created various scripts and a document to assist in setting up the computer that runs a unit, but the process isn't yet simple (see #115). There are too many steps that the user must do. I propose that we create a script that is used immediately after Linux has been installed, wifi or ethernet configured and the owner's account (**not** `panoptes`) has been created. The script will require internet access to download dependencies (and itself), and will be executed like this:

```bash <(curl -s https://raw.githubusercontent.com/panoptes/POCS/develop/scripts/install/install-panoptes.sh)```

This script might trigger other scripts to be downloaded and executed.

I propose that the script do the following, specifically targeting Ubuntu 18.04 only for now:

- Present a set of environment variable values (e.g. PANUSER, PANDIR, etc.) and their default values
- Indicate if the user or directories already exists (esp. the user as the script will overwrite some of the files in the home dir).
- Allow the user to override the env vars.
- IFF user $PANUSER already exists, tell the user to delete the user before running this script again. The alternative is to complicate this script with lots of testing for already installed packages, etc.
- Create user $PANUSER. This will prompt for a password. Need to provide a suggestion that the user record this somewhere safe, for example a password vault such as LastPass.
- Add user $PANUSER to all of the important groups (e.g. adm, sudo, dialup, plugdev) so that it will have the ability to install software and access devices).
- Create the directories specified by the env vars, owned by user $PANUSER.
- Install anaconda for user $PANUSER, ideally without prompting, which I think the anaconda installer does by default.
- Install various software packages using conda (preferably) or pip (if not supported by conda) or directly (if supported by neither, such as mongodb or git).
- Replace the .profile and .bashrc of $PANUSER so that all shells run with the desired env vars, and with the panoptes anaconda environment.
- Clone the github repos and use setup tools to install them (e.g. python $POCS/setup.py install) and then to test them once they are all installed.

Assigning to @wtgee for review of the proposal. I'll proceed with doing some implementation.

wtgee:
This sounds good, thanks. A few quick comments:

* I think we can allow the user to override defaults but encourage them not to unless they have a good reason. With the separate user it should be a lot cleaner to just leave as is.
* anaconda needs a flag to auto-accept the EULA otherwise it hangs.
* Use miniconda over anaconda.
* Should probably make $USER a part of the $PANUSER group and make $PANDIR group r/w.
* Setup the cron jobs (at least the data download one)
[...]

PR #824, opened 2019-04-15
Add scripts/install/auto-install.sh

jamessynge:
Add script which can be used to install POCS and dependencies via a single command:

```
wget -O - https://raw.githubusercontent.com/panoptes/POCS/develop/scripts/install/auto-install.sh | bash
```

## Description

The script does the following:

1. Installs git if it isn't already installed
2. Creates `/var/panoptes` with the appropriate ownership.
3. Clones POCS if it isn't already cloned, else does a git pull to refresh it
4. Runs `scripts/install/install-dependencies.sh`

## Related Issue
#531

## How Has This Been Tested?
Tested by repeatedly setting up PAN005 for Doug and Christina.

## Types of changes

## Checklist:

- [x ] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.
- [ x] I have read the **CONTRIBUTING** document.


scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+#   which curl
+
+################################################################################
+# Env vars used for debugging of this script; these allow you to point to your
+# fork of POCS on github, so that you can download your fork instead of the
+# primary repo. There is no support for doing the same with PAWS.
+
+[[ -z "${POCS_GITHUB_USER}" ]] && export POCS_GITHUB_USER="panoptes"
+[[ -z "${POCS_GIT_URL}" ]] && export POCS_GIT_URL="https://github.com/${POCS_GITHUB_USER}/POCS.git"
+[[ -z "${POCS_BRANCH}" ]] && export POCS_BRANCH="develop"
```
wtgee: I think this is fine, but I have been thinking lately that we should go back to pushing release versions to `master` and having non-developer units always pull from `master`. Mostly the onus here would be on me to actually be good about making releases.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+of the way) and re-run this script.
+  "
+      exit 1
+    fi
+    echo "
+Cloning ${REPO_URL} into ${REPO_DIR}
+"
+    (set -x ; git clone "${REPO_URL}" "${REPO_DIR}")
+    cd "${REPO_DIR}"
+    git checkout "${BRANCH}"
```
wtgee: If the `POCS_GITHUB_USER` is not `panoptes` I think we should go ahead and set `upstream`.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+  else
+    echo "
+Pulling the latest software into the worktree at ${REPO_DIR}
```
wtgee:
```suggestion
Pulling the latest software into the worktree at ${REPO_DIR}.
```

You do realize you've made me anal about periods in comments now, right? :wink:

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+
+# COPIED from default-env-vars.sh
+[[ -z "${PANUSER}" ]] && export PANUSER="panoptes"      # Default user
+[[ -z "${PANDIR}" ]] && export PANDIR="/var/panoptes"   # Main Dir
+[[ -z "${PANLOG}" ]] && export PANLOG="${PANDIR}/logs"  # Log files
+[[ -z "${POCS}" ]] && export POCS="${PANDIR}/POCS"      # Main Observatory Control
+[[ -z "${PAWS}" ]] && export PAWS="${PANDIR}/PAWS"      # Web Interface
+
+if [[ "$(whoami)" != "${PANUSER}" ]] ; then
+  echo >2 "Please run this script as ${PANUSER}, not as $(whoami)"
```
wtgee:
```suggestion
  echo >2 "Please run this script as ${PANUSER}, not as $(whoami)".
```

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+  echo_bar
+  echo "
+git is not installed, so installing it...
+"
+  do_sudo apt-get install -y git
+fi
+
+echo_bar
+echo "
+Ensuring that ${PANDIR} exists
```
wtgee:
```suggestion
Ensuring that ${PANDIR} exists.
```

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+"
+if [ ! -d "${PANDIR}" ]
+then
+  do_sudo mkdir -p "${PANDIR}"
+fi
+
+echo_bar
+echo "
+Ensuring that ${PANDIR} is owned by user ${PANUSER}
```
wtgee:
```suggestion
Ensuring that ${PANDIR} is owned by user ${PANUSER}.
```

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+"
+do_sudo chown "${PANUSER}" "${PANDIR}"
```
wtgee:
```suggestion
do_sudo chown -R "${PANUSER}:${PANUSER}" "${PANDIR}"
```

I added the recursive option because I was thinking the subdirectories existed at this point but it looks like not but the option shouldn't hurt. Also not sure if wanted the group changed or not, I usually default to changing both.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,147 @@
...
+
+clone_or_update "${POCS}" "${POCS_GIT_URL}" "${POCS_BRANCH}"
+clone_or_update "${PAWS}" "https://github.com/panoptes/PAWS.git" "develop"
```
wtgee: Could obviously add the ability to clone a forked version of PAWS but I suppose not entirely necessary at this point. Can leave until we overhaul PAWS, which needs to happen soonish.

scripts/install/install-dependencies.sh
```diff
@@ -574,7 +574,10 @@ function maybe_install_conda() {
 function get_installed_astrometry_version() {
   local -r solve_field="${ASTROMETRY_DIR}/bin/solve-field"
   if [[ -x "${solve_field}" ]] ; then
-    "${solve_field}" --help|(grep -E '^Revision [0-9.]+,' || /bin/true)|cut -c10-|cut -d, -f1
+    ("${solve_field}" --help \
```
wtgee: I use `$SOLVE_FIELD` elsewhere to point to the `solve-field` executable. Here it is just checking a revision number, correct? (Also, I keep meaning to submit a PR to astrometry.net to add a simple `--version` flag. Very annoying it doesn't have one.)

wtgee:
@jamessynge what's the plan with this? Should I merge?

scripts/install/auto-install.sh
jamessynge: I think that might be a good idea to use master that way. I.e. on rare occasions, and after testing on multiple units, we declare a new version that non-developers should use.

scripts/install/auto-install.sh
jamessynge: Did that this morning.

scripts/install/auto-install.sh
jamessynge: Excellent! Done in my local repo, will upload shortly.

scripts/install/auto-install.sh
jamessynge: Done in my local repo, will upload shortly.

scripts/install/auto-install.sh
jamessynge: Done in my local repo, will upload shortly.

scripts/install/auto-install.sh
jamessynge: Done in my local repo, will upload shortly.

scripts/install/auto-install.sh
jamessynge: I tried that a while ago, even had it in this code earlier today, but then I remembered that it was really very slow if the full tree was already there. I can reconsider/retest if you like.

scripts/install/auto-install.sh
jamessynge: Exactly.

scripts/install/install-dependencies.sh
jamessynge: Yup, just checking version number.

jamessynge:
PTAL.

I've tested/adjusted this repeatedly, trying to streamline the process, and improve the experience for folks that aren't Linux experts. For example, it now invokes the set-panoptes-env.sh script twice, once at the start of .bashrc, and again at the end. This ensures that non-interactive shells are configured with the desired environment (e.g. for crontab), and also updates the prompt for interactive shells.

scripts/install/auto-install.sh
wtgee: Interesting. It seems like there shouldn't be _that_ many files in the tree. But okay not doing it. That means are assuming if it already exists then the user has already set their permissions accordingly, which I think is fine.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,245 @@
...
+Executing ${POCS}/scripts/install/install-dependencies.sh, which will
+install the tools needed to run POCS.
+"
+
+${POCS}/scripts/install/install-dependencies.sh
+
+# QUESTION: Should we add these lines here:
+# source $PANDIR/set-panoptes-env.sh
+# cd $POCS
+# python setup.py install
```
wtgee: This should be set to `develop` instead of `install`.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,245 @@
...
+echo_bar
+echo "
+Executing ${POCS}/scripts/install/install-dependencies.sh, which will
+install the tools needed to run POCS.
+"
+
+${POCS}/scripts/install/install-dependencies.sh
+
+# QUESTION: Should we add these lines here:
+# source $PANDIR/set-panoptes-env.sh
```
wtgee: What is the alternative? Set them manually?

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,245 @@
...
+    cd "${REPO_DIR}"
+    git checkout "${BRANCH}"
+  else
+    echo "
+Pulling the latest software into the worktree at ${REPO_DIR}.
+"
+    cd "${REPO_DIR}"
+    git fetch --all
+    git checkout "${BRANCH}"
+    git pull
```
wtgee: If they have already set up the repo they might have `origin` set to their repo and this pull might not necessarily be doing what we want. Or they might have files in staging, in which case this will fail.

I think if the dir is there and a repo we just do nothing other than maybe print a message reminding them to update. Thoughts?

wtgee:
@jamessynge I wonder if we also shouldn't copy `pocs.yaml` to `pocs_local.yaml` somewhere in this process.

scripts/install/auto-install.sh
jamessynge: I added it, and it work. I now recall that the slow case was inside of docker.

scripts/install/apt-packages-list.txt
```diff
@@ -78,3 +78,6 @@ coreutils
 # includes this file).
 git
 
+# Installing some POCS dependencies (esp. astroscrappy) require a C compiler,
```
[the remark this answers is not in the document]
jamessynge: Acknowledged.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,245 @@
...
+  echo
+  echo
+  do_sudo apt-get install -y git
+  echo
+fi
+
+################################################################################
+# Clone the POCS repo from github.
+
+clone_or_update "${POCS}" https://github.com/panoptes/POCS.git upstream develop
```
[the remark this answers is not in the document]
jamessynge: Yes. This is used as the origin name when cloning, so that `https://github.com/panoptes/POCS.git` will be the upstream repo, and `https://github.com/${POCS_GITHUB_USER}/POCS.git` will be the origin.

scripts/install/auto-install.sh
```diff
@@ -0,0 +1,245 @@
...
+
+if [[ -n "${POCS_GITHUB_USER}" && "${POCS_GITHUB_USER}" != "panoptes" ]]
+then
+  cd "${POCS}"
+  git remote add -f origin "https://github.com/${POCS_GITHUB_USER}/POCS.git" || true
+  git checkout "origin/${POCS_BRANCH:-develop}"
+fi
+
+clone_or_update \
+    "${PAWS}" "https://github.com/panoptes/PAWS.git" upstream develop
```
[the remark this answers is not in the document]
jamessynge: Yup, that is intended.

scripts/install/auto-install.sh
jamessynge: Yes. The install-dependencies.sh script spits out some lines for the user to run by hand, and I'm suggesting that I could do so automatically as part of validating the install.

scripts/install/auto-install.sh
jamessynge: Yeah, clearly if there are already files staged then we need to avoid messing with the repo.
Of course, if they have a repo, they don't need this auto-install script, and can run install-dependencies.sh themselves.

I'll adjust the script to stop if there is a POCS or PAWS repo.

jamessynge:
I've added some code for copying pocs.yaml to pocs_local.yaml, and reminding the user to customize it.

Issue #809, opened 2019-04-02
Update pySerial and deprecated commands


wtgee:
pyserial commands like reset_input_buffer( ) and  reset_output_buffer( ) may come in handy.
Maybe there is a need for a new function like clear_buffer ( ) under which these commands will be executed because peas_shell doesn't have one. 

Also i have noticed a line
https://github.com/panoptes/POCS/blob/4ef6e70e0484156c67b5c8178235b2a9987cca58/bin/peas_shell#L210
here flushInput( ) is supported for python versions 2.x the latest commands is reset_input_buffer( ) 
https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.reset_input_buffer

_Originally posted by @kmeagle1515 in https://github.com/panoptes/POCS/issues/347#issuecomment-478951014_

SMe12435:
```
def clean_buffer(self):
	    serial_connection = self.environment.serial_readers[relay_info['board']]['reader']

 	    try:
		    serial_connection.ser.reset_input_buffer()
		    serial_connection.write("{},9".format(relay_info['pin']))  #not sure what this line of code does.
	    except Exception as e:
		    print("Unable to reset Input Buffer: ")
		    print(e)

 	    try:
		    serial_connection.ser.reset_output_buffer()
		    serial_connection.write("{},9".format(relay_info['pin']))  #not sure what this line of code does.
	    except Exception as e:
		    print("Unable to reset output buffer: ")
		    print(e)

```

@wtgee As @kmeagle1515 mentioned in #810, we can do something like this.

PR #826, opened 2019-04-17
small fixes 

kmeagle1515:
## Description

small changes related to input/output buffer

## Related Issue

#809 

## Types of changes

## Checklist:

### DIFF
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -165,10 +165,12 @@ class PanSensorShell(cmd.Cmd):
         self.weather = AAGCloudSensor(serial_address=port, store_result=True)
         self.do_enable_sensor('weather')
 
-    def clear_buffer(self):
-        ser=serial.Serial('/dev/ttyUSB0')
-        ser.reset_input_buffer()
-        ser.reset_output_buffer()
+    def do_clear_buffer(self, sensor , sensor_name):
+        if hasattr(self, sensor) and sensor in self.active_sensors:
+            port = self.config[sensor][sensor_name]['serial_port']
+            ser=serial.Serial(port)
+            ser.reset_input_buffer()
+            ser.reset_output_buffer()
 
 ##################################################################################################
 # Relay Methods
--- a/bin/peas_shell
+++ b/bin/peas_shell
@@ -4,6 +4,7 @@ import datetime
 import os
 import readline
 import sys
+import serial
 
 from astropy.utils import console
 from threading import Timer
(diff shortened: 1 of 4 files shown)


kmeagle1515:
@wtgee i have added some fixes to the buffer issue #809

bin/peas_shell
```diff
@@ -164,6 +165,11 @@ class PanSensorShell(cmd.Cmd):
         self.weather = AAGCloudSensor(serial_address=port, store_result=True)
         self.do_enable_sensor('weather')
 
+    def clear_buffer(self):
```
jamessynge: This appears to be unused.

kmeagle1515: Thanks! @jamessynge 
clear_buffer will help to clean both i/p o/p buffers of the assigned port, i am not sure where it can be helpful.
Obviously it will be required before reading / writing something in to the port but i am not sure where in peas_shell.
It will require a bit discussion.

wtgee: I would suggest making it a callable option, i.e name it `do_clear_buffer`.

It should then be made more generic, so it can be resetting either `environment` or `weather` via `self.active_sensors[sensor].ser.reset_input_buffer()`. There are other example in the file of handling the params passed by the user.

kmeagle1515:
@wtgee i have made those changes.

bin/peas_shell
```diff
@@ -164,6 +165,13 @@ class PanSensorShell(cmd.Cmd):
         self.weather = AAGCloudSensor(serial_address=port, store_result=True)
         self.do_enable_sensor('weather')
 
+    def do_clear_buffer(self, sensor , sensor_name):
+        if hasattr(self, sensor) and sensor in self.active_sensors:
+            port = self.config[sensor][sensor_name]['serial_port']
+            ser=serial.Serial(port)
```
wtgee: The `self.active_sensors` holds actual sensor objects, either a `ArduinoSerialMonitor` or `AAGCloudSensor`. Those object already have a `SerialData` connection, so we don't want to create a new serial device connection here but instead use the existing ones.

Also, we don't seem to have any issues with the AAGCloudSensor as it is handling its own `read` and `write` methods so clears them appropriately.

The more I think about it the more we might just want to try to manage the buffer better via our read/writes in `pocs.utils.rs232.SerialData` and not expose this command. Specifically, I think we just need to reset the input buffer every time we are trying to issue a write command. @jamessynge thoughts?

kmeagle1515: Actually i did not thought about the pre-existing serial connection. Regardless of the issue #809 yeah the fix that you recently mentioned won't need an extra `do_clear_buffer` method instead it will only require the `self.ser.reset_input_buffer()` , also the `reset_input_buffer()` and `reset_output_buffer()` methods in the same rs232.py file won't be needed.

wtgee:
This should have been closed by #951. Please re-open with update if still applicable. Thanks!

PR #827, opened 2019-04-23
Add is_ready and related properties, rename camera properties

AnthonyHorton:
This PR continues tidying up of properties of the `Camera` and related classes (`Focuser`, `Filterwheel`), with the goal of facilitating the switch to the distributed model.

## Description

The main change is the addition of an `is_ready` property to the `Camera`, `Focuser`, and `Filterwheel` classes in order to determine whether they are ready for the next exposure. In the case of `Focuser` and `Filterwheel` they are 'ready' if they are not moving, and for `Camera` it is ready if it is not currently taking an exposure, its subcomponents are also ready, and, for cooled cameras, the image sensor temperature is stable.

In the process I have also added `is_cooled_camera` and `is_temperature_stable` properties to `Camera`.

I have also renamed the image sensor temperature related properties and parameters to make them more intuitive and more accurate.

## How Has This Been Tested?

Standard unit test suite, using hardware simulators.

## Types of changes

## Checklist:

- [X] I have added tests to cover my changes.


wtgee:
@AnthonyHorton is this ready for review?

AnthonyHorton:
@wtgee I need to add some tests, but otherwise yes, I think so.

pocs/camera/camera.py
```diff
@@ -189,15 +208,59 @@ def filter_type(self):
...
 
+    @property
+    def is_cooled_camera(self):
+        """ True if camera has image sensor cooling capability """
+        return self._is_cooled_camera
+
+    @property
+    def is_temperature_stable(self):
+        """ True if image sensor temperature is stable, False if not.
+
```
wtgee:
```suggestion
See also:
    See `temperature_tolerance` for more information about the temperature stability.
```

pocs/camera/camera.py
```diff
@@ -189,15 +208,59 @@ def filter_type(self):
...
+        An uncooled camera, or cooled camera with cooling disabled, will always return False.
+        """
+        if self.is_cooled_camera and self.cooling_enabled:
+            at_target = abs(self.temperature - self.target_temperature) \
```
wtgee: Probably overkill for this PR but does it make sense to have some kind of moving average for the "current" temperature. As far as I can tell this is only checking for momentary stability (which is probably find in most cases).

pocs/camera/camera.py
```diff
@@ -634,47 +698,54 @@ def _process_fits(self, file_path, info):
...
                     module = load_module(module_name)
                 except AttributeError as err:
                     self.logger.critical("Couldn't import {} module {}!".format(
                         class_name, module_name))
                     raise err
-                else:
-                    subcomponent_kwargs = copy.copy(subcomponent)
-                    subcomponent_kwargs.update({'camera': self, 'config': self.config})
-                    setattr(self, sub_name, getattr(module, class_name)(**subcomponent_kwargs))
+                subcomponent_kwargs = copy.copy(subcomponent)
```
wtgee: Do you want a `deepcopy` here? If the `subcomponent` contains any lists or other types within it we would probably want to. It's a small enough object that it probably doesn't hurt.

pocs/focuser/birger.py
```diff
@@ -260,16 +266,22 @@ def move_to(self, position):
...
-            return int(r[:-2])
+        self._is_moving = True
+        try:
+            response = self._send_command('fa{:d}'.format(int(position)), response_length=1)
+            if response[0][:4] != 'DONE':
+                self.logger.error("{} got response '{}', expected 'DONENNNNN,N'!".format(
+                    self, response[0].rstrip()))
+            else:
+                r = response[0][4:].rstrip()
+                self.logger.debug("Moved to {} encoder units".format(r[:-2]))
```
wtgee:
```suggestion
                self.logger.debug("Moved to {} encoder position".format(r[:-2]))
```

pocs/focuser/birger.py
```diff
@@ -260,16 +266,22 @@ def move_to(self, position):
...
-                self, response[0].rstrip()))
-        else:
-            r = response[0][4:].rstrip()
-            self.logger.debug("Moved to {} encoder units".format(r[:-2]))
-            if r[-1] == '1':
-                self.logger.warning('{} reported hitting a focus stop'.format(self))
-            return int(r[:-2])
+        self._is_moving = True
+        try:
+            response = self._send_command('fa{:d}'.format(int(position)), response_length=1)
```
wtgee: Even though it's a little extra work, maybe split this up to make more readable:

```suggestion
            full_response = self._send_command('fa{:d}'.format(int(position)), response_length=1)[0]
            status = full_response[:4]
            final_position = full_response[4:-2]
            hit_stop = full_response[-1]
```
And then change the corresponding entries below.

pocs/focuser/simulator.py
```diff
@@ -15,9 +15,32 @@ def __init__(self,
...
+
+    @property
+    def max_position(self):
+        """
+        Returns position of far limit of focus travel, in encoder units
+        """
+        return self._max_position
+
+    @property
+    def is_moving(self):
```
wtgee: This class inherits from `AbstractFocuser` so this property isn't needed here?

pocs/focuser/simulator.py
```diff
@@ -31,39 +54,24 @@ def connect(self):
         self._serial_number = 'SF{:04d}'.format(random.randint(0, 9999))
         self._min_position = 0
         self._max_position = 22200
+        if self.position is None:
+            self._position = random.randint(0, 22200)
```
wtgee:
```suggestion
            self._position = random.randint(0, self._max_position)
```

Not that it will matter much here, but still.

AnthonyHorton: Fair point.

pocs/camera/camera.py
AnthonyHorton: The assumption being made is that if it's at the right temperature now and isn't having to use 100% cooling power to get there then the camera's temperature control system should be able to maintain the correct temperature, at least until the end of the next exposure. I think that will generally be true, unless the ambient temperature increases rapidly.

Actually measuring the stability with a series of temperature readings would be better than just making assumptions about it, but probably not a priority.

pocs/camera/camera.py
AnthonyHorton: Yes, I should be using `deepcopy`. `subcomponent` will generally contain lists, etc.

pocs/focuser/birger.py
AnthonyHorton: Opted for "Moved to encoder position {}"

pocs/focuser/simulator.py
AnthonyHorton: `AbstractFocuser` has `is_moving` as an `abstractmethod`. Arguably should just implement it there, though.

wtgee:
@AnthonyHorton this is passing except for some coverage. Was there anything else to be done here?

AnthonyHorton:
@wtgee Not really, just wanted to fix up the coverage.

AnthonyHorton:
I'd forgotten this still hadn't been merged! The only reason it hasn't is the test coverage, I think. Should get this done asap.

AnthonyHorton:
@wtgee OK, I've improved the camera tests so they exercise the new properties properly, and added some simple temperature control simulation to the SDK camera simulator so that some of the code specific to cooled cameras also gets a workout. Otherwise this is basically as it was when you approved it before, but that was a _long_ time ago so wanted to check that you are still OK with merging this.

Note, because of the change of names of several temperature control related properties and corresponding config file keywords this is a breaking change for Huntsman, but not for PANOPTES (he asserts, confidently).

pocs/camera/camera.py
```diff
@@ -56,13 +58,14 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         firmware.  This can be done using ASICAP, or `pocs.camera.libasi.ASIDriver.set_ID()`.
     """
 
+    sub_classes = {'Focuser', 'FilterWheel'}
+    sub_names = {sub_class.casefold() for sub_class in sub_classes}
```
wtgee: Do these need to be public? Also, ask the risk of verbosity, I think a full `subcomonent_classes` and `subcomponent_names` might be more clear, especially if public.

wtgee: Also, why not just pass these in during initialization for those cameras that need them?

pocs/camera/camera.py
```diff
@@ -634,47 +699,54 @@ def _process_fits(self, file_path, info):
...
+                    class_name, base_module_name))
+                raise err
+            base_class = getattr(base_module, "Abstract{}".format(class_name))
+
             if isinstance(subcomponent, base_class):
                 self.logger.debug("{} received: {}".format(class_name, subcomponent))
-                setattr(self, sub_name, subcomponent)
-                getattr(self, sub_name).camera = self
+                setattr(self, class_name.casefold(), subcomponent)
+                getattr(self, class_name.casefold()).camera = self
```
wtgee: This assumes the subcomponent has a `camera` attribute already, which is probably a safe assumption, but could use another `setattr` here.

pocs/focuser/focuser.py
```diff
@@ -1,26 +1,26 @@
...
-import matplotlib.colors as colours
-
+from abc import ABCMeta
+from abc import abstractmethod
 from threading import Event
 from threading import Thread
 
 from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
 from matplotlib.figure import Figure
+import matplotlib.colors as colours
```
wtgee: Hm, seems non-standard. ;)

pocs/camera/camera.py
AnthonyHorton: It was a while ago, but I think my thinking here was that I wanted `Camera`s to always have all of these subcomponent attributes, but for them to be `None` if the corresponding hardware isn't present. I think this was motivated by a preference for `if self.<subcomponent>`: over
```
try:
    Do something with self.<subcomponent>
except AttributeError:
    pass
```
for conditional code that depends on the presence of subcomponents.

pocs/camera/camera.py
```diff
@@ -634,47 +699,54 @@ def _process_fits(self, file_path, info):
...
-                `pocs.focuser.AbtractFocuser`, used to check whether subcomponent is an instance.
+            class_name (str): name of the subcomponent class, e.g. 'Focuser'. Lower cased version
+                will be used as the attribute name, and must also match the name of the
+                corresponding POCS submodule for this subcomponent, e.g. `pocs.focuser`.
         """
         if subcomponent:
+            base_module_name = "pocs.{}.{}".format(class_name.casefold(), class_name.casefold())
+            try:
+                base_module = load_module(base_module_name)
+            except AttributeError as err:
```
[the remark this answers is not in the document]
AnthonyHorton: I think I was getting `AttributeError` here during testing, yes.  `base_module_name` here is something like `pocs.<submodule_class_name>.<submodule_class_name>` and I think it ends up an `AttributeError` because the import of `pocs` succeeds? We test that again to confirm, though.

pocs/camera/camera.py
AnthonyHorton: Yeah, the subcomponent classes do always have a `camera` attribute (actually a property, I think), though it might be `None`.

pocs/camera/camera.py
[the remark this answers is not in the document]
AnthonyHorton: OK, tested this and you are right, `load_module` does indeed raise `error.NotFound`. Both `AttributeError` and `ImportError` get raised along the way (so appear in the TraceBack) but they get caught lower down the stack.

pocs/camera/camera.py
```diff
@@ -634,47 +699,54 @@ def _process_fits(self, file_path, info):
...
-            sub_name (str): name of the subcomponent, e.g. 'focuser'. Will be used as the attribute
-                name, and must also match the name corresponding POCS submodule for this
-                subcomponent, e.g. `pocs.focuser`
-            class_name (str): name of the subcomponent class, e.g. 'Focuser'
-            base_class (class): the base class for the subcomponent, e.g.
-                `pocs.focuser.AbtractFocuser`, used to check whether subcomponent is an instance.
+            class_name (str): name of the subcomponent class, e.g. 'Focuser'. Lower cased version
+                will be used as the attribute name, and must also match the name of the
+                corresponding POCS submodule for this subcomponent, e.g. `pocs.focuser`.
         """
```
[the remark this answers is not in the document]
AnthonyHorton: Yes. Yes it does.

pocs/camera/camera.py
AnthonyHorton: Made private, and renamed.

Issue #837, opened 2019-05-29
Creation of scheduler outside of Observatory class


ImmortalRabbit:
**Is your feature request related to a problem? Please describe.**
In order to add dependency injection, creation of scheduler should be done outside of Observatory class.
**Describe the solution you'd like**
I would like to add creation of scheduler outside of Observatory class. In order to do that I am willing to send scheduler into Observatory class as cameras were sent: 
`def __init__(self, cameras=None, scheduler=None, *args, **kwargs):`
or
`def __init__(self, cameras=None, scheduler, *args, **kwargs):`
I am not quiet sure, should I consider situation when scheduler will not be sent to Observatory class. According to my understanding scheduler is just a software and it's not like cameras that can be not attached. 
Then planning to replace A with B:
A:
```
        self.scheduler = None
        self._create_scheduler()
```
B:
```
       self.scheduler = None 
       if scheduler:
            self.logger.info('Creation of the scheduler in the observatory')
            self.scheduler = scheduler
```
**Describe alternatives you've considered**

**Additional context**

**Related Issues/PRs**
If there are any related Issues or PRs list them here:
##### Required by: 

@wtgee @jamessynge

wtgee:
> am not quiet sure, should I consider situation when scheduler will not be sent to Observatory class.

Yes, I think we want to be able to create an Observatory without necessarily specifying a scheduler. As such, you should use the keyword based (i.e. `scheduler=None`) approach in the constructor.

One thing we will need to do is identify the methods inside the `Observatory` that rely on a scheduler and give appropriate warnings (or handle in some other way) if a scheduler is not present.  For example, an observatory cannot `.observe()` without a scheduler so it makes sense to do some checks before allowing that method.

ImmortalRabbit:
> One thing we will need to do is identify the methods inside the `Observatory` that rely on a scheduler and give appropriate warnings (or handle in some other way) if a scheduler is not present. For example, an observatory cannot `.observe()` without a scheduler so it makes sense to do some checks before allowing that method.

```
        if not self.scheduler:
            return dict()
```

Or should I raise error than just empty dictionary? Or empty dictionary will cause an error?

jamessynge:
The simple approach to dealing with _scheduler being None is simply to raise an exception (e.g. pocs.utils.error.NoObservation) if the observe method is called.

I'd convert .scheduler to ._scheduler, and introduce an @property accessor called scheduler which can then raise the NoObservation error if called when _scheduler isn't set.

[...]

Issue #836, opened 2019-05-29
Remove creation of scheduler from Observatory class


ImmortalRabbit:
**Is your feature request related to a problem? Please describe.**
In order to add dependency injection, creation of scheduler from Observatory class has to be removed.

**Describe the solution you'd like**
I would like to delete creation of scheduler from Observatory class. In order to do that I am willing to delete method _create_scheduler from Observatory class.
Then planning to replace A with B:
A:
```
        self.scheduler = None
        self._create_scheduler()
```
B:
```
       self.scheduler = None 
       
```
**Describe alternatives you've considered**

**Additional context**

**Related Issues/PRs**
If there are any related Issues or PRs list them here:
##### Required by:
@wtgee @jamessynge

wtgee:
@ImmortalRabbit is this the same as #837?

PR #838, opened 2019-05-31
Added scheduler as arg to Observatory and error NoScheduler

ImmortalRabbit:
I added a new argument to Observatory class and  a new error NoScheduler. 

## Description

Currently, I just don't use _create_scheduler. Functions that rely on scheduler can throw an error NoScheduler("Scheduler not found.") if scheduler is None. I am not quite sure that I did it right. I literally wrote code in every function that checks function that rely on scheduler. Is that okay? Also, I have a question, according to my understanding I will have to create a function `create_scheduler_from_config(config = None, logger=None, observer=None)`. Code of this function will be similar to `_create_scheduler`. However, if I will decide to test it, what kind of observer should I send to it? In Observatory class observer changes multiple times and I am not quite sure what kind of observer it should receive. I am confused at this part.

## Related Issue

#837 #836 #838 

## Types of changes

## Checklist:

@wtgee @jamessynge

### DIFF
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -14,12 +14,8 @@ import pocs.dome
 from pocs.base import PanBase
 from pocs.camera import AbstractCamera
 from pocs.images import Image
-from pocs.scheduler.constraint import Altitude
-from pocs.scheduler.constraint import Duration
-from pocs.scheduler.constraint import MoonAvoidance
 from pocs.utils import current_time
 from pocs.utils import error
-from pocs.utils import horizon as horizon_utils
 from pocs.utils import load_module
 
 
@@ -766,47 +762,3 @@ class Observatory(PanBase):
         self.mount = module.Mount(location=self.earth_location)
 
         self.logger.debug('Mount created')
-
-    def _create_scheduler(self):
-        """ Sets up the scheduler that will be used by the observatory """
-
-        scheduler_config = self.config.get('scheduler', {})
-        scheduler_type = scheduler_config.get('type', 'dispatch')
-
-        # Read the targets from the file
-        fields_file = scheduler_config.get('fields_file', 'simple.yaml')
-        fields_path = os.path.join(self.config['directories'][
-                                   'targets'], fields_file)
-        self.logger.debug('Creating scheduler: {}'.format(fields_path))
-
-        if os.path.exists(fields_path):
-
-            try:
-                # Load the required module
-                module = load_module(
-                    'pocs.scheduler.{}'.format(scheduler_type))
-
-                obstruction_list = self.config['location'].get('obstructions', list())
-                default_horizon = self.config['location'].get('horizon', 30 * u.degree)
-
-                horizon_line = horizon_utils.Horizon(
-                    obstructions=obstruction_list,
-                    default_horizon=default_horizon.value
-                )
-
-                # Simple constraint for now
-                constraints = [
-                    Altitude(horizon=horizon_line),
-                    MoonAvoidance(),
-                    Duration(default_horizon)
-                ]
-
-                # Create the Scheduler instance
-                self.scheduler = module.Scheduler(
-                    self.observer, fields_file=fields_path, constraints=constraints)
-                self.logger.debug("Scheduler created")
-            except ImportError as e:
-                raise error.NotFound(msg=e)
-        else:
-            raise error.NotFound(
-                msg="Fields file does not exist: {}".format(fields_file))
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -14,8 +14,12 @@ import pocs.dome
 from pocs.base import PanBase
 from pocs.camera import AbstractCamera
 from pocs.images import Image
+from pocs.scheduler.constraint import Altitude
+from pocs.scheduler.constraint import Duration
+from pocs.scheduler.constraint import MoonAvoidance
 from pocs.utils import current_time
 from pocs.utils import error
+from pocs.utils import horizon as horizon_utils
 from pocs.utils import load_module
 
 
@@ -762,3 +766,47 @@ class Observatory(PanBase):
         self.mount = module.Mount(location=self.earth_location)
 
         self.logger.debug('Mount created')
+
+    def _create_scheduler(self):
+        """ Sets up the scheduler that will be used by the observatory """
+
+        scheduler_config = self.config.get('scheduler', {})
+        scheduler_type = scheduler_config.get('type', 'dispatch')
+
+        # Read the targets from the file
+        fields_file = scheduler_config.get('fields_file', 'simple.yaml')
+        fields_path = os.path.join(self.config['directories'][
+                                       'targets'], fields_file)
+        self.logger.debug('Creating scheduler: {}'.format(fields_path))
+
+        if os.path.exists(fields_path):
+
+            try:
+                # Load the required module
+                module = load_module(
+                    'pocs.scheduler.{}'.format(scheduler_type))
+
+                obstruction_list = self.config['location'].get('obstructions', list())
+                default_horizon = self.config['location'].get('horizon', 30 * u.degree)
+
+                horizon_line = horizon_utils.Horizon(
+                    obstructions=obstruction_list,
+                    default_horizon=default_horizon.value
+                )
+
+                # Simple constraint for now
+                constraints = [
+                    Altitude(horizon=horizon_line),
+                    MoonAvoidance(),
+                    Duration(default_horizon)
+                ]
+
+                # Create the Scheduler instance
+                self.scheduler = module.Scheduler(
+                    self.observer, fields_file=fields_path, constraints=constraints)
+                self.logger.debug("Scheduler created")
+            except ImportError as e:
+                raise error.NotFound(msg=e)
+        else:
+            raise error.NotFound(
+                msg="Fields file does not exist: {}".format(fields_file))
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -128,20 +128,23 @@ class Observatory(PanBase):
 
     @property
     def current_observation(self):
-        if self.scheduler is None:
-            raise error.NoScheduler("Scheduler nor found.")
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
         return self.scheduler.current_observation
 
     @current_observation.setter
     def current_observation(self, new_observation):
-        if self.scheduler is None:
-            raise error.NoScheduler("Scheduler nor found.")
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
         self.scheduler.current_observation = new_observation
 
     @property
     def has_dome(self):
         return self.dome is not None
 
+    @property
+    def can_observe(self):
+        return self.scheduler is not None
 ##########################################################################
 # Device Getters/Setters
 ##########################################################################
@@ -203,8 +206,8 @@ class Observatory(PanBase):
         """
         status = {}
         try:
-            if self.scheduler is None:
-                raise error.NoScheduler("Scheduler nor found.")
+            if not self.can_observe:
+                raise error.NoObservation("No valid observations found.")
 
             t = current_time()
             local_time = str(datetime.now()).split('.')[0]
@@ -246,7 +249,6 @@ class Observatory(PanBase):
 
         return status
 
-
     def get_observation(self, *args, **kwargs):
         """Gets the next observation from the scheduler
 
@@ -260,8 +262,8 @@ class Observatory(PanBase):
 
         self.logger.debug("Getting observation for observatory")
 
-        if self.scheduler is None:
-            raise error.NoScheduler("Scheduler nor found.")
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
 
         # If observation list is empty or a reread is requested
         reread_fields_file = (
@@ -311,8 +313,8 @@ class Observatory(PanBase):
             except KeyError:
                 keep_jpgs = True
 
-        if self.scheduler is None:
-            raise error.NoScheduler("Scheduler nor found.")
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
 
         process_script = 'upload_image_dir.py'
         process_script_path = os.path.join(os.environ['POCS'], 'scripts', process_script)
@@ -422,8 +424,8 @@ class Observatory(PanBase):
             "Analyzing recent image using pointing image: '{}'".format(pointing_image))
 
         try:
-            if self.scheduler is None:
-                raise error.NoScheduler("Scheduler nor found.")
+            if not self.can_observe:
+                raise error.NoObservation("No valid observations found.")
 
             # Get the image to compare
             image_id, image_path = self.current_observation.last_exposure
@@ -472,8 +474,8 @@ class Observatory(PanBase):
         """
         if self.current_offset_info is not None:
             self.logger.debug("Updating the tracking")
-        if self.scheduler is None:
-            raise error.NoScheduler("Scheduler nor found.")
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
             # Get the pier side of pointing image
             _, pointing_image = self.current_observation.pointing_image
             pointing_ha = pointing_image.header_ha
@@ -515,8 +517,8 @@ class Observatory(PanBase):
 
         self.logger.debug("Getting headers for : {}".format(observation))
 
-        if self.scheduler is None:
-            raise error.NoScheduler("Scheduler nor found.")
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
 
         t0 = current_time()
         moon = get_moon(t0, self.observer.location)
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -59,6 +59,9 @@ class Observatory(PanBase):
 
         self.logger.info('\tSetting up scheduler')
         self.scheduler = scheduler
+        if self.scheduler is None:
+            self.scheduler = pocs.scheduler.create_scheduler_from_config(self.config, logger=self.logger,
+                                                                         observer=self.observer)
 
         self.current_offset_info = None
 
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -158,7 +158,7 @@ class Observatory(PanBase):
         if can_observe and self.scheduler is None:
             self.logger.info(f'Scheduler not present, cannot observe.')
             can_observe = False
-        if can_observe and self.cameras is None:
+        if can_observe and not self.has_cameras:
             self.logger.info(f'Cameras not present, cannot observe.')
             can_observe = False
         if can_observe and self.mount is None:
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -526,6 +526,10 @@ class Observatory(PanBase):
         Returns:
             dict: The standard headers
         """
+
+        if not self.can_observe:
+            raise error.NoObservation("Cannot be observed.")
+
         if observation is None:
             observation = self.current_observation
 
@@ -535,9 +539,6 @@ class Observatory(PanBase):
         field = observation.field
 
         self.logger.debug("Getting headers for : {}".format(observation))
-
-        if not self.can_observe:
-            raise error.NoObservation("Cannot be observed.")
 
         t0 = current_time()
         moon = get_moon(t0, self.observer.location)
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -775,8 +775,7 @@ class Observatory(PanBase):
 
         # Read the targets from the file
         fields_file = scheduler_config.get('fields_file', 'simple.yaml')
-        fields_path = os.path.join(self.config['directories'][
-                                       'targets'], fields_file)
+        fields_path = os.path.join(self.config['directories']['targets'], fields_file)
         self.logger.debug('Creating scheduler: {}'.format(fields_path))
 
         if os.path.exists(fields_path):
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -14,12 +14,8 @@ import pocs.dome
 from pocs.base import PanBase
 from pocs.camera import AbstractCamera
 from pocs.images import Image
-from pocs.scheduler.constraint import Altitude
-from pocs.scheduler.constraint import Duration
-from pocs.scheduler.constraint import MoonAvoidance
 from pocs.utils import current_time
 from pocs.utils import error
-from pocs.utils import horizon as horizon_utils
 from pocs.utils import load_module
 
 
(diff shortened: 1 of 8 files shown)


wtgee:
> I literally wrote code in every function that checks function that rely on scheduler. Is that okay?

I think we will inevitably need to do this so it's the right direction. However, in order to properly observe we will need more items than just a scheduler, i.e. we need a camera, a mount, etc.

What we should probably do is create another method (it could potentially also be a property), `can_observe` that can check all of these items (as well as the weather safety) and then replace what you have done here with a call to `can_observe`.  For now you can just put the check for scheduler in `can_observe` but eventually it will have a few other items in there as well.

ImmortalRabbit:
@wtgee @jamessynge What do say about my new commits?

Is that correct pull request. I just didn't know how to do divide scheduler process into separate pull request because this codes are connected, so I decided to do a few commits with one pull request for each of them (scheduler, mount, dome and writing tests). 

Currently, I just implemented create_scheduler_from_config and in order to check it I create scheduler in Observatory class. 

Next step will be moving calling create_scheduler_from_config outside of Observatory class.

pocs/observatory.py
```diff
@@ -59,8 +58,10 @@ def __init__(self, cameras=None, *args, **kwargs):
         self.dome = pocs.dome.create_dome_from_config(self.config, logger=self.logger)
 
         self.logger.info('\tSetting up scheduler')
-        self.scheduler = None
-        self._create_scheduler()
+        self.scheduler = scheduler
+        if self.scheduler is None:
+            self.scheduler = pocs.scheduler.create_scheduler_from_config(self.config, logger=self.logger,
```
wtgee: We don't want to automatically create a scheduler if one is not passed during construction. Probably nothing needs to happen here, except maybe a log message stating that there is no scheduler.

pocs/observatory.py
```diff
@@ -130,17 +131,23 @@ def primary_camera(self, cam):
 
     @property
     def current_observation(self):
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
```
wtgee: `can_observe` could evaluate to False for a number of reasons (no scheduler, no cameras, no mount). That is not the same thing as saying there is not a valid observation. This message implies that all of the hardware is correct and that the scheduler is working, it's just that there are no particular targets available at this time, which is not what we want to say.

The specific reason that `can_observe` is False should be handled inside `can_observe` itself.  

Also, one could technically have a `scheduler`, and a `current_observation`, and still `can_observe` is False, e.g. when there are no cameras present.  We shouldn't be prevented from getting the `current_observation` in that situation.

pocs/observatory.py
```diff
@@ -130,17 +131,23 @@ def primary_camera(self, cam):
...
         return self.scheduler.current_observation
 
     @current_observation.setter
     def current_observation(self, new_observation):
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
```
wtgee: Same as above.

pocs/observatory.py
```diff
@@ -130,17 +131,23 @@ def primary_camera(self, cam):
...
         self.scheduler.current_observation = new_observation
 
     @property
     def has_dome(self):
         return self.dome is not None
 
-
+    @property
+    def can_observe(self):
+        return self.scheduler is not None
```
wtgee: This is going to expand to include things like the camera, mount, etc. Let's set up the framework to test each of those and give appropriate message if not found.

This will end up something like:
```python
def can_observe(self):
   can_observe = True

    if can_observe and self.scheduler is None:
        self.logger.info(f'Scheduler not present, cannot observe')
        can_observe = False

    # Same thing for mount
    # Same thing for cameras
    # Etc.

    return can_observe
```

Also, docstrings. ;)

pocs/observatory.py
```diff
@@ -202,6 +209,9 @@ def status(self):
         """
         status = {}
         try:
+            if not self.can_observe:
```
wtgee: We should not be raising an Exception in the status call. We always want to be able to return a status. That status should indicate if observing is possible or not. Basically, `status['can_observe'] = self.can_observe` is all that's needed here.

pocs/observatory.py
```diff
@@ -255,6 +265,9 @@ def get_observation(self, *args, **kwargs):
 
         self.logger.debug("Getting observation for observatory")
 
+        if not self.can_observe:
+            raise error.NoObservation("No valid observations found.")
```
wtgee: Same comments as above w.r.t. the observations vs the scheduler.

pocs/observatory.py
```diff
@@ -411,6 +427,9 @@ def analyze_recent(self):
             "Analyzing recent image using pointing image: '{}'".format(pointing_image))
 
         try:
+            if not self.can_observe:
```
wtgee: Not sure this is necessary here in the analyze face.

pocs/observatory.py
```diff
@@ -500,6 +520,9 @@ def get_standard_headers(self, observation=None):
 
         self.logger.debug("Getting headers for : {}".format(observation))
 
+        if not self.can_observe:
```
wtgee: Not necessary here.

pocs/scheduler/__init__.py
```diff
@@ -1 +1,69 @@
...
+
+    if not logger:
+        logger = logger_module.get_root_logger()
+
+    if 'scheduler' not in config:
+        logger.info("No scheduler in config")
+        return None
+
+    if not observer:
+        logger.info("No valid observation found.")
```
wtgee:
```suggestion
        logger.info("No valid Observer found.")
```

An `Observer` is different from an `Observation`. The `Observer` is the telescope itself, which has knowledge of the timezone and location where it is sitting.

pocs/observatory.py
ImmortalRabbit: According to your commentary, I understood that I should not raise error when can_observe is False. can_observe should display reason of it's falseness ( like no scheduler, no cameras, no mount) and should return True or False. Then, when I will start using it, I am thinking that I can do like that
```if not can_observe:
         return None #what else should I return instead of None or how else I have to stop function if it's not possible to observe```

File: pocs/scheduler/__init__.py [hunk 10]
```diff
@@ -1 +1,69 @@
...
+    """ Sets up the scheduler that will be used by the observatory """
+
+    if not logger:
+        logger = logger_module.get_root_logger()
+
+    if 'scheduler' not in config:
+        logger.info("No scheduler in config")
+        return None
+
+    if not observer:
```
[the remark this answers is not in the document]
wtgee: @jamessynge any thoughts on this?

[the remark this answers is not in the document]
ImmortalRabbit: > @jamessynge any thoughts on this?

@wtgee I made separate issue #840.

pocs/observatory.py
```diff
@@ -303,6 +329,9 @@ def cleanup_observations(self, upload_images=None, make_timelapse=None, keep_jpg
             except KeyError:
                 keep_jpgs = True
 
+        if not self.can_observe:
```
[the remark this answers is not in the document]
ImmortalRabbit: > For now let's just check for `self.scheduler` here (again with no error) and not worry about the rest for this PR. However we will want to make an Issue to handle the cleanup situation when a scheduler is removed.

@wtgee I made separate issue #840.

ImmortalRabbit:
> HI @ImmortalRabbit, there are still some of the `NoObservation`s popping up in a few places they probably shouldn't. Also it looks like the tests aren't passing although I haven't yet looked at the reasons why.

It's probably because _create_scheduler is not called right know and create_scheduler_from_config is not used yet. For example, test_bad_scheduler sends config with bad scheduler type which had to lead to error during creation of scheduler but it didn't because Observatory class have to receive scheduler and if it doesn't receives it sets scheduler as None. Therefore, Observatory class can be initialized with scheduler which is None.

pocs/tests/test_scheduler.py
```diff
@@ -0,0 +1,32 @@
...
+from pocs.scheduler import create_scheduler_from_config
+from pocs.site_location_details import setup_site_location_details_from_config
+from pocs.utils import error
+
+
+def test_bad_scheduler_type(config):
+    conf = config.copy()
+    conf['scheduler']['type'] = 'foobar'
+    details = setup_site_location_details_from_config(config)
+    with pytest.raises(error.NotFound):
```
wtgee: This is actually a good example of how code coverage can help expose bugs even if the tests don't seem to. Looking at your PR I'm realizing we've had a problem in our scheduler testing, which is not something that you introduced but instead was already there.

If you look at the code coverage for the scheduler you'll see that there is one line that is not getting hit, namely the error that should be raised for an invalid scheduler type ([line 64](https://codecov.io/gh/panoptes/POCS/pull/838/diff?src=pr&el=tree#D5-64)). What should be happening is that the code should be unable to find the `foobar` scheduler (because it doesn't exist) and so it should raise a ` NotFound` exception on line 64. However the red is telling us that piece of code is never getting run, which means we are never actually using it.

What is going on is that this test, which should be checking for the bad scheduler type, is actually failing because of the missing fields path check on [line 37](https://github.com/panoptes/POCS/blob/70416431f23476a3b535513ff59ffc1057e6407c/pocs/scheduler/__init__.py#L37). (The fields file is checked in the next test below).  So, basically we are not really ever checking for a bad scheduler type.

And looking even deeper, the `load_module` function itself is [catching the ImportError](https://github.com/panoptes/POCS/blob/develop/pocs/utils/__init__.py#L316), and is raising the `error.NotFound`, which means the [ImportError on line 63](https://github.com/panoptes/POCS/blob/70416431f23476a3b535513ff59ffc1057e6407c/pocs/scheduler/__init__.py#L63) could never possibly get raised because `load_module` is re-raising the `ImportError` but as an `error.NotFound` instead.

Again, this was not something you have done but was something that was already there. However let's use this time to go ahead and clean it up.

Two things need to be done here:

* [Line 63](https://github.com/panoptes/POCS/blob/70416431f23476a3b535513ff59ffc1057e6407c/pocs/scheduler/__init__.py#L63) should change to be `except error.NotFound as e`. This is a little bit pointless as it just re-raises the same exception on the next line and so you could alternatively just remove the entire `try/except` block (lines 39 and 63). However let's keep it and just re-raise the `error.NotFound` for now. In the future we would maybe want to do something smarter there instead (such as create a `error.ImportError` of our own), so it makes sense to leave it in place.
* This test needs to have a valid `conf['scheduler']['fields_file']`. That is, it should be set to a real path name such that the existence check (line [37](https://github.com/panoptes/POCS/blob/70416431f23476a3b535513ff59ffc1057e6407c/pocs/scheduler/__init__.py#L37)) evaluates to True.

Once this is done the tests should pass, as they do now, but that [red marker on line 64](https://codecov.io/gh/panoptes/POCS/pull/838/diff?src=pr&el=tree#D5-64) should disappear and our coverage should go up (a tiny bit).

That's a pretty subtle bug. Let me know if anything doesn't make sense.  Thanks!

wtgee: But yes, this is a good example of why you should look at the coverage reports and try to figure out _why_ the red chunks are red.

ImmortalRabbit: @wtgee Should I try to increase coverage in Observatory class?

pocs/observatory.py
```diff
@@ -490,6 +526,10 @@ def get_standard_headers(self, observation=None):
         Returns:
             dict: The standard headers
         """
+
+        if not self.can_observe:
```
wtgee: This check isn't necessary as should be handled by the `assert` on line 536 below.

pocs/tests/test_scheduler.py
wtgee: Left a message in slack a few days ago, but no, not unless it is directly related to this PR. We want to try to keep PRs focused.

pocs/tests/test_observatory.py
```diff
@@ -27,7 +29,10 @@ def simulator():
 @pytest.fixture
 def observatory(config, simulator, images_dir):
     """Return a valid Observatory instance with a specific config."""
+    details = setup_site_location_details_from_config(config)
```
wtgee: See comments above but `details` is also a meaningless name as it could refer to the details of anything. This is `location_info` or `site_info`. Since it contains keys called `location` then it should probably be `site_details`.

pocs/site_location_details/__init__.py
```diff
@@ -0,0 +1,75 @@
+from astroplan import Observer
```
wtgee: This is an awkard name for a directory and also kind of just an awkward file. All of this is really just setting up configuration details and shouldn't really be set up as a module.

Could move this file into `pocs/utils/location.py` for now although we might want to change that in the long run.

pocs/tests/test_observatory.py
```diff
@@ -1,16 +1,18 @@
...
-from pocs import hardware
 import pocs.version
+from pocs import hardware
+from pocs.camera import create_cameras_from_config
 from pocs.observatory import Observatory
+from pocs.scheduler import create_scheduler_from_config
 from pocs.scheduler.dispatch import Scheduler
 from pocs.scheduler.observation import Observation
-from pocs.camera import create_cameras_from_config
+from pocs.site_location_details import setup_site_location_details_from_config
```
wtgee: See comments in file above.

pocs/tests/test_observatory.py
```diff
@@ -37,6 +42,19 @@ def observatory(config, simulator, images_dir):
     return obs
 
 
+def test_camera_already_exists(observatory, config):
```
wtgee: I see you added this already. We try to keep PRs focused on a single thing and adding tests for the camera doesn't really have anything to do with the scheduler. Ok to leave for now but in the future I'd probably ask you to more to a separate PR.

pocs/tests/test_observatory.py
```diff
@@ -37,6 +42,19 @@ def observatory(config, simulator, images_dir):
...
+    cameras = create_cameras_from_config(config)
+    for cam_name, cam in cameras.items():
+        observatory.add_camera(cam_name, cam)
+
+
+def test_remove_cameras(observatory, config):
+    obs = observatory
```
wtgee: This line is unnecessary.

pocs/tests/test_observatory.py
```diff
@@ -69,20 +87,44 @@ def test_bad_mount_driver(config):
...
 
 
-def test_bad_scheduler(config):
+def test_can_observe_no_scheduler(config):
     conf = config.copy()
-    simulator = ['all']
-    conf['scheduler']['type'] = 'foobar'
-    with pytest.raises(error.NotFound):
-        Observatory(simulator=simulator, config=conf, ignore_local_config=True)
+    cameras = create_cameras_from_config(conf)
```
wtgee: No reason to add the cameras to this test. The Observatory should be able to be created without cameras and you are trying to test the scheduler here, so you want to make a minimal example required to test the scheduler.

wtgee: Also, you've removed the test for a badly named scheduler, which is different from no scheduler.

pocs/tests/test_observatory.py
```diff
@@ -69,20 +87,44 @@ def test_bad_mount_driver(config):
...
+
+    obs = Observatory(cameras=cameras,
+                      scheduler=None,
+                      config=conf
+                      )
 
+    assert obs.can_observe is False
```
wtgee: You've probably added the cameras because otherwise this is `False`. The situation for no scheduler is different from the situation with a badly named scheduler so that needs to be resolved first.

pocs/tests/test_observatory.py
wtgee: I guess you have that down in the test_scehduler, so that is okay.

wtgee: You can capture the logs that are output and make sure it is outputting the correct log message rather than just checking for the False below.

pocs/tests/test_observatory.py
```diff
@@ -69,20 +87,44 @@ def test_bad_mount_driver(config):
...
 
-def test_bad_scheduler_fields_file(config):
+
+def test_can_observe_no_cameras(config):
```
wtgee: This test and the next test should follow above. That is, we should be testing for the correct log message with minimal examples added. You will want to do that for the cameras since you created the `can_observe` method and the cameras are already part of that. However see not on mount below.

pocs/tests/test_observatory.py
```diff
@@ -69,20 +87,44 @@ def test_bad_mount_driver(config):
...
+    obs = Observatory(cameras=None,
+                      scheduler=scheduler,
+                      config=conf
+                      )
+
+    assert obs.can_observe is False
+
+
+def test_can_observe_no_mount(config):
+    conf = config.copy()
```
wtgee: We haven't yet moved the mount creation out of the Observatory so probably didn't need to modify the `can_observe` at this point, which means we don't need to create this test right now.

pocs/tests/test_observatory.py
```diff
@@ -122,6 +164,12 @@ def test_primary_camera(observatory):
     assert observatory.primary_camera is not None
 
 
+def test_primary_camera_no_primary_camera(observatory):
+    obs = observatory
```
wtgee: This line is not necessary.

Did this test increase coverage of some line? I would be surprised if we weren't already testing that.

pocs/tests/test_observatory.py
```diff
@@ -209,6 +257,13 @@ def test_standard_headers(observatory):
     assert headers['longitude'] == test_headers['longitude']
 
 
+def test_get_standard_headers_no_scheduler(observatory):
```
wtgee: We *do* want to be able to get headers without a scheduler, as mentioned in the `get_standard_headers` above.

pocs/tests/test_observatory.py
```diff
@@ -227,6 +282,23 @@ def test_get_observation(observatory):
     assert observatory.current_observation == observation
 
 
+def test_get_observation_no_scheduler(observatory):
+    obs = observatory
```
wtgee: This line is unnecessary.

pocs/tests/test_observatory.py
```diff
@@ -227,6 +282,23 @@ def test_get_observation(observatory):
...
+    obs.scheduler = None
+    assert obs.get_observation() is None
+
+
+def test_cleanup_observations_no_scheduler(observatory):
+    obs = observatory
```
wtgee: Same.

pocs/tests/test_observatory.py
```diff
@@ -227,6 +282,23 @@ def test_get_observation(observatory):
...
+    obs.scheduler = None
+    assert obs.cleanup_observations() is None
+
+
+def test_cleanup_observations_keep_jpgs(observatory):
+    obs = observatory
```
wtgee: Same.

pocs/tests/test_observatory.py
```diff
@@ -227,6 +282,23 @@ def test_get_observation(observatory):
...
+    assert obs.get_observation() is None
+
+
+def test_cleanup_observations_no_scheduler(observatory):
+    obs = observatory
+    obs.scheduler = None
+    assert obs.cleanup_observations() is None
+
+
+def test_cleanup_observations_keep_jpgs(observatory):
```
wtgee: This is unrelated to this PR and I'm not sure is testing what we want. Let's just remove for now.

pocs/tests/test_pocs.py
```diff
@@ -1,18 +1,20 @@
...
 
+import pytest
 from astropy import units as u
 
 from pocs import hardware
+from pocs.camera import create_cameras_from_config
 from pocs.core import POCS
 from pocs.observatory import Observatory
+from pocs.scheduler import create_scheduler_from_config
+from pocs.site_location_details import setup_site_location_details_from_config
```
wtgee: See note about this file.

pocs/tests/test_observatory.py
ImmortalRabbit: > This line is not necessary.
> 
> Did this test increase coverage of some line? I would be surprised if we weren't already testing that.

I wrote this test because it wasn't tested. It was marked in red color.

pocs/tests/test_observatory.py
```diff
@@ -1,17 +1,19 @@
...
 import pocs.version
+from pocs import hardware
+from pocs.camera import create_cameras_from_config
 from pocs.observatory import Observatory
+from pocs.scheduler import create_scheduler_from_config
 from pocs.scheduler.dispatch import Scheduler
 from pocs.scheduler.observation import Observation
-from pocs.camera import create_cameras_from_config
 from pocs.utils import error
+from pocs.utils.location import setup_site_location_details_from_config
```
wtgee:
```suggestion
from pocs.utils.location import create_location_from_config
```

Let's just call this `create_location_from_config` even though it returns more than just a strict `location`.

pocs/tests/test_observatory.py
```diff
@@ -27,7 +29,10 @@ def simulator():
 @pytest.fixture
 def observatory(config, simulator, images_dir):
     """Return a valid Observatory instance with a specific config."""
+    site_details = setup_site_location_details_from_config(config)
```
wtgee:
```suggestion
    site_details = create_location_from_config(config)
```

pocs/tests/test_observatory.py
```diff
@@ -69,20 +86,17 @@ def test_bad_mount_driver(config):
...
-    conf = config.copy()
-    simulator = ['all']
-    conf['scheduler']['fields_file'] = 'foobar'
-    with pytest.raises(error.NotFound):
-        Observatory(simulator=simulator, config=conf, ignore_local_config=True)
+    obs = Observatory(config=conf)
+    assert obs.can_observe is False
+    assert caplog.records[-1].levelname == "INFO" and caplog.records[
+        -1].message == "Scheduler not present, cannot observe."
+    site_details = setup_site_location_details_from_config(conf)
```
wtgee:
```suggestion
    site_details = create_location_from_config(conf)
```

pocs/tests/test_observatory.py
```diff
@@ -147,8 +166,7 @@ def test_default_config(observatory):
     """ Creates a default Observatory and tests some of the basic parameters """
 
     assert observatory.location is not None
-    assert observatory.location.get('elevation') - \
-        observatory.config['location']['elevation'] < 1. * u.meter
+    assert observatory.location.get('elevation') - observatory.config['location']['elevation'] < 1. * u.meter
```
wtgee:
```suggestion
    assert observatory.location.get('elevation') == pytest.approx(observatory.config['location']['elevation'], rel=1 * u.meter)
```

I'm not actually sure if this will work with the units in place. Might need to just add `.value` to the end of each.

But the `pytest.approx` is the better way to test things like this.

pocs/tests/test_pocs.py
```diff
@@ -1,18 +1,20 @@
...
 from pocs.core import POCS
 from pocs.observatory import Observatory
+from pocs.scheduler import create_scheduler_from_config
 from pocs.utils import CountdownTimer
-from pocs.utils.messaging import PanMessaging
-from pocs.utils import error
 from pocs.utils import current_time
-from pocs.camera import create_cameras_from_config
+from pocs.utils import error
+from pocs.utils.location import setup_site_location_details_from_config
```
wtgee:
```suggestion
from pocs.utils.location import create_location_from_config
```

pocs/tests/test_pocs.py
```diff
@@ -42,10 +44,17 @@ def cameras(config):
 
 
 @pytest.fixture(scope='function')
-def observatory(config, db_type, cameras):
+def scheduler(config):
+    site_details = setup_site_location_details_from_config(config)
```
wtgee:
```suggestion
    site_details = create_location_from_config(config)
```

pocs/tests/test_scheduler.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+
+from pocs.scheduler import create_scheduler_from_config
+from pocs.utils import error
+from pocs.utils.location import setup_site_location_details_from_config
```
wtgee:
```suggestion
from pocs.utils.location import create_location_from_config
```

pocs/tests/test_scheduler.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_bad_scheduler_type(config):
+    conf = config.copy()
+    conf['scheduler']['type'] = 'foobar'
+    site_details = setup_site_location_details_from_config(config)
```
wtgee:
```suggestion
    site_details = create_location_from_config(config)
```

pocs/tests/test_scheduler.py
```diff
@@ -0,0 +1,32 @@
...
+    with pytest.raises(error.NotFound):
+        create_scheduler_from_config(conf, observer=site_details['observer'])
+
+
+def test_bad_scheduler_fields_file(config):
+    conf = config.copy()
+    conf['scheduler']['fields_file'] = 'foobar'
+    site_details = setup_site_location_details_from_config(config)
```
wtgee:
```suggestion
    site_details = create_location_from_config(config)
```

pocs/tests/test_scheduler.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_no_observer(config):
+    assert create_scheduler_from_config(config, observer=None) is None
+
+
+def test_no_scheduler_in_config(config):
+    conf = config.copy()
+    del conf['scheduler']
+    site_details = setup_site_location_details_from_config(conf)
```
wtgee:
```suggestion
    site_details = create_location_from_config(conf)
```

pocs/utils/location.py
```diff
@@ -0,0 +1,75 @@
+from astroplan import Observer
+from astropy import units as u
+from astropy.coordinates import EarthLocation
+
+from pocs.utils import error
+from pocs.utils.logger import get_root_logger
+
+
+def setup_site_location_details_from_config(config, logger=None):
```
wtgee:
```suggestion
def create_location_from_config(config, logger=None):
```

pocs/utils/location.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+     Sets up the site and location details
```
wtgee:
```suggestion
     Sets up the site and location details.
```

Don't forget the periods! ;) I miss this all the time too.

Issue #843, opened 2019-06-11
Moving creation of dome outside of Observatory class.


ImmortalRabbit:
*<Enter description of change or enhancement here>*
I will remove creation of mount from Observatory and will send it as argument.

**Describe the solution you'd like**
* Removed dome creation from Observatory class
* Implemented function that creates dome outside of Observatory class
* Implemented function that can remove dome
* Implemented unit tests to check modification that were done

**Describe alternatives you've considered**

**Additional context**

**Related Issues/PRs**
If there are any related Issues or PRs list them here:

PR #849, opened 2019-06-21
Dome injection into Observatory

ImmortalRabbit:
Moved creation of dome outside of Observatory class.
## Description

`create_dome_from_config` was already there, so I just moved it outside of Observatory class and added a new argument for Observatory class.
## Related Issue

#843 
## Types of changes

## Checklist:

### DIFF
--- a/pocs/tests/test_observatory.py
+++ b/pocs/tests/test_observatory.py
@@ -8,6 +8,7 @@ from astropy.time import Time
 import pocs.version
 from pocs import hardware
 from pocs.camera import create_cameras_from_config
+from pocs.dome import create_dome_from_config
 from pocs.observatory import Observatory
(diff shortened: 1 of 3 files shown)


pocs/tests/test_observatory.py
```diff
@@ -388,9 +391,11 @@ def test_no_dome(observatory):
     assert observatory.close_dome()
 
 
-def test_operate_dome(config_with_simulated_dome):
+def test_operate_dome(config_with_simulated_dome, config):
```
wtgee: This is a little weird to me but appears to be working. I think we could just have one config. What happens if you remove the `config_with_simulated_dome` and just use the `config`? @jamessynge might have some insight with the simulated dome.

Issue #850, opened 2019-06-22
Separate simulator creation from hardware creation


wtgee:
The various `create_<hardware>_from_config` scripts are currently also capable of creating a "simulator" version of the hardware. To me that seems non-logical as creating from config should necessarily be creating real hardware.  

**Describe the solution you'd like**
I would propose we remove all simulator creation from these convenience functions and handle simulator creation separately.  This could look like a `create_from_config` (one name for each type of hardware, specified as a class method on the base class) and `create_simulator`.

**Additional context**
I think this would help with testing, both for software testing as well as real hardware testing.

**Related Issues/PRs**
#848 #636 #849 #478 #29 #842 #843 probably more.

Thoughts? @jamessynge @ImmortalRabbit  @AnthonyHorton

jamessynge:
I'm concerned about this proposal because I don't want the production code to have branches such as "if should use simulator". I'm wondering if we can't have a third method: create_<hardware>_driver, which is used by the production code, and is responsible for deciding if we're using a hardware driver simulator or a real driver (possibly with simulated hardware).

I discussed the above with @ImmortalRabbit this morning, and noted that this would mean we'd have a repeated pattern for each type of hardware. This lead me to wonder about some common approach to this, such as an application of the Registry pattern. Thoughts?

PR #851, opened 2019-06-23
Move POCS to panoptes-utils

wtgee:
**Breaking Changes**

This is a large PR that overhauls almost every single file in POCS.  The primary change is to use the utils as provided by the `panoptes-utils` module. As part of this there is a switch to using the config_server for all configuration parameters, which implies some changes in how we are doing things

## Description
This PR mostly changes all the `pocs.utils` to `panoptes.utils`.  It also includes necessary changes to the testing suite. In order to make this work a config_server is created as a `function` level fixture.  The `PanBase` class then provides a new `get_config` method to each class that will make an http (localhost only be default) to the running config_server. For testing, the config_server is run on port 6565 (6563 is the default for production use) and is set to not use local files and not save any parameters.

The `config_port` is a `session` level fixture that most be used with all calls to `get_config` outside of objects that inherit from `PanBase`. Objects that do inherit from `PanBase` should pass `config_port` as one of the initialization parameters.

There is also some intermittent and misc cleanup. Ideally this should all be done in smaller PRs but it's hard to structure and make work versus just doing a large-scale breaking change.

## Related Issue
Depends on: https://github.com/panoptes/panoptes-utils/pull/44
#850 
#778 

## How Has This Been Tested?
Most of the PR is changing the underlying tests to make it all work the same. Hardware tests will need to be done separately.  There are still some outstanding issues with the simulator setups that needs to be resolved (see #850)

## Types of changes
**Breaking Changes**

## Checklist:


wtgee:
Closing in favor of #859

Issue #842, opened 2019-06-11
Moving creation of mount outside of Observatory class.


ImmortalRabbit:
*<Enter description of change or enhancement here>*
I am going to remove mount creation from Observatory class and create function `create_mount_from_config` outside of Observatory class.

**Describe the solution you'd like**
* Removed mount creation from Observatory class
* Implemented function that creates mount outside of Observatory class
* Implemented function that can remove mount
* Implemented unit tests to check modification that were done

**Describe alternatives you've considered**

**Additional context**

**Related Issues/PRs**
If there are any related Issues or PRs list them here:

PR #852, opened 2019-06-24
create_mount_from_config was created.

ImmortalRabbit:
I have modified _create_mount into create_mount_from_config

## Description

create_mount_from_config was created similar to _create_mount. I deleted part that used simulator during creation of mount. It means that create_mount_from_config creates actual mount only. 

## Related Issue

#842 #855

## Types of changes

## Checklist:

### DIFF
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -27,7 +27,6 @@ def create_mount_from_config(config, mount_info=None, earth_location=None):
 
     model = mount_info.get('model')
 
-    model = mount_info.get('brand')
     driver = mount_info.get('driver')
 
     # See if we have a serial connection
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -22,6 +22,7 @@ def create_mount_from_config(config, mount_info=None, earth_location=None):
         earth_location = site_details['earth_location']
 
     if mount_info is None:
+        logger.debug(f'No mount info provided, using values from config.')
         mount_info = config.get('mount')
 
     model = mount_info.get('model')
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -1,1 +1,55 @@
+from glob import glob
+
 from pocs.mount.mount import AbstractMount  # pragma: no flakes
+from pocs.utils import error
+from pocs.utils import load_module
+from pocs.utils.location import create_location_from_config
+from pocs.utils.logger import get_root_logger
+
+
+def create_mount_from_config(config, mount_info=None, earth_location=None):
+    """ Sets up the mount that will be used by the observatory """
+
+    logger = get_root_logger()
+
+    if 'mount' not in config:
+        logger.info("No mount in config")
+        return None
+
+    if earth_location is None:
+        site_details = create_location_from_config(config)
+        earth_location = site_details['earth_location']
+
+    if mount_info is None:
+        mount_info = config.get('mount')
+
+    model = mount_info.get('model')
+
+    model = mount_info.get('brand')
+    driver = mount_info.get('driver')
+
+    # See if we have a serial connection
+    try:
+        port = mount_info['serial']['port']
+        if port is None or len(glob(port)) == 0:
+            msg = "Mount port({}) not available. ".format(port) + "Use simulator = mount for simulator. Exiting."
+            raise error.MountNotFound(msg=msg)
+    except KeyError:
+        # TODO(jamessynge): We should move the driver specific validation into the driver
+        # module (e.g. module.create_mount_from_config). This means we have to adjust the
+        # definition of this method to return a validated but not fully initialized mount
+        # driver.
+        if model != 'bisque':
+            msg = "No port specified for mount in config file. Use simulator = mount for simulator. Exiting."
+            raise error.MountNotFound(msg=msg)
+
+    logger.debug('Creating mount: {}'.format(model))
+
+    module = load_module('pocs.mount.{}'.format(driver))
+
+    # Make the mount include site information
+    mount = module.Mount(location=earth_location)
+
+    logger.debug('Mount created')
+
+    return mount
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -49,7 +49,7 @@ def create_mount_from_config(config, mount_info=None, earth_location=None):
     module = load_module('pocs.mount.{}'.format(driver))
 
     # Make the mount include site information
-    mount = module.Mount(location=earth_location)
+    mount = module.Mount(location=earth_location, *args, **kwags)
 
     logger.debug('Mount created')
 
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -69,7 +69,7 @@ def create_mount_from_config(config, mount_info=None, earth_location=None, *args
     module = load_module('pocs.mount.{}'.format(driver))
 
     # Make the mount include site information
-    mount = module.Mount(location=earth_location, *args, **kwargs)
+    mount = module.Mount(config=config, location=earth_location, *args, **kwargs)
 
     logger.info(f'{driver} mount created')
 
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -51,6 +51,6 @@ def create_mount_from_config(config, mount_info=None, earth_location=None, *args
     # Make the mount include site information
     mount = module.Mount(location=earth_location, *args, **kwags)
 
-    logger.debug('Mount created')
+    logger.debug('Mount created from config')
 
     return mount
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -13,7 +13,7 @@ def create_mount_from_config(config, mount_info=None, earth_location=None):
     logger = get_root_logger()
 
     if 'mount' not in config:
-        logger.info("No mount in config")
+        logger.info("No mount information in config, cannot create.")
         return None
 
     if earth_location is None:
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -32,6 +32,7 @@ def create_mount_from_config(config, mount_info=None, earth_location=None, *args
     Raises:
         `error.MountNotFound`: Missing the serial.port info for a mount which requires it (i.e. not
             a Software Bisque mount).
+        `error.MountNotFound`: Wrong model, it is not bisque and serial port is not specified.
     """
     logger = get_root_logger()
 
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -3,45 +3,66 @@ from glob import glob
 from pocs.mount.mount import AbstractMount  # pragma: no flakes
 from pocs.utils import error
 from pocs.utils import load_module
+from pocs.utils.error import MountNotFound
 from pocs.utils.location import create_location_from_config
 from pocs.utils.logger import get_root_logger
 
 
 def create_mount_from_config(config, mount_info=None, earth_location=None, *args, **kwargs):
-    """ Sets up the mount that will be used by the observatory """
+    """Create a mount instance based on the provided config.
 
+    Creates an instance of the AbstractMount sub-class in the module specified in the config.
+    Specifically, the class must be in a file called pocs/mount/<DRIVER_NAME>.py,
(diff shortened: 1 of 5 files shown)


pocs/mount/__init__.py
```diff
@@ -1 +1,55 @@
...
+    if earth_location is None:
+        site_details = create_location_from_config(config)
+        earth_location = site_details['earth_location']
+
+    if mount_info is None:
+        mount_info = config.get('mount')
+
+    model = mount_info.get('model')
+
+    model = mount_info.get('brand')
```
wtgee: Sorry for the confusion, we have some things that are not named properly in the config.  Let's try to clear this up.  The `driver` is used to load the module so that is okay. Let's remove `brand` from here as well as from the [pocs.yaml](https://github.com/panoptes/POCS/blob/develop/conf_files/pocs.yaml#L43) config file.  But let's also change the `model` in the config to be `model: iOptron iEQ30 Pro`, which is just a descriptive name.

```suggestion
```

pocs/mount/__init__.py
```diff
@@ -1 +1,55 @@
...
+    driver = mount_info.get('driver')
+
+    # See if we have a serial connection
+    try:
+        port = mount_info['serial']['port']
+        if port is None or len(glob(port)) == 0:
+            msg = "Mount port({}) not available. ".format(port) + "Use simulator = mount for simulator. Exiting."
```
wtgee:
```suggestion
            msg = f"Mount port ({port}) not available. Use simulator = mount for simulator."
```

pocs/mount/__init__.py
```diff
@@ -1 +1,55 @@
...
+            raise error.MountNotFound(msg=msg)
+
+    logger.debug('Creating mount: {}'.format(model))
+
+    module = load_module('pocs.mount.{}'.format(driver))
+
+    # Make the mount include site information
+    mount = module.Mount(location=earth_location)
+
+    logger.debug('Mount created')
```
wtgee:
```suggestion
    logger.debug('Mount created from config')
```
The mount itself gives out a message when it is created, so let's make this specific to this function.

pocs/mount/__init__.py
```diff
@@ -1 +1,55 @@
+from glob import glob
+
 from pocs.mount.mount import AbstractMount  # pragma: no flakes
+from pocs.utils import error
+from pocs.utils import load_module
+from pocs.utils.location import create_location_from_config
+from pocs.utils.logger import get_root_logger
+
+
+def create_mount_from_config(config, mount_info=None, earth_location=None):
```
wtgee:
```suggestion
def create_mount_from_config(config, mount_info=None, earth_location=None, *args, **kwargs):
```

Not sure if I said this on the other ones, but let's pass values down to the actual mount creation. See below.

pocs/mount/__init__.py
```diff
@@ -1 +1,55 @@
...
+        if model != 'bisque':
+            msg = "No port specified for mount in config file. Use simulator = mount for simulator. Exiting."
+            raise error.MountNotFound(msg=msg)
+
+    logger.debug('Creating mount: {}'.format(model))
+
+    module = load_module('pocs.mount.{}'.format(driver))
+
+    # Make the mount include site information
+    mount = module.Mount(location=earth_location)
```
wtgee:
```suggestion
    mount = module.Mount(location=earth_location, *args, **kwags)
```

Pass in arbitrary values from the user.

pocs/mount/__init__.py
```diff
@@ -1 +1,56 @@
...
+    if 'mount' not in config:
+        logger.info("No mount information in config, cannot create.")
+        return None
+
+    if earth_location is None:
+        logger.debug(f'No location provided, using values from config.')
+        site_details = create_location_from_config(config)
+        earth_location = site_details['earth_location']
+
+    if mount_info is None:
```
jamessynge: I suggest you move the mount_info block up above the current config['mount'] code (line 15 currently). That way we don't even check for the mount key in config if mount_info is provided.

pocs/mount/__init__.py
```diff
@@ -1 +1,56 @@
...
+        logger.debug(f'No mount info provided, using values from config.')
+        mount_info = config.get('mount')
+
+    model = mount_info.get('model')
```
jamessynge: @wtgee Should we require model be provided? Or use driver as a fallback?

pocs/mount/__init__.py
```diff
@@ -1 +1,56 @@
...
+
+    driver = mount_info.get('driver')
```
jamessynge: Please add a check that driver is valid (especially that it is a non-empty string). For example:

```
if not driver or not isinstance(driver, str):
    raise MountNotFound('Mount info in config is missing a driver name.')
```

pocs/mount/__init__.py
```diff
@@ -1 +1,56 @@
...
+    """ Sets up the mount that will be used by the observatory """
+
+    logger = get_root_logger()
+
+    if 'mount' not in config:
+        logger.info("No mount information in config, cannot create.")
+        return None
+
+    if earth_location is None:
+        logger.debug(f'No location provided, using values from config.')
```
jamessynge: No need for the f at the beginning of the string: there are no substitutions required.

pocs/mount/__init__.py
```diff
@@ -1 +1,56 @@
...
+        site_details = create_location_from_config(config)
+        earth_location = site_details['earth_location']
+
+    if mount_info is None:
+        logger.debug(f'No mount info provided, using values from config.')
```
jamessynge: No need for the f at the beginning of the string: there are no substitutions required.

pocs/mount/__init__.py
```diff
@@ -1 +1,56 @@
...
+
+    model = mount_info.get('model')
+
+    driver = mount_info.get('driver')
+
+    # See if we have a serial connection
+    try:
+        port = mount_info['serial']['port']
+        if port is None or len(glob(port)) == 0:
+            msg = f"Mount port ({port}) not available. Use simulator = mount for simulator."
```
jamessynge: Please use the same single quote characters around all of the string literals (except docstrings), unless the contents of the string include a single quote. We're aiming for consistency, where appropriate.

pocs/mount/__init__.py
```diff
@@ -3,54 +3,75 @@
 from pocs.mount.mount import AbstractMount  # pragma: no flakes
 from pocs.utils import error
 from pocs.utils import load_module
+from pocs.utils.error import MountNotFound
```
wtgee: `error` is already imported, just use `error.MountNotFound`.

pocs/mount/__init__.py
```diff
@@ -1 +1,77 @@
...
+
+def create_mount_from_config(config, mount_info=None, earth_location=None, *args, **kwargs):
+    """Create a mount instance based on the provided config.
+
+    Creates an instance of the AbstractMount sub-class in the module specified in the config.
+    Specifically, the class must be in a file called pocs/mount/<DRIVER_NAME>.py,
+    and the class must be called Mount.
+
+    Args:
+        config: A dictionary of name to value, as produced by panoptes.utils.config.load_config.
```
wtgee:
```suggestion
        config: A dictionary of name to value, as produced by `panoptes.utils.config.load_config`.
```

I'll get to documentation updates one day, in which case we will want this.

pocs/mount/__init__.py
```diff
@@ -1 +1,77 @@
...
+        mount_info: Optional param which overrides the 'mount' entry in config if provided.
+            Useful for testing.
+        earth_location: astropy.coordinates.EarthLocation instance, representing the
```
wtgee:
```suggestion
        earth_location: `astropy.coordinates.EarthLocation` instance, representing the
```

Same thing regarding documentation.

pocs/mount/__init__.py
```diff
@@ -1 +1,77 @@
...
+            location of the mount on the Earth. If not specified, the config must include the
+            observatory's location (Latitude, Longitude and Altitude above mean sea level).
+            Useful for testing.
+        *args: Other positional args will be passed to the concrete class specified in the config.
+        **kwargs: Other keyword args will be passed to the concrete class specified in the config.
+
+    Returns:
+        A class called Mount if the config (or mount_info) is complete. None if neither
```
wtgee:
```suggestion
        An instance of the Mount class if the config (or mount_info) is complete. `None` if neither
```

pocs/mount/__init__.py
```diff
@@ -1 +1,77 @@
...
+        mount_info nor config['mount'] is provided.
+
+    Raises:
+        MountNotFound: Missing the serial.port info for a mount which requires it (i.e. not
```
wtgee:
```suggestion
        `error.MountNotFound`: Missing the serial.port info for a mount which requires it (i.e. not
```

pocs/mount/__init__.py
wtgee: I'm undecided on this. My thought is that the default comes from the config and therefore this should fail if config is empty.

pocs/tests/test_mount.py
```diff
@@ -0,0 +1,30 @@
...
+    conf = config.copy()
+    conf['mount']['driver'] = 'foobar'
+    with pytest.raises(MountNotFound):
+        create_mount_from_config(conf)
+    conf['mount']['driver'] = 1234
+    with pytest.raises(MountNotFound):
+        create_mount_from_config(conf)
+
+
+def test_create_mount_with_earth_location(config):
```
wtgee: Let's also add a basic `test_create_mount_from_config` that doesn't use the `earth_location`.

pocs/mount/__init__.py
jamessynge: I don't think I understand. How can the default come from the config when it is missing?

pocs/mount/__init__.py
```diff
@@ -1 +1,77 @@
...
+        site_details = create_location_from_config(config)
+        earth_location = site_details['earth_location']
+
+    driver = mount_info.get('driver')
+    if not driver or not isinstance(driver, str):
+        raise error.MountNotFound('Mount info in config is missing a driver name.')
+
+    model = mount_info.get('model', driver)
+
+    if driver != 'simulator' and model != 'simulator':
```
wtgee: Just checking the `driver` should be enough but this is okay to leave here for now.

pocs/mount/__init__.py
```diff
@@ -1 +1,77 @@
...
+        # See if we have a serial connection
+        try:
+            port = mount_info['serial']['port']
+            if port is None or len(glob(port)) == 0:
+                msg = f'Mount port ({port}) not available. Use simulator = mount for simulator.'
+                raise error.MountNotFound(msg=msg)
+        except KeyError:
+            if model != 'bisque':
```
wtgee: Can you add back in the big comment that describes what is going on here?

pocs/tests/test_mount.py
```diff
@@ -0,0 +1,35 @@
...
+    conf['mount']['driver'] = 'foobar'
+    with pytest.raises(MountNotFound):
+        create_mount_from_config(conf)
+    conf['mount']['driver'] = 1234
+    with pytest.raises(MountNotFound):
+        create_mount_from_config(conf)
+
+
+def test_create_mount_with_earth_location(config_with_simulated_mount):
+    conf = config_with_simulated_mount.copy()
```
wtgee: Move this line into the fixture and then remove from here and next test so it's not needed each time.

pocs/mount/__init__.py
ImmortalRabbit: > Can you add back in the big comment that describes what is going on here?
 `error.MountNotFound`: Wrong model, it is not bisque and serial port is not specified..
Will this comment describe it? Sorry, I am not pretty sure how to describe it. Generally, I see what is going on but I am still learning how to write commentary.

wtgee: Yeah, sorry, the original comment isn't totally relevant I suppose. Posting here and then creating a separate issue.

```
                # TODO(jamessynge): We should move the driver specific validation into the driver
                # module (e.g. module.create_mount_from_config). This means we have to adjust the
                # definition of this method to return a validated but not fully initialized mount
                # driver.
```

In general the `create_mount_from_config` will need to handle specific cases that are different. For now let's say:

```suggestion
            # Note: see Issue #999
            if model != 'bisque':
```

Issue #853, opened 2019-06-24
Scheduler creation requires observer


ImmortalRabbit:
**Describe the bug**
Scheduler creation requires observer to be passed. However, it can be created from same config, in case if it didn't passed. 

**To Reproduce**
Scheduler creation requires also config which can be used to create observer if observer is None.

**Expected behavior**

**Related Issues**
Please link any related Issues if known.

**Screenshots**

**Additional context**

PR #854, opened 2019-06-24
Creation of scheduler can create observer if it's None

ImmortalRabbit:
Creation of scheduler required to observer to be passed. However, when observer is None, it can be created from config. 

## Description

During creation of scheduler I check if observer is None then it creates observer from config. 

## Related Issue

#853 
## Types of changes

## Checklist:

### DIFF
--- a/pocs/scheduler/__init__.py
+++ b/pocs/scheduler/__init__.py
@@ -9,6 +9,7 @@ from pocs.scheduler.scheduler import BaseScheduler  # pragma: no flakes
 from pocs.utils import error
 from pocs.utils import horizon as horizon_utils
 from pocs.utils import load_module
+from pocs.utils.location import create_location_from_config
 from pocs.utils.logger import get_root_logger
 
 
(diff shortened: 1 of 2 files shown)


pocs/scheduler/__init__.py
```diff
@@ -22,8 +23,8 @@ def create_scheduler_from_config(config, observer=None):
         return None
 
     if not observer:
-        logger.info("No valid Observer found.")
-        return None
+        site_details = create_location_from_config(config)
```
wtgee:
```suggestion
        logger.debug(f'No Observer provided, creating from config.')
        site_details = create_location_from_config(config)
```

wtgee: We can have plenty of debug level comments, especially if doing something the user might not expect.

pocs/tests/test_scheduler.py
```diff
@@ -22,7 +22,7 @@ def test_bad_scheduler_fields_file(config):
 
 
 def test_no_observer(config):
-    assert create_scheduler_from_config(config, observer=None) is None
+    create_scheduler_from_config(config, observer=None)
```
wtgee: Without an `assert` statement there isn't actually any test performed. Probably want to check the return type is actually a scheduler.

Issue #856, opened 2019-06-24
Create convenience methods for adding/removing dome to Observatory


ImmortalRabbit:
*<Enter description of change or enhancement here>*
**Describe the solution you'd like**
We want to make an add_dome and remove_dome convenience method on the Observatory class, similar to the add_camera and remove_camera methods. Because we want to dynamically support adding/removing of dome, we should have convenience methods that can handle any required details.

**Describe alternatives you've considered**
Could just assign directly to the .dome property, but for consistency we should have actual getter/setter methods.

**Additional context**
Part of GSoC 2019 decoupling work.

**Related Issues/PRs**
If there are any related Issues or PRs list them here:

PR #857, opened 2019-06-27
Dynamic adding/removing dome.

ImmortalRabbit:
add_dome and remove_dome were added.

## Description

add_dome does self.dome = dome and remove_dome does self.dome = None.

## Related Issue

#856 

## Types of changes

## Checklist:

### DIFF
--- a/pocs/observatory.py
+++ b/pocs/observatory.py
@@ -62,9 +62,9 @@ class Observatory(PanBase):
         self._image_dir = self.config['directories']['images']
         self.logger.info('\t Observatory initialized')
 
-##########################################################################
-# Helper methods
-##########################################################################
+    ##########################################################################
+    # Helper methods
+    ##########################################################################
 
     def is_dark(self, horizon='observe', at_time=None):
         """If sun is below horizon.
@@ -90,9 +90,9 @@ class Observatory(PanBase):
 
         return is_dark
 
-##########################################################################
-# Properties
-##########################################################################
+    ##########################################################################
+    # Properties
+    ##########################################################################
 
     @property
     def sidereal_time(self):
@@ -167,9 +167,10 @@ class Observatory(PanBase):
             can_observe = False
 
         return can_observe
-##########################################################################
-# Device Getters/Setters
-##########################################################################
+
+    ##########################################################################
+    # Device Getters/Setters
+    ##########################################################################
 
     def add_camera(self, cam_name, camera):
         """Add camera to list of cameras as cam_name.
@@ -221,8 +222,8 @@ class Observatory(PanBase):
         self.logger.info('Removing scheduler')
         self.scheduler = None
 
-    def add_dome(self, dome):
-        """Add dome.
+    def set_dome(self, dome=None):
+        """Set's dome or remove's dome.
 
         Args:
             dome (`pocs.dome.AbstractDome`): An instance of the `~AbstractDome` class.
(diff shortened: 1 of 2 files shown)


pocs/observatory.py
```diff
@@ -12,6 +12,7 @@
 
 from pocs.base import PanBase
 from pocs.camera import AbstractCamera
+from pocs.dome.bisque import Dome
```
wtgee: The `bisque` dome is a specific type of dome. We want to use the `AbstractDome` instead.

pocs/observatory.py
```diff
@@ -216,10 +217,27 @@ def add_scheduler(self, scheduler):
...
-        """Remove scheduler. """
+        """Remove scheduler."""
         self.logger.info('Removing scheduler')
         self.scheduler = None
 
+    def add_dome(self, dome):
+        """Add dome.
+
+        Args:
+            dome (`pocs.dome.bisque.Dome`): An instance of the `~Dome` class.
```
wtgee: Should be changed to the `AbstractDome`.

ImmortalRabbit:
> These need some test to go along with them inside `test_observatory`.

I thought I will do in a separate pull request.

wtgee:
> > These need some test to go along with them inside `test_observatory`.
> 
> I thought I will do in a separate pull request.

Every PR should have tests if possible. There is no reason to separate them. Tests are part of the PR (it's in the checklist up above even).

pocs/observatory.py
```diff
@@ -61,9 +63,9 @@ def __init__(self, cameras=None, scheduler=None, dome=None, *args, **kwargs):
         self._image_dir = self.config['directories']['images']
         self.logger.info('\t Observatory initialized')
 
-##########################################################################
-# Helper methods
-##########################################################################
+    ##########################################################################
```
wtgee: Do you have an autoformatter that was indenting these?

pocs/observatory.py
```diff
@@ -216,13 +219,28 @@ def add_scheduler(self, scheduler):
...
-        """Remove scheduler. """
+        """Remove scheduler."""
         self.logger.info('Removing scheduler')
         self.scheduler = None
 
-##########################################################################
-# Methods
-##########################################################################
+    def set_dome(self, dome=None):
+        """Set's dome or remove's dome.
```
wtgee:
```suggestion
        """Set's dome or remove the dome for the `Observatory`.
        
        Note:
            The default argument of `None` will remove the current dome.
```

pocs/tests/test_observatory.py
```diff
@@ -156,6 +156,19 @@ def test_add_remove_scheduler(config, observatory, caplog):
         observatory.add_scheduler("scheduler")
 
 
+def test_set_dome(config_with_simulated_dome):
+    conf = config_with_simulated_dome.copy()
+    dome = create_dome_from_config(conf)
+    obs = Observatory(config=conf, dome=dome)
+    assert obs.has_dome is True
+    obs.set_dome()
```
wtgee: The more I see this the more I dislike it in term of removing the dome.

I'm thinking we should make this a positional parameter rather than a keyword, so that the user has to specifically say `obs.set_dome(None)` to remove the dome.  Then `obs.set_dome()` with nothing would be an error.  Okay for now but I will make an Issue out of this.

pocs/observatory.py
ImmortalRabbit: I think yes, should I stop doing it?

pocs/tests/test_observatory.py
wtgee: #865 created.

pocs/observatory.py
wtgee: We should just make sure it has the same settings that the project uses. We'll worry about it when you're back.

PR #875, opened 2019-08-17
db_name as fixture

wtgee:
## Description
Turn the db_name into a fixture rather than hard-coding throughout.

## Related Issue
Pulled from #859 

Depends on #876 which depends on #877 

## How Has This Been Tested?
Regular testing suite run.


conftest.py
```diff
@@ -215,31 +220,29 @@ def can_connect_to_mongo():
...
-        db_type=db_type, db_name='panoptes_testing', logger=get_root_logger(), connect=True)
+        db_type=db_type, db_name=db_name, logger=get_root_logger(), connect=True)
 
 
 @pytest.fixture(scope='function')
-def memory_db():
-    PanDB.permanently_erase_database(
-        'memory', 'panoptes_testing', really='Yes', dangerous='Totally')
-    return PanDB(db_type='memory', db_name='panoptes_testing')
+def memory_db(db_name):
```
jamessynge: Why does it make sense to add db_name as a positional param to this fixture?

wtgee: You mean for the `memory_db` in particular? I suppose it is not necessary as it will never persist beyond the tests but it was mostly just for the consistency of having all the db types having the same name that is easily changed in one place. Can remove from here.

PR #876, opened 2019-08-17
IERS Mirror

wtgee:
Use our own mirror for the IERS data.

## Description
The IERS mirrors go down way too often and it breaks all kinds of things. Here we use our own mirror for download. This mirror is updated every few days so should always contain a local copy.

Depends on #877


pocs/utils/data.py
```diff
@@ -7,6 +7,9 @@
 import sys
 import warnings
 
+from astropy.utils import iers
+iers.IERS_A_URL = 'https://storage.googleapis.com/panoptes-resources/iers/ser7.dat'
```
mimming: How does the file at this path get updated?   A cron job somewhere, or is the plan to do it manually for now?

wtgee: There's a Cloud Scheduler job pulling it down once per week.

Note that there's a bunch of discussion still going on via astroplan and astropy about these IERS tables. Not sure we will use this solution long-term or not.

wtgee: Note that I've reverted this commit while dealing with the IERS in another way. #890

PR #886, opened 2019-08-20
Cleanup base constructor calls - WIP

wtgee:
Consistency with how we are calling the `PanBase` base class constructor. Pulled from #859.

### DIFF
--- a/pocs/filterwheel/filterwheel.py
+++ b/pocs/filterwheel/filterwheel.py
@@ -29,7 +29,7 @@ class AbstractFilterWheel(PanBase):
                  timeout=None,
                  serial_number='XXXXXX',
                  *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        PanBase.__init__(self, *args, **kwargs)
 
         self._model = model
         self._name = name
--- a/pocs/focuser/focuser.py
+++ b/pocs/focuser/focuser.py
@@ -69,7 +69,8 @@ class AbstractFocuser(PanBase):
                  autofocus_merit_function_kwargs=None,
                  autofocus_mask_dilations=None,
                  *args, **kwargs):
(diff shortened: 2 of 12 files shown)


jamessynge:
IIUC, if a class has a single base class called `BaseClassName`, then `super().__init__(self)` and `BaseClassName.__init__(self)` should be identical. But if the class has multiple base classes, then the first form will call all of the base classes' `__init__` methods in the MRO determined order, whereas the second form will only call the specifically named `__init__` method, and the other base classes will not be initialized. That seems problematic.

So, please tell me more about why this change is appropriate.

wtgee:
> IIUC, if a class has a single base class called `BaseClassName`, then `super().__init__(self)` and `BaseClassName.__init__(self)` should be identical. But if the class has multiple base classes, then the first form will call all of the base classes' `__init__` methods in the MRO determined order, whereas the second form will only call the specifically named `__init__` method, and the other base classes will not be initialized. That seems problematic.
> 
> So, please tell me more about why this change is appropriate.

The change was supposed to be making it so that anything that only has `PanBase` as a parent uses the single form (i.e. `BaseClassName`) whereas inherited classes lower use the `super()` call. So for instance the `AbstractMount` has only `PanBase` has a parent so uses the specific form whereas the `Bisque` mount uses the `super()` call.

But, TBH, I'm not exactly sure why I went this route. There were some things that were not working properly with how `PanBase` was pulling in the config and logger and this resolved the issue by explicitly calling `PanBase`.   

I'll leave this open for now as we merge in the other items and perhaps it will help me figure out what the issue was again, but this will allow me to pull these changes out of #859 and still keep them around somewhere. I'll mark as WIP for now.

wtgee:
In light of comments above, I'm closing this until a need is identified.

PR #887, opened 2019-08-21
Slew to home blocking

wtgee:
Provides a parameter that allows the `slew_to_home` command to block until home is reached. Defaults to `False` for backwards compatibility.

Replaces some custom blocking in various places with new parameter.

Pulled from #859.

### DIFF
--- a/pocs/mount/mount.py
+++ b/pocs/mount/mount.py
@@ -514,20 +514,30 @@ class AbstractMount(PanBase):
 
         return success
 
-    def slew_to_home(self):
+    def slew_to_home(self, blocking=False):
         """ Slews the mount to the home position.
 
         Note:
             Home position and Park position are not the same thing
+
+        Args:
+            blocking (bool, optional): If command should block while slewing to
+                home, default False.
 
         Returns:
             bool: indicating success
         """
         response = 0
 
+        block_time = 2  # seconds
+
         if not self.is_parked:
             self._target_coordinates = None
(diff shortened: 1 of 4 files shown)


jamessynge:
FWIW, I'm reminded by reading the description of this PR that I encountered a problem where slewing never ended, so the dome was stuck open until I noticed a day+ later.

pocs/mount/mount.py
```diff
@@ -514,28 +511,38 @@ def slew_to_target(self):
 
         return success
 
-    def slew_to_home(self):
+    def slew_to_home(self, blocking=False):
```
jamessynge: Should this have a maximum time to block (e.g. 10 minutes)?

wtgee: Yes, agreed. I almost marked this WIP until I could add that in as it does seem like a potential bad failure. I'll add before merge and re-request a review. Thanks!

wtgee: I've added a timeout, however it's not clear to me what the behavior should be. We slew to home before we park, so if we have a timeout while slewing to home we can't actually then attempt a park.  Note also that #889 potential adds some confusion to this as that new park behaviour is based off movement from the home position.

> FWIW, I'm reminded by reading the description of this PR that I encountered a problem where slewing never ended, so the dome was stuck open until I noticed a day+ later.

IIRC you were stuck in a slew to target but not a slew to home. We would need to add a timeout there as well.

PR #888, opened 2019-08-21
Small mount updates

wtgee:
Mostly minor updates to add more logging, comments, and consistency with constructor calls. Pulled from #859

### DIFF
--- a/pocs/mount/__init__.py
+++ b/pocs/mount/__init__.py
@@ -7,7 +7,11 @@ from pocs.utils.location import create_location_from_config
 from pocs.utils.logger import get_root_logger
 
 
-def create_mount_from_config(config, mount_info=None, earth_location=None, *args, **kwargs):
+def create_mount_from_config(config,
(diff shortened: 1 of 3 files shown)


pocs/mount/__init__.py
```diff
@@ -7,7 +7,11 @@
 from pocs.utils.logger import get_root_logger
 
 
-def create_mount_from_config(config, mount_info=None, earth_location=None, *args, **kwargs):
+def create_mount_from_config(config,
```
jamessynge: FWIW, if we expect the caller will usually pass in mount_info, it begins to seem as if config should be a kwarg (i.e. optional).

wtgee: Agreed. In the switch to `panoptes-utils` the `config` parameter disappears entirely so I'm inclined to leave this as is for now as it is mostly temporary. With that switch over we can figure out what would be the best.

PR #892, opened 2019-09-08
Set DSLR camera time correctly

### LINKED ISSUE #891: Set DSLR Camera time
**Describe the bug**
The camera times on the DSLR are not being synced to the host computer during setup, leading to inconsistent camera times.

Note that there is a command in [canon_gphoto2.py#L52](https://github.com/panoptes/POCS/blob/develop/pocs/camera/canon_gphoto2.py#L52), which means it is not working so needs to be debugged.

You can see this as a discrpeancy for the listed times on panoptes-data.net.

See https://github.com/gphoto/libgphoto2/issues/443

Edit: note that this seems to be the case for any (canon?) camera and could therefore be tested by anyone with a dslr.

See also #953

wtgee:
These were incorrectly trying to be set via the index instead of the value.

Closes #891.
Related to #953 

Reported by @preethi524

Note: the timezone needs to be set to UTC manually on the camera for this to be correct. The daylight savings should also be disabled.

Edit: see discussion in https://github.com/gphoto/libgphoto2/issues/443


wtgee:
There seems to be some issue on the gphoto2 (or libgphoto2) side of things where you can't simultaneously set the `datetime` and the `datetimeutc`. Still trying to work through it. Best bet would probably be to set the camera to UTC time and only use that.

PR #893, opened 2019-09-08
Scheduler updates

wtgee:
* Duration constraint has a higher default weight (randomly picked)
* Altitude constraint has a score normalized to 1. instead of 100.
* Dispatch scheduler:
    * **Final score is multipled by priority instead of added to**
    * Base score is 0.0 instead of 1.0
    * More logging output showing scoring

@joshwalawender I can't remember why we have been adding the priority instead of mulitiplying it although I do remember that was a conscious choice. The Denny paper is not explicit about this (in my quick re-scanning of the paper).

Note: These were some changes I had made on various nights to both PAN001 and PAN008 for testing some scheduling and I wanted to get the code into a PR rather than have the changes hanging out on those units.

### DIFF
--- a/pocs/scheduler/__init__.py
+++ b/pocs/scheduler/__init__.py
@@ -54,7 +54,7 @@ def create_scheduler_from_config(config, observer=None):
             constraints = [
                 Altitude(horizon=horizon_line),
                 MoonAvoidance(),
-                Duration(default_horizon)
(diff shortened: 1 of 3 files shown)


joshwalawender:
@wtgee I don't recall why we added priority in.  I would think using it as a multiplicative factor as you say would be better.

PR #894, opened 2019-09-12
Mount updates - slew_to_target timeout

wtgee:
## Description
* `slew_to_target` has an optional `blocking` parameter to match #887.
* Mount simulator
    * Update simulate `slew_to_target` to actually use the `AbstractMount` code.

What's not entirely clear here is what should be done if the `Timeout` is actually raised but this at least adds the mechanism for it.

## Related Issue
#887

### DIFF
--- a/pocs/state/states/default/pointing.py
+++ b/pocs/state/states/default/pointing.py
@@ -94,7 +94,7 @@ def on_enter(event_data):
                         if observation.field is not None:
                             pocs.logger.debug("Slewing back to target")
                             pocs.observatory.mount.set_target_coordinates(observation.field)
-                            pocs.observatory.mount.slew_to_target()
+                            pocs.observatory.mount.slew_to_target(blocking=True)
                 else:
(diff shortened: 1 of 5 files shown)


pocs/state/states/default/pointing.py
```diff
@@ -94,7 +94,11 @@ def on_enter(event_data):
                         if observation.field is not None:
                             pocs.logger.debug("Slewing back to target")
                             pocs.observatory.mount.set_target_coordinates(observation.field)
-                            pocs.observatory.mount.slew_to_target()
+                            pocs.observatory.mount.slew_to_target(blocking=True)
+
+                    if img_num == (num_pointing_images - 1):
+                        pocs.logger.info(f'Separation outside threshold but at max corrections. ' +
+                                         'Will proceed to observations.')
```
wtgee: If separation is largely than some threshold after `num_pointing_images` then the system should **not** proceed but should shut down.

PR #898, opened 2019-09-26
Fix import of ctypes.util

AnthonyHorton:
## Description

The `pocs.utils.library.load_library()` function was calling the `ctypes.util.find_library` function but `ctypes.util` was not being imported in `pocs/utils/library.py` so this call was failing. This was causing SDK `Camera` classes (e.g. SBIG, FLI, ZWO) to fail on initialisation unless the full path to their driver library was explicitly specified, which avoids the call to `ctypes.util.find_library()`.

The changes here are a duplicate of changes made to fix the same bug in `panoptes-utils`, in https://github.com/panoptes/panoptes-utils/pull/72. Because of the impact of this bug on Huntsman operations it seems worthwhile to fix it here too, rather than waiting for POCS to transition to using `panoptes-utils` for this function.

## How Has This Been Tested?

Test added to `pocs/tests/utils/test_utils.py` to ensure call to `ctypes.util.find_library` now works.

## Types of changes

## Checklist:

### DIFF
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -12,7 +12,7 @@ from pocs.utils import load_module
 from pocs.utils import CountdownTimer
 from pocs.utils import error
 from pocs.camera import list_connected_cameras
-from pocs.utils.library import load_library
+from pocs.utils.library import load_library as load_c_library
 
 
 def test_error(capsys):
--- a/pocs/tests/utils/test_utils.py
+++ b/pocs/tests/utils/test_utils.py
@@ -12,6 +12,7 @@ from pocs.utils import load_module
 from pocs.utils import CountdownTimer
 from pocs.utils import error
 from pocs.camera import list_connected_cameras
+from pocs.utils.library import load_library
 
 
 def test_error(capsys):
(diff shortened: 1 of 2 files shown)


pocs/tests/utils/test_utils.py
```diff
@@ -12,6 +12,7 @@
 from pocs.utils import CountdownTimer
 from pocs.utils import error
 from pocs.camera import list_connected_cameras
+from pocs.utils.library import load_library
```
wtgee:
```suggestion
from pocs.utils.library import load_library as load_c_library
```

I'm opting to change this here to keep consistency with what the name will be in `panoptes-utils` but feel free to reject this change and change line 47 instead.

wtgee:
That's a pretty large decrease in coverage (0.76%), which seems odd.  I was all proud of myself for raising it by about that much earlier today.  I think that report must be off somehow.

AnthonyHorton:
> That's a pretty large decrease in coverage (0.76%), which seems odd. I was all proud of myself for raising it by about that much earlier today. I think that report must be off somehow.

Yeah, I was a bit surprised how I could _add a test_ and still end up reducing overall coverage. Definitely something screwy in how coverage percentages are calculated.

Issue #899, opened 2019-10-03
Incorrectly scaled ZWO camera pixel values


AnthonyHorton:
**Describe the bug**
The ZWO ASI cameras capture images using a 10 or 12 bit analogue-to-digital converter. The resulting pixel values are padded to 16 bits in the 'raw' image data returned by the ASI camera driver library. Somewhat counter-intuitively the pixel data is padded with trailing zeros (least significant bits) instead of leading zeros (most significant bits), which has the effect of multiplying the real raw pixel values by a factor of 64 or 16. This then throws off any attempt to convert images from ADU to electron units using the `EGAIN` FITS keyword value taken from the camera driver (which doesn't account for the 64x/16x scaling) and any subsequent analysis of read noise, dark current, full well depth, quantum efficiency, etc...

**To Reproduce**
1. Take an exposure with a ZWO ASI camera operating in `RAW16` format using the `pocs.camera.zwo.Camera` class.

**Expected behavior**
A raw image FITS file with pixel values in the range 0 to 1023 or 4095 with any integer value within those ranges possible.

**Actual behaviour**
A raw image FITS file with pixels values which are all multiples of 64 or 16.

AnthonyHorton:
Closed by #902 & #903

PR #902, opened 2019-10-04
Fix scaling of ZWO camera pixel values

AnthonyHorton:
Fixes the incorrect scaling of pixel values in image data from ZWO ASI cameras.

## Description
Adds rescaling when reading image data from the camera driver in RAW16 format. The raw image data is shifted by the appropriate number of bits to the right to remove the zero padding that ZWO ASI cameras incorrectly apply to the least significant bits instead of the most significant bits.

## Related Issue
#899 

## How Has This Been Tested?
A regression test has been added to the unit tests.

## Types of changes

## Checklist:


AnthonyHorton:
I can stick in a regression test, but it'll inevitably be hardware dependent.

AnthonyHorton:
Added a regression test, at which point it occurred to me that it would also be good to add bit depth to the FITS headers (as `BITDEPTH`) for cameras with ADC bit depth that differs from the bits per pixel (`BITPIX`) of the data type used.

Issue #900, opened 2019-10-03
Use actual serial number of ZWO ASI cameras


AnthonyHorton:
*<Enter description of change or enhancement here>*
The changelog of the ZWO ASI camera driver library (https://astronomy-imaging-camera.com/software/change_log/change_ASISDK_Unix.txt) notes the addition of a `ASIGetSerialNumber` function in V1.14.0227 of the camera driver library. Before this was added there was no way to access the serial number of a connected ZWO ASI camera. Consequently the `pocs.camera.zwo.Camera` class worked around this issue (with considerable effort) by instead using the camera 'name' (an up to 8 character string), which can be set by the user and is stored in the camera firmware. POCS requires a unique ID with which to identify cameras and the 'name' was the only solution available at the time, but it is not ideal for several reasons:
1. Before a ZWO camera can be used with POCS the 'name' must be set. This must be done either with ZWO ASICAP application or the low level `pocs.camera.libasi` library.
1. It is up to the user to ensure that the 'names' given to ZWO cameras are unique.
1. The 'name' can be changed, meaning it is not possible to associate previously taken FITS files with a particular camera with total confidence.

**Describe the solution you'd like**
The POCS ZWO ASI Camera class should be changed to use the camera's real serial number now that it can be access through the camera driver API.

**Describe alternatives you've considered**
The use of ZWO camera 'names' as an alternative to a serial number could be continued.

AnthonyHorton:
Closed by #904

PR #904, opened 2019-10-09
Use real serial numbers for ZWO ASI cameras

AnthonyHorton:
Taking advantage of the function added to the ZWO SDK in order to use the camera's real serial number as its UID instead of the user set 'string ID'.

## Description
This PR adds to the methods of the `pocs.camera.libasi.ASIDriver` class so that there are methods corresponding to all functions of the ZWO ASI camera SDK API as of version v1.14.0715. All but one of these methods are implemented. Most of the new methods are of niche interest except `.get_serial_number()`, which calls the new `ASIGetSerialNumber` function.

Now that real camera serial numbers are accessible via the camera driver the `get_cameras()` function has been modified to return these serial numbers instead of the user set string ID as before, if possible. This brings the ZWO ASI cameras in line with the other SDK cameras (SBIG, FLI) in using serial numbers for their UID.

## Related Issue
Addresses #900.

## How Has This Been Tested?
Most of these changes are hardware dependent and cannot be tested without real ZWO cameras. Most of the new methods relate to trigger or guiding functionality that is only supported by a subset of ZWO cameras (which we don't have). I will test what I can next week and update this PR with details of the confirmed functionality.

## Types of changes

Breaking only to the extent that any existing config files for ZWO cameras would need to be changed to specify the camera serial numbers instead of string IDs.

## Checklist:

### DIFF
--- a/pocs/camera/libasi.py
+++ b/pocs/camera/libasi.py
@@ -430,7 +430,7 @@ class ASIDriver(AbstractSDKDriver):
                             ctypes.bytef(pin_high),
                             ctypes.byref(delay),
                             ctypes.byref(duration))
-        self.logger.debug("Got trigger config from camera {}".format(camera_IF))
+        self.logger.debug("Got trigger config from camera {}".format(camera_ID))
         return TrigOutput(pin).name, bool(pin_high), int(delay), int(duration)
 
     def set_trigger_ouput_io_conf(self, camera_ID, pin, pin_high, delay, duration):
--- a/pocs/camera/libasi.py
+++ b/pocs/camera/libasi.py
@@ -337,7 +337,7 @@ class ASIDriver(AbstractSDKDriver):
                             GuideDirection[direction])
         dname = GuideDirection[direction].name
         msg = f"PulseGuide on camera {camera_ID} on in direction {dname}."
-        self.logger.debug(msg}
+        self.logger.debug(msg)
 
     def pulse_guide_off(self, camera_ID, direction):
         """Turn off PulseGuide on ST4 port of given camera in given direction."""
@@ -346,7 +346,7 @@ class ASIDriver(AbstractSDKDriver):
                             GuideDirection[direction])
         dname = GuideDirection[direction].name
         msg = f"PulseGuide on camera {camera_ID} off in direction {dname}."
-        self.logger.debug(msg}
+        self.logger.debug(msg)
 
     def get_gain_offset(self, camera_ID):
         """Get pre-setting parameters."""
@@ -371,7 +371,7 @@ class ASIDriver(AbstractSDKDriver):
                             ctypes.byref(modes_struct.modes))
         supported_modes = []
         for mode_int in modes_struct.modes:
-            if mode_int = CameraMode.END:
+            if mode_int == CameraMode.END:
                 break
             supported_modes.append(CameraMode(mode_int).name)
 
@@ -402,7 +402,7 @@ class ASIDriver(AbstractSDKDriver):
         self._call_function('ASISendSoftTrigger',
                             camera_ID,
                             int(bool(start_stop_signal)))
-        self.logger.debug('Soft trigger sent to camera {}.'format(camera_ID))
+        self.logger.debug('Soft trigger sent to camera {}.'.format(camera_ID))
 
     def get_serial_number(self, camera_ID):
         """Get serial number of the camera with given integer ID.
@@ -687,6 +687,7 @@ class TrigOutput(enum.IntEnum):
     PINA = 0  # Only Pin A output
     PINB = enum.auto()  # Only Pin B outoput
     NONE = -1
+
 
 @enum.unique
 class ErrorCode(enum.IntEnum):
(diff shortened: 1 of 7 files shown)


pocs/camera/libasi.py
```diff
@@ -64,17 +70,30 @@ def get_cameras(self):
...
+                serial_number = self.get_serial_number(camera_ID)
+            except RuntimeError as err:
+                # If at first you don't succeed, try, except, else, finally again.
+                self.logger.error(f"Error getting serial number: {err}")
+                try:
+                    string_ID = self.get_ID(camera_ID)
+                except RuntimeError as err:
+                    self.logger.error(f"Error getting string ID: {err}")
+                    msg = f"Skipping ZWO ASI camera {camera_ID} with no serial number or string ID."
+                    self.logger.warning(msg)
```
wtgee: You have a logger `error` with no serial id but a `warning` here when it's neither. Seems like you would want the error here as that means the camera is being skipped, but just a warning on the lack of real serial id.

pocs/camera/libasi.py
```diff
@@ -97,6 +131,18 @@ def get_camera_property(self, camera_index):
...
 
+    def get_camera_property_by_id(self, camera_ID):
+        """Get properties of the camera with a given integer ID."""
+        camera_info = CameraInfo()
+        self._call_function('ASIGetCameraPropertyByID',
+                            camera_ID,
+                            ctypes.byref(camera_info))
+
+        pythonic_info = self._parse_info(camera_info)
+        self.logger.debug("Got info from camera {}, {}".format(pythonic_info['camera_ID'],
```
wtgee: Just a note that you could do `"{camera_ID}, {name}".format(**pythonic_info)` if you wanted. Might keep it on one line. ;)

And where are those multi-line suggestions when you need them!

pocs/camera/libasi.py
```diff
@@ -256,6 +302,147 @@ def set_start_position(self, camera_ID, start_x, start_y):
...
+        self._call_function('ASIGetCameraMode',
+                            camera_ID,
+                            ctypes.byref(mode))
+        mode_name = CameraMode(mode).name
+        self.logger.debug('Camera {} is in trigger mode {}'.format(camera_ID, mode_name))
+        return mode_name
+
+    def set_camera_mode(self, camera_ID, mode_name):
+        """Set trigger mode for camera with given integer ID."""
+        mode = CamerMode[mode_name]
```
wtgee:
```suggestion
        mode = CameraMode[mode_name]
```

pocs/camera/libasi.py
```diff
@@ -256,6 +302,147 @@ def set_start_position(self, camera_ID, start_x, start_y):
...
+        pin_high = ctypes.c_int()
+        delay = ctypes.c_long()
+        duration = ctypes.c_long()
+        self._call_function('ASIGetTriggerOutputIOConf',
+                            camera_ID,
+                            ctypes.byref(pin),
+                            ctypes.bytef(pin_high),
+                            ctypes.byref(delay),
+                            ctypes.byref(duration))
+        self.logger.debug("Got trigger config from camera {}".format(camera_IF))
```
wtgee:
```suggestion
        self.logger.debug("Got trigger config from camera {}".format(camera_ID))
```

pocs/camera/sbig.py
```diff
@@ -32,6 +32,11 @@ def __del__(self):
 
 # Properties
 
+    @property
+    def egain(self):
+        """Image sensor gain in e-/ADU as reported by the camera."""
+        return self._info['readout modes']['RM_1X1']['gain']
```
wtgee: Will this always be there or does it need some exception handling?

pocs/camera/libasi.py
AnthonyHorton: Yeah, I kept changing my mind about whether these 3 should be `warning` or `error`. `warning`, `warning`, `error` is probably more logical & self-consistent than what I currently have.

Part of the problem is that `._call_function()` doesn't (currently) make any distinctions based on which error code it gets back from the driver library, it just raises a generic `RuntimeError` for everything and embeds the error name in the exception's sole argument. This makes it a bit awkward for code that's catching these exceptions to respond in the most appropriate way. What `._call_function()` should probably be doing is raising `RuntimeError` with multiple arguments e.g. (human readable summary, SDK function name, error code, error name) so that it's easier for higher level code to work out what has happened or, even better, `._call_function()` should be raising different custom exceptions for each type of error, again with multiple attributes to make it easy to write conditional error handling code.

As an example, `ASIGetSerialNumber()` can return `ASI_ERROR_GENERAL_ERROR` if the camera doesn't have a serial number (which probably should just be a `warning` if it happens in `get_cameras()`) or `ASI_ERROR_CAMERA_CLOSED` if the camera is not open (which would definitely merit an `error`).

pocs/camera/sbig.py
AnthonyHorton: I think it will always be there. I should be accessing it via the read only property though, i.e. `self.properties['readout modes']['RM_1x1']['gain']` .

`self._info` is the dictionary of stuff that is read from the camera driver on `.connect()`. That will always contain `'readout modes'` for SBIG cameras (they're essential for taking exposures), and those `'readout modes'` will always include `'RM_1x1'` because that's the unbinned readout mode, and finally any readout mode must have a `gain` because it's part of the readout mode C `struct`.

pocs/camera/libasi.py
AnthonyHorton: Written up as #905, will be the subject of another PR.

PR #909, opened 2019-10-18
Make SDK camera simulator accessible

AnthonyHorton:
Make it possible to use the other camera simulator class via config files and `create_cameras_from_config()`, etc.

## Description
`create_cameras_from_config()` and related functions can only create `Camera` objects which can be imported as `pocs.camera.<model name>.Camera`.  There are however two simulated `Camera` classes, in `pocs/camera/simulator/dslr.py` and `pocs/camera/simulator/ccd.py`. A `from pocs.camera.simulator.dslr import Camera` in `pocs/camera/simulator/__init__.py` makes the DSLR camera simulator accessible using the model name `simulator`, but there is currently no equivalent way to access the SDK camera simulator. This PR imports `pocs.camera.simulator.ccd` into `poc.camera` as `simulator_sdk` so that the 2nd type of camera simulator can be created using the model name `simulator_sdk`. This is required for `huntsman-pocs` testing.

## Types of changes

## Checklist:


pocs/camera/__init__.py
```diff
@@ -12,6 +12,7 @@
 
 from pocs.camera.camera import AbstractCamera  # pragma: no flakes
 from pocs.camera.camera import AbstractGPhotoCamera  # pragma: no flakes
+from pocs.camera.simulator import ccd as simulator_sdk
```
wtgee: My concern here is that it will be importing this for every camera that is used. Can we put this in `huntsman-pocs` or is it necessary here?

AnthonyHorton: Good point. I'd say there probably ought to some way to instantiate the SDK camera simulators from a config file in POCS itself though.

AnthonyHorton: Tricky to do this in `huntsman-pocs`. The problem is that camera creation in both `POCS` _and_ `huntsman-pocs` assumes **all** `Camera` classes are `pocs.camera.<model name>.Camera`.

To work around this in `huntsman-pocs` you'd need to do something like 
```
import pocs.camera.simulator.ccd as pocs.camera.simulator_sdk
```
but that's illegal syntax. Would have to start either manually fiddling with the `pocs.camera` namespace, or special casing the SDK camera simulator so it doesn't have to be in `pocs.camera`.  Neither option seems very appealing.

AnthonyHorton: What I could do is basically do what the DSLR simulator is currently doing. Move `ccd.py` out of `pocs/camera/simulator` into it's own directory and import the `Camera` class in the `__init__.py`. Means we get two simulated camera directories.

PR #912, opened 2019-10-20
Fix target temperature in simulator

wtgee:
## Description
The `target_temperature` was not properly being set in the `ccd` simulator.

This will set the `target_temperature` even if `is_cooled_camera=False` but will provide a warning.

On current `develop` running the following will fail because it relies on a previous test and the `module` level scope for the camera object.:

```bash
pytest -xvrs pocs/tests/test_camera.py -k 'test_exposure'
```

## Checklist:

### DIFF
--- a/pocs/camera/simulator_sdk/ccd.py
+++ b/pocs/camera/simulator_sdk/ccd.py
@@ -29,12 +29,12 @@ class Camera(AbstractSDKCamera, Camera):
                  driver=SDKDriver,
                  target_temperature=0 * u.Celsius,
                  *args, **kwargs):
-        kwargs['target_temperature'] = target_temperature
         super().__init__(name, driver, *args, **kwargs)
 
         self._is_cooled_camera = True
         self._cooling_enabled = False
         self._temperature = 25 * u.Celsius
+        self._target_temperature = target_temperature
         self._max_temp = 25 * u.Celsius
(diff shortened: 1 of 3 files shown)


pocs/camera/simulator_sdk/ccd.py
```diff
@@ -34,6 +34,7 @@ def __init__(self,
         self._is_cooled_camera = True
         self._cooling_enabled = False
         self._temperature = 25 * u.Celsius
+        self._target_temperature = target_temperature
```
wtgee: I tried to do a `kwargs.update({'target_temperature': target_temperature})` before the `super()` call but it's not working for some reason. I think that would be preferred so will try to look at how the inheritance is working later.

pocs/camera/simulator_sdk/ccd.py
```diff
@@ -57,7 +59,9 @@ def target_temperature(self):
 
     @target_temperature.setter
     def target_temperature(self, target):
-        self._last_temp = self.temperature
+        # Upon init the camera won't have an existing temperature.
+        with suppress(AttributeError):
+            self._last_temp = self.temperature
```
AnthonyHorton: Ah, bitten again by `super().__init__()` happening before all the initialisation specific to this class...

Issue #926, opened 2019-11-14
Making pocs script check if docker containers are already running


zacharyt20:
The ```pocs``` bash script that starts the docker containers should check if containers are already running. Currently, it only checks for this when ```bin/pocs up -d``` is used but doesn't when ```bin/pocs up``` is used.
 
**Describe the solution you'd like**
Use an if statement evaluating the output of ```docker ps``` to check if docker containers are already running and exit the script early with an error if they are.

wtgee:
Hi @zacharyt20, thanks for the Issue.

> Currently, it only checks for this when bin/pocs up -d is used but doesn't when bin/pocs up is used.

There shouldn't be any difference between the two so I'm curious as to what evidence you saw that there was a difference?

Note that `bin/pocs-cmd` does currently check if running, so the same code could probably work.

In general we should list some things that are desired for the start up, e.g. not start up the weather if not weather station, check for proper config, etc.

zacharyt20:
Hi @wtgee , 
I just re-tested the commands again on our unit and they are behaving the same way. Initially I didn't see the first output of ```bin/pocs up -d``` because of all the log messages.

Just by looking at the output that I had missed, ```bin/pocs up``` detects if containers have been started and gives a different text output, but still tries to load the containers. I was just looking in ```bin/pocs-cmd``` and it looks like it is meant to start the docker containers, but I don't see it called in ```bin/pocs```.  Is it called somewhere else and is it even connected to ```bin/pocs```?

I agree that giving the user the ability to specify which containers to start would be a great addition. I know the POCS shell uses a simulator system, would we need to simulate a sensor or simply not have it load? 

Also, I think that determining whether or not docker containers are running should be the first thing that happens in the script.

wtgee:
> I was just looking in `bin/pocs-cmd` and it looks like it is meant to start the docker containers, but I don't see it called in `bin/pocs`. Is it called somewhere else and is it even connected to `bin/pocs`?

It doesn't start the containers, it just checks if started or not here:
https://github.com/panoptes/POCS/blob/bfe33157431ec2c72ce7cd6d9193ff7b1884355f/bin/pocs-cmd#L6-L8

The idea with `pocs-cmd` is that it will run a command inside the container from outside the container.  This is what `docker exec` does. So, for instance, if you want to run `jupyter-console` inside the `pocs-shell` container, you call `bin/pocs-cmd jupyter-console` from _outside_ the container, that is, just from the regular command line.

[...]

PR #927, opened 2019-11-14
Bug Fix - Start Docker Containers

zacharyt20:
$POCS/bin/pocs tries to start docker containers when they are already running. Added an if statement that exits the code if docker containers are already running.  

## Description
Compares the docker ps command output with the pocs-shell container ID to see if the pocs-shell container is running. If it finds that the container is running it exits the script before attempting to start the docker containers and outputs an error with information on how to manually kill docker containers in case a single docker container failed.

## Related Issue
[926](https://github.com/panoptes/POCS/issues/926)

## How Has This Been Tested?
Killed all docker containers, then ran ```bin/pocs up -d``` from the ```$POCS``` directory to make sure containers still start, and then ran ```bin/pocs up -d``` again which behaved as expected and in the way described in the description section.
## Types of changes

## Checklist:

### DIFF
--- a/bin/pocs
+++ b/bin/pocs
@@ -25,6 +25,12 @@ usage() {
 
     # Run the software tests (no hardware)
     $POCS/scripts/pocs-docker.sh up
+    
+    #Start specific docker containers in the back ground with
+    $POCS/scripts/pocs-docker.sh up --no-deps -d <container name> 
+    
+    #Manually stop docker containers in the with
+    docker stop <container name> 
 "
 }
 
@@ -34,16 +40,24 @@ if [ "${START}" = 'help' ] || [ "${START}" = '-h' ] || [ "${START}" = '--help' ]
 	exit 1
 fi
 
-if [ "$(docker ps -q -f name="pocs-shell")" ] || [ "$(docker ps -q -f name="peas-shell")" ] || [ "$(docker ps -q -f name="messaging-hub")" ] || [ "$(docker ps -q -f name="config-server")" ] || [ "$(docker ps -q -f name="aag-weather-reader")" ] || [ "$(docker ps -q -f name="aag-weather-server")" ]; then
-	echo "At least one PANOPTES docker container is already running! Manually stop docker containers with docker stop <container name> if needed."
-	echo "If an single docker container needs to be started do so with bin/pocs up --no-deps <container name>."
-        exit 1
-fi
-
+PARAMS="$@"   
+    
 cd "$PANDIR"
-docker-compose \
-   --project-directory "${PANDIR}" \
-        -f panoptes-utils/docker/docker-compose.yaml \
-        -f PAWS/docker/docker-compose.yaml \
-        -f POCS/docker/docker-compose.yaml \
-                -p panoptes "$@"
+CMD="docker-compose \
+    --project-directory ${PANDIR} \
+    -f panoptes-utils/docker/docker-compose.yaml \
+    -f PAWS/docker/docker-compose.yaml \
+    -f POCS/docker/docker-compose.yaml \
+    -p panoptes"
+    
+# If user only asked to start, check if already running and if so use "-d" option.
+if [[ "$PARAMS" == "up" ]]; then
+    if [[ ! -z $(eval "${CMD} ps -q") ]]; then
+        echo "Some containers already running, using -d to only start non-running containers."
+	echo "For more info on managing docker containers manually, run bin/pocs."
+        PARAMS="up -d"    
+    fi    
+fi    
+    
+# Run the docker-compose command with user params.
+eval "${CMD} ${PARAMS}"
--- a/bin/pocs
+++ b/bin/pocs
@@ -34,31 +34,15 @@ if [ "${START}" = 'help' ] || [ "${START}" = '-h' ] || [ "${START}" = '--help' ]
 	exit 1
 fi
 
-#Array storing container names
-docker_names=("pocs-shell" "peas-shell" "messaging-hub" "config-server" "aag-weather-reader" "aag-weather-server")
+if [ "$(docker ps -q -f name="pocs-shell")" ] || [ "$(docker ps -q -f name="peas-shell")" ] || [ "$(docker ps -q -f name="messaging-hub")" ] || [ "$(docker ps -q -f name="config-server")" ] || [ "$(docker ps -q -f name="aag-weather-reader")" ] || [ "$(docker ps -q -f name="aag-weather-server")" ]; then
+	echo "At least one PANOPTES docker container is already running! Manually kill docker containers with docker kill <container name> if needed."
+        exit 1
+fi
 
-let counter=6
-#While loop that checks to see if any docker containers are running
-while [ "${counter}" > 0 ]
-do
-        element=$(($counter - 1))
-        DOCKER_NAME=${docker_names[$element]}
-	#If running
-        if [ "$(docker ps -q -f name=${DOCKER_NAME})" ]; then
-                echo "At least one docker container is already running! Manually kill docker containers with docker kill <container name> if needed."
-                exit 1
-	#if not running
-        else
-	#I placed this in an else statement instead of keeping it outside the while loop since I was running into an 
-	#error where the while loop exits the bash script and stops the containers from stopping.
-                cd /var/panoptes
-                docker-compose \
-                   --project-directory "${PANDIR}" \
-                        -f panoptes-utils/docker/docker-compose.yaml \
-                        -f PAWS/docker/docker-compose.yaml \
-                        -f POCS/docker/docker-compose.yaml \
-                        -p panoptes "$@"
-                exit
-        fi
-        let counter=$counter-1
-done
+cd "$PANDIR"
+docker-compose \
+   --project-directory "${PANDIR}" \
+        -f panoptes-utils/docker/docker-compose.yaml \
+        -f PAWS/docker/docker-compose.yaml \
+        -f POCS/docker/docker-compose.yaml \
+                -p panoptes "$@"
--- a/bin/pocs
+++ b/bin/pocs
@@ -52,7 +52,7 @@ CMD="docker-compose \
     
 # If user only asked to start, check if already running and if so use "-d" option.
 if [[ "$PARAMS" == "up" ]]; then
-    if [[ ! -z $(eval "${CMD} ps -q") ]]; then
+    if [[ ! -z $(eval "${CMD} top") ]]; then
         echo "Some containers already running, using -d to only start non-running containers."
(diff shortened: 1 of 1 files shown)


bin/pocs
```diff
@@ -34,6 +34,17 @@ if [ "${START}" = 'help' ] || [ "${START}" = '-h' ] || [ "${START}" = '--help' ]
 	exit 1
 fi
 
+#Patch could be improved by using for loops to test for all docker containers
+#in case the user manually kills pocs-shell. Also the if statement should be true
+#if the string doesn't have characters
+
+#exit script before trying to start containers if containers are running
+docker_ps=$(docker ps -q -f name=pocs-shell)
+if [ "$docker_ps" = "87d6fde2a139" ]; then
```
wtgee: This is hard-coded to a specific container name but is the right approach.  Basically if it returns anything other than nothing than it is running something (how's that for a vague statement :grin:). 

In `bin/pocs-cmd` we just check for an empty string via:

```bash
if [ ! "$(docker ps -q -f name=${DOCKER_NAME})" ]; then
```

That should be good enough here. If we wanted to force certain versions we could use a specific id otherwise best to be generic.

zacharyt20:
> Note that this assumes that POCS is the only container-based system running on the computer, which might be a bad assumption but I think is ok.

Using an if statement with multiple parameters avoids using a while loop and doesn't make this assumption.  

Also, I think it needs to check if each POCS container is running. If there is an issue in the future with docker or POCS that causes only a portion of the containers to stop(fail), the method of checking only one container would likely cause ```bin/pocs``` to start containers that are already running.

wtgee:
> Using an if statement with multiple parameters avoids using a while loop and doesn't make this assumption.
> 
> Also, I think it needs to check if each POCS container is running. If there is an issue in the future with docker or POCS that causes only a portion of the containers to stop(fail), the method of checking only one container would likely cause `bin/pocs` to start containers that are already running.

In this situation you would either want to start the individual container with `bin/pocs up --no-deps pocs-shell` or stop all containers and restart them, so I don't think that's an issue.

Also, just a note, I just checked this and running `bin/pocs up` does not restart existing containers, although it does attached the current terminal to the log output messages (unless run with `-d` of course).  This can make it awkward as then you would have two terminals running the same thing. But there shouldn't be any conflict in terms of trying to start resources again.

zacharyt20:
I think the average user would still attempt ```bin/pocs up -d``` to start the containers, even if they realize only certain containers are running. I could keep the same if statement and explain the `bin/pocs up --no-deps pocs-shell` syntax to the user in the error message so that way they don't have to do research on their own / ask on the forum.

wtgee:
> I could keep the same if statement and explain the bin/pocs up --no-deps pocs-shell syntax to the user in the error message so that way they don't have to do research on their own / ask on the forum.

This is already explained in the help for `bin/pocs`. So if a user runs `bin/pocs` or `bin/pocs -h` or `bin/pocs --help` it will show the explanation.

To be clear, the only issue is that if people run `bin/pocs up` in two different terminals then it is attached in two terminals, which is awkward but doesn't prevent any functionality. So we could either always force `-d`, or only force `-d` when *any* service is already running.

This latter solution looks like the following and is pretty clean.

```bash
#!/bin/bash    
 
# Get user params
PARAMS="$@"    
    
cd "$PANDIR" 
CMD="docker-compose \    
    --project-directory ${PANDIR} \    
    -f panoptes-utils/docker/docker-compose.yaml \    
    -f PAWS/docker/docker-compose.yaml \    
    -f POCS/docker/docker-compose.yaml \    
    -p panoptes"    
    
# If user only asked to start, check if already running and if so use "-d" option.
if [[ "$PARAMS" == "up" ]]; then    
    if [[ ! -z $(eval "${CMD} ps -q") ]]; then    
        echo "Some containers already running, using -d to only start non-running containers."    
        PARAMS="up -d"    
    fi    
fi    
    
# Run the docker-compose command with user params.
eval "${CMD} ${PARAMS}"
```

zacharyt20:
I was under the impression that `bin/pocs up` and `bin/pocs up -d` did the same thing with only difference being that the `-d` command runs in the background while `pocs up` runs in the foreground and shows in-depth debug messages. How/why would they be doing different things? From my understanding of `bin/pocs up`, it seems that adding the `-d` to `bin/pocs up` only serves to hide the ugly errors in the debug messages that are spat out when `bin/pocs up` is run with containers already started.

In the google docs documentation it says to use `bin/pocs up -d`. The only reason why someone would use `bin/pocs up` is to see the additional debug messages, which this solution hides and doesn't offer a way to see those messages or point the user towards the instructions in `bin/pocs`.

It's also again making the assumption that there are no other docker containers other than the ones made by ```pocs```.  This could be problematic for future developers trying to test new hardware who are using containers with the hopes of eventually integrating with the docker version of POCS.

wtgee:
> I was under the impression that `bin/pocs up` and `bin/pocs up -d` did the same thing with only difference being that the `-d` command runs in the background while `pocs up` runs in the foreground and shows in-depth debug messages. How/why would they be doing different things? From my understanding of `bin/pocs up`, it seems that adding the `-d` to `bin/pocs up` only serves to hide the ugly errors in the debug messages that are spat out when `bin/pocs up` is run with containers already started.

That's correct, they are doing the same thing.

> 
> In the google docs documentation it says to use `bin/pocs up -d`. The only reason why someone would use `bin/pocs up` is to see the additional debug messages, which this solution hides and doesn't offer a way to see those messages or point the user towards the instructions in `bin/pocs`.

I use `bin/pocs up` because I prefer to keep the logs running so I can be sure that I have started the containers. Therefore I've never actually had the problem of trying to run `bin/pocs up` twice (with or without the `-d`) because I can clearly see that it's still running.

Also, this solution does not in fact do as you mention. It only forces the `-d` option if there are containers already running, in which case you do not want to use just `up` by itself, which is the actual issue this PR is trying to solve.  If no containers are running you can use just `bin/pocs up` to view log message and attach to the terminal just fine.

As it also says in the `bin/pocs --help` message, you can view the logs at any time with `bin/pocs logs <service-name>`, whether attached or not.

> It's also again making the assumption that there are no other docker containers other than the ones made by `pocs`. This could be problematic for future developers trying to test new hardware who are using containers with the hopes of eventually integrating with the docker version of POCS.

This is using the `docker-compose` version of `ps`, not the `docker` version, so it only reports on the containers that are defined in the various `docker-compose.yaml` files that are specified. (You can get it to report all running containers with `-a` but that is not the default). Other stand-alone containers won't be reported and won't affect the operation. I tested this all before I posted the solution.

zacharyt20:
I have been assuming that the debug messages that I saw in terminal with `bin/pocs up` were also happening in the `bin/pocs up -d` log file (Note: I don't have the log file mentioned in the help section of `bin/pocs` on PAN015). I had tested `docker-compose ps` on our unit but since it only has containers within `$PANDIR` I didn't realize it filtered docker containers so cleanly!

I would also like to clarify that the log file does not show the errors that appear in `bin/pocs up` since they are not on my unit or their location is different. 

My only small concern is that someone completely new to bash/programming will try starting the containers for the first or second time and try to see the `bin/pocs up` debug messages in terminal, and won't know how to get that working. In the newest commit I've added another echo line pointing them toward the help section with docker commands.

bin/pocs
```diff
@@ -25,6 +25,12 @@ usage() {
 
     # Run the software tests (no hardware)
     $POCS/scripts/pocs-docker.sh up
+    
+    #Start specific docker containers in the back ground with
+    $POCS/scripts/pocs-docker.sh up --no-deps -d <container name>
```
wtgee: Should maintain consistency in what we call
```suggestion
    $POCS/bin/pocs up --no-deps -d <container name> 
```

bin/pocs
```diff
@@ -34,11 +40,24 @@ if [ "${START}" = 'help' ] || [ "${START}" = '-h' ] || [ "${START}" = '--help' ]
...
+    -f panoptes-utils/docker/docker-compose.yaml \
+    -f PAWS/docker/docker-compose.yaml \
+    -f POCS/docker/docker-compose.yaml \
+    -p panoptes"
+    
+# If user only asked to start, check if already running and if so use "-d" option.
+if [[ "$PARAMS" == "up" ]]; then
+    if [[ ! -z $(eval "${CMD} ps -q") ]]; then
+        echo "Some containers already running, using -d to only start non-running containers."
+	echo "For more info on managing docker containers manually, run bin/pocs."
```
wtgee:
```suggestion
	 echo "For more info on managing docker containers manually, run bin/pocs --help".
```

zacharyt20:
Thanks for helping me with my first PR!

zacharyt20:
I applied the changes to `bin/pocs` on our unit and it is not working. It looks like the output of `docker-compose` does not change if the docker containers are running or stopped. This causes the second if statement to always be true, and the user can never see `bin/pocs` in the foreground and gets the "Some containers are already running" message when there are no containers running.

bin/pocs
```diff
@@ -34,11 +40,24 @@ if [ "${START}" = 'help' ] || [ "${START}" = '-h' ] || [ "${START}" = '--help' ]
...
+CMD="docker-compose \
+    --project-directory ${PANDIR} \
+    -f panoptes-utils/docker/docker-compose.yaml \
+    -f PAWS/docker/docker-compose.yaml \
+    -f POCS/docker/docker-compose.yaml \
+    -p panoptes"
+    
+# If user only asked to start, check if already running and if so use "-d" option.
+if [[ "$PARAMS" == "up" ]]; then
+    if [[ ! -z $(eval "${CMD} top") ]]; then
```
wtgee: Why the change to `top` from `ps -q`? I think either will work, just curious.

zacharyt20: I replaced `ps -q` with `top` to filter for running processes. 

On our system, when we tried `bin/pocs up` with **no** containers running, the bash script added the `-d` option, which is not the intended behavior.  We echoed the parameter of the if statement (`$(eval "${CMD} ps -q")`), and saw that this parameter is the always the same, and does not change based on the status of the docker containers. My best guess as to why this happens is that `docker ps` lists running containers, while `docker-compose ps` lists any information it can get from the `docker-compose.yaml` files, which it can find when docker containers aren't running.

Using `top` lists only the running processes.

wtgee: Hm, that doesn't happen for me, `ps -q` will only give output. I guess maybe if you have some stopped machines? I tend to do `docker system prune` a lot as it cleans up a lot of hard drive space, but that would remove stopped and unused containers.

wtgee:
Will be closed by #951

wtgee:
As per comment above, I'm closing this now as part of my cleanup effort.

PR #930, opened 2019-11-29
State Machine - State Tags

wtgee:
* Add tags to the states
* Use the `always_safe` tag during the `check_safety` transition rather than using the transition name.
* Support the `horizon='flat'` attribute for the `at_twilight` tag, e.g. for a `calibrating` state.

## Description
This adds the ability to support `pytransitions` "tags" feature for states and then lets us use those tags for logic controller. Here we change the `check_safety` method so that it looks for an `always_safe` tag on the state name rather than relying on hard-coding the transition name. Also will call the `is_safe` with a `horizon='flat'` for `at_twilight` tags, with PANOPTES currently doesn't need but Huntsman will.

## Related Issue
https://github.com/AstroHuntsman/huntsman-pocs/issues/124

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,9 @@ install:
   - pip install -r requirements.txt
   - pip install -r docs/requirements.txt
   - pip install -e .
+  # Force creation of config file.
+  - python -c "import astroplan"
+  # Custom IERS urls in config file.
   - |
     echo "\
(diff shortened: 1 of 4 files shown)


wtgee:
Tests currently failing due to #929

PR #932, opened 2019-12-08
horizon=flat

danjampro:
-Allow kwargs to be passed to wait_until_safe so we can pass the horizon argument

-Passed horizon='flat' to wait_until_safe and is_safe in machine.run. I am not sure about this, as maybe the horizon used should depend on the next state?

### DIFF
--- a/pocs/state/machine.py
+++ b/pocs/state/machine.py
@@ -129,8 +129,8 @@ class PanStateMachine(Machine):
 
                 # If sleeping, wait until safe (or interrupt)
                 if self.state == 'sleeping':
-                    if self.is_safe() is not True:
-                        self.wait_until_safe()
+                    if self.is_safe(horizon='flat') is not True:
+                        self.wait_until_safe(horizon='flat')
 
                 try:
                     state_changed = self.goto_next_state()
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -587,13 +587,13 @@ class POCS(PanStateMachine, PanBase):
             # Sleep for a little bit.
             time.sleep(sleep_delay)
 
-    def wait_until_safe(self):
+    def wait_until_safe(self, **kwargs):
         """ Waits until weather is safe.
 
         This will wait until a True value is returned from the safety check,
         blocking until then.
         """
-        while not self.is_safe(no_warning=True):
+        while not self.is_safe(no_warning=True, **kwargs):
             self.sleep(delay=self._safe_delay)
 
 ##################################################################################################
--- a/pocs/utils/theskyx.py
+++ b/pocs/utils/theskyx.py
@@ -50,12 +50,14 @@ class TheSkyX(object):
 
     def read(self, timeout=5):
         try:
-            self.socket.settimeout(timeout)
+            #self.socket.settimeout(timeout)
             response = None
             err = None
 
             try:
-                response = self.socket.recv(2048).decode()
+                #response = self.socket.recv(2048).decode()
+                msg, ancdata, flags, addr = self.socket.recvmsg(1024)
+                response = msg.decode()
                 if '|' in response:
                     response, err = response.split('|')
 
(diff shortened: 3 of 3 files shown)


AnthonyHorton:
@danjampro Looks like you need to either merge from develop or rebase because you have some unrelated changes to `pocs/utils/theskyx.py` in the diff that appear to be causing the test failures.

danjampro:
@AnthonyHorton Not sure where that came from - I don't think they were my changes.... Anyway, I've updated the PR

pocs/state/machine.py
```diff
@@ -129,8 +129,8 @@ def run(self, exit_when_done=False, run_once=False):
 
                 # If sleeping, wait until safe (or interrupt)
                 if self.state == 'sleeping':
-                    if self.is_safe() is not True:
-                        self.wait_until_safe()
+                    if self.is_safe(horizon='flat') is not True:
+                        self.wait_until_safe(horizon='flat')
```
AnthonyHorton: You're replacing one hard coded horizon for exiting the `sleeping` state (-18 degrees) with another hard coded horizon (-6 degrees). That will fix the immediate problem for Huntsman, but might not be great for Panoptes. 

Ideal solution, long term, would be to make this choice of horizon a config file thing.

lspitler:
@danjampro looks like this is in Review mode. Maybe convert from WIP to Review.

pocs/state/machine.py
wtgee: PANOPTES wants to switch to a default of taking flat-fields as well, so this should be okay. We just need to make sure to actually add in a flat field state before this is merged.

danjampro:
> PANOPTES is moving toward assuming a flat-fielding state as well, so I might end up closing this PR for another one that incorporates the same changes as well as a flat-field state for PANOPTES.

Do you have a timeframe in mind?

wtgee:
As soon as possible. However, we don't actually have to wait to develop
that state, we could instead just add a separate check for the observe
horizon.

I  won't get to it tonight but could tomorrow or soon. Should be fairly
trivial.

wtgee:
@danjampro for now you should be able to change this to:

```python
                # Wait for different states as it gets darker
                if self.state == 'sleeping':
                    self.wait_until_safe(horizon='flat')
                elif self.state == 'focusing':
                    self.wait_until_safe(horizon='focus')
                elif self.state == 'scheduling':
                    self.wait_until_safe(horizon='observe')
```

wtgee:
Better yet, we could do a lookup like:

```python
from contextlib import suppress

...

                horizon_lookup = {
                    'sleeping': 'flat',
                    'focusing': 'focus',
                    'scheduling': 'observe',
                }

                # Wait for horizon level if state requires.
                with suppress(KeyError):
                    self.wait_until_safe(horizon=horizon_lookup[self.state])
```

Then in the future (or now if you want) can alter it so that we are populating `horizon_lookup` from the state file. A state entry (in the state table yaml file) could be:

```yaml
    -
        source: scheduling
        dest: slewing
        trigger: start_slewing
        horizon: observe
```

wtgee:
Actually, here's a patch. I basically had it done by explaining it.

You can `git apply horizon-patch.txt`.

[horizon-patch.txt](https://github.com/panoptes/POCS/files/4055476/horizon-patch.txt)

danjampro:
> Actually, here's a patch. I basically had it done by explaining it.
> 
> You can `git apply horizon-patch.txt`.
> 
> [horizon-patch.txt](https://github.com/panoptes/POCS/files/4055476/horizon-patch.txt)
> 
> This patch is slightly different from the above yaml example. That was putting it in the transitions section and I have just moved it to the actual states section.

Great, thanks! You you recommend adding the patch to this PR?

wtgee:
Closing in favor of #936

Issue #933, opened 2019-12-11
POCS-SHELL: reset_pocs bypasses exit's safety functions


zacharyt20:
*In the docker version of POCS, the `pocs-shell` can be exited quickly by using `reset_pocs` before entering `exit`. If the user accidentally types `reset_pocs` while the mount is in the home position, the mount does not automatically park itself, and does not warn the user. This is dangerous as the camera's can be easily damaged if left in the home position during the day. If a unit isn't using an external webcam to monitor the unit, this would likely go unnoticed.*

**Solution**
Warn the user if they try running `reset_pocs` while the mount is not in the park position, and do not execute `reset_pocs`. Allow the user to enter `reset_pocs` again (twice in a row) to reset the pocs instance from anywhere.

**Alternatives**
*Have reset_pocs park the unit before resetting the pocs instance.*
I think this shouldn't be implemented because the purpose of `reset_pocs` is to allow the user to throw away a dysfunctional pocs instance for debugging or troubleshooting purposes. I don't think it would be a good idea to add a possibility that `reset_pocs` fails because the park function throws an error.

**Home and Park Position**
Home position: Head unit is up right and cameras are pointed toward polaris(sky)
Park position: Cameras are facing down towards the ground and the RA axis is horizontal to the ground. It is the safe position for the telescope, where the cameras can't be damaged.

**Reproduce**
Enter the pocs-shell and setup pocs.
Unpark the mount and get it into the home position through `go_home`. (Note: We first found this issue after running the `polar_alignment_test`
Enter `reset_pocs`.

wtgee:
Will be closed by #951

wtgee:
Closed by #951. Please re-open if necessary after testing.

PR #934, opened 2019-12-11
Config file updates and better user warnings in the pocs-shell

zacharyt20:
## Description
Added endpoint URL for AGG weather reader in pocs.yaml.

Added if statements and an input prompt to pocs-shell.py so that users don't discard a POCS instance with `reset_pocs` while the mount is in the home position without knowing they have to manually park it after discarding the instance.

## Related Issue
[933](https://github.com/panoptes/POCS/issues/933)

## How Has This Been Tested?

Config file changes that added the AAG weather sensor endpoint URL are working on PAN015. 

The `pocs-shell` changes in `reset_pocs` have been tested on PAN015. When parked, `reset_pocs` behavior does not changed. When not parked, the user is warned about leaving the unit in the home position and is prompted if they would like to run `reset_pocs` with a yes/no response.  If input is not a yes/no response the `reset_pocs` method is called again and the users is prompted with the same question. 

## Types of changes

## Checklist:

### DIFF
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -91,14 +91,14 @@ observations:
 # By default all images are stored on googlecloud servers and we also
 # use a few google services to store metadata, communicate with servers, etc.
 #
-# See $POCS/pocs/utils/google/README.md for details about authentication.
+# See $PANDIR/panoptes/utils/google/README.md for details about authentication.
 #
 # Options to change:
 #   image_storage: If images should be uploaded to Google Cloud Storage.
 #   service_account_key: Location of the JSON service account key.
 ################################################################################
 panoptes_network:
-    image_storage: True
+    image_storage: False
     service_account_key: # Location of JSON account key
     project_id: panoptes-survey
     buckets:
--- a/conf_files/pocs.yaml
+++ b/conf_files/pocs.yaml
@@ -15,13 +15,13 @@ pan_id: PAN000
 
 location:
     name: Mauna Loa Observatory
-    latitude: 19.54 deg
-    longitude: -155.58 deg
-    elevation: 3400.0 m
-    horizon: 30 deg # targets must be above this to be considered valid.
-    flat_horizon: -6 deg # Flats when sun between this and focus horizon.
-    focus_horizon: -12 deg # Dark enough to focus on stars.
-    observe_horizon: -18 deg # Sun below this limit to observe.
+    latitude: 19.54 deg # Degrees
+    longitude: -155.58 deg # Degrees
+    elevation: 3400.0 m # Meters
+    horizon: 30 deg # Degrees; targets must be above this to be considered valid.
+    flat_horizon: -6 deg # Degrees - Flats when sun between this and focus horizon.
+    focus_horizon: -12 deg # Degrees - Dark enough to focus on stars.
+    observe_horizon: -18 deg # Degrees - Sun below this limit to observe.
     timezone: US/Hawaii
     gmt_offset: -600 # Offset in minutes from GMT during.
                      # standard time (not daylight saving).
@@ -91,14 +91,14 @@ observations:
 # By default all images are stored on googlecloud servers and we also
 # use a few google services to store metadata, communicate with servers, etc.
 #
-# See $PANDIR/panoptes/utils/google/README.md for details about authentication.
+# See $POCS/pocs/utils/google/README.md for details about authentication.
 #
 # Options to change:
 #   image_storage: If images should be uploaded to Google Cloud Storage.
 #   service_account_key: Location of the JSON service account key.
 ################################################################################
 panoptes_network:
-    image_storage: False
+    image_storage: True
     service_account_key: # Location of JSON account key
     project_id: panoptes-survey
     buckets:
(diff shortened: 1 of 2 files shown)


conf_files/pocs.yaml
```diff
@@ -91,14 +91,14 @@ observations:
 # By default all images are stored on googlecloud servers and we also
 # use a few google services to store metadata, communicate with servers, etc.
 #
-# See $PANDIR/panoptes/utils/google/README.md for details about authentication.
+# See $POCS/pocs/utils/google/README.md for details about authentication.
```
wtgee: This shouldn't be changed, the readme comes from the `panoptes-utils` repository.

conf_files/pocs.yaml
```diff
@@ -91,14 +91,14 @@ observations:
...
 #
 # Options to change:
 #   image_storage: If images should be uploaded to Google Cloud Storage.
 #   service_account_key: Location of the JSON service account key.
 ################################################################################
 panoptes_network:
-    image_storage: False
+    image_storage: True
```
wtgee: We don't want to default this to `True` for everyone. We will have people change it to `True` once they have a valid service account key and so they actually have the ability to upload.

scripts/pocs-shell.py
```diff
@@ -190,7 +190,17 @@ def do_reset_pocs(self, *arg):
 
         Does NOT park the mount, nor execute power_down.
         """
-        self.pocs = None
+        if self.pocs.observatory.mount.is_parked is False:
+            print("ATTENTION: The mount is not in the parked position and reset_pocs will not move the mount to park position!")
```
wtgee: There are some convenience functions at the bottom that can print in different colors.

```suggestion
            print_warning("ATTENTION: The mount is not in the parked position and reset_pocs will not move the mount to park position!")
```

scripts/pocs-shell.py
```diff
@@ -190,7 +190,17 @@ def do_reset_pocs(self, *arg):
...
+            response = input("The unit could be damaged if it is not parked before sunrise. Execute reset_pocs anyway?[y/n]")
+            if response == "y" or response == "yes":
```
wtgee: A slightly cleaner way to do something like this is to force their input into a format that is easier to check:

```suggestion
            if response.lower().startswith('y'):
```

scripts/pocs-shell.py
```diff
@@ -190,7 +190,17 @@ def do_reset_pocs(self, *arg):
...
+                self.pocs = None
+                print("POCS instance discarded. Remember to park the mount later.")
```
wtgee:
```suggestion
                print_info("POCS instance discarded. Remember to park the mount later.")
```

scripts/pocs-shell.py
```diff
@@ -190,7 +190,17 @@ def do_reset_pocs(self, *arg):
...
+            elif response == "n" or response == "no":
```
wtgee:
```suggestion
            elif response.lower().startswith('n'):
```

scripts/pocs-shell.py
```diff
@@ -190,7 +190,17 @@ def do_reset_pocs(self, *arg):
...
+                print("Keeping POCS instance.")
+            else:
+                print("A 'yes' or 'no' response is required")
+                self.do_reset_pocs()
```
wtgee: It's a little weird to recursively call this and should probably be avoided. It's slightly better to do something like:

```python
while True:
    response = input(...)
    if response.lower().starstswith('y'):
        # print message here
        break # break of out while loop
    elif...
```

scripts/pocs-shell.py
```diff
@@ -190,7 +190,24 @@ def do_reset_pocs(self, *arg):
...
+                response = input("The unit could be damaged if it is not parked before sunrise. Execute reset_pocs anyway[y/n]?")
+                if response.lower().startswith('y'):
+                    self.pocs = None
+                    print_info("POCS instance discarded. Remember to park the mount later.")
+                    break
+                elif response.lower().startswith('n'):
+                    print_info("Keeping POCS instance.")
+                    break
+                else:
+                    print("A 'yes' or 'no' response is required")
```
wtgee: POCS (mostly the underlying `Observatory`) can support a mode where you don't actually have a mount, which is a bit odd but could be used for things like testing cameras. I could see wanting to reset pocs (especially) in one of those scenarios. So this might be a bit better as:

```python
if self.pocs is None:
    print_info("POCS does not have an instance. Create one using 'setup_pocs'.")
    return
elif self.pocs.observatory.mount and not self.pocs.observatory.mount.is_parked:
    print_warning("ATTENTION: The mount is not in the parked position and reset_pocs will not move the mount to park position!")
    while True:
        response = input("The unit could be damaged if it is not parked before sunrise. Execute reset_pocs anyway [y/n]?")
        if response.lower().startswith('y'):
            print_info("POCS instance discarded. Remember to park the mount later.")
            break
        elif response.lower().startswith('n'):
            print_info("Keeping POCS instance.")
            return
        else:
            print("A 'yes' or 'no' response is required")

# Reset in all other scenarios
self.pocs = None
```

wtgee:
Hi @zacharyt20 , sorry for delay. The tests have been hanging on this for a while, which is an annoying thing with our Travis CI testing that we can't see what is going wrong.

I got some develop permissions from travis yesterday so that I can get access to the logs when this kind of thing happens, so I am going to get to debugging this soon.  Thanks!

wtgee:
Will be closed by #951

wtgee:
As per comment above, I'm closing this now as part of my cleanup effort. Thanks @zacharyt20 !

PR #935, opened 2019-12-16
ensure exptime Quantity doesn't impact the Huntsamn-pocs pyro client serialiser

lspitler:
A tiny fix to bug found from Huntsman-POCS and pyro 

## Description
Details are given here, including a fix discussion with @AnthonyHorton : 
https://github.com/AstroHuntsman/huntsman-pocs/issues/105#issuecomment-565915906
This PR prompted another solution as this one didn't work. See below.

## Related Issue
https://github.com/AstroHuntsman/huntsman-pocs/issues/105

## How Has This Been Tested?
On live control computer.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x ] My code follows the code style of this project.

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -700,7 +700,7 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         }
         metadata.update(headers)
 
-        exptime = kwargs.pop('exptime', observation.exptime.value)
+        exptime = kwargs.get('exptime', observation.exptime.value)
         # The exptime header data is set as part of observation but can
         # be override by passed parameter so update here.
         metadata['exptime'] = exptime
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -300,6 +300,8 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
                                                                          filename,
                                                                          **kwargs)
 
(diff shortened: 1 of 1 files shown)


lspitler:
@AnthonyHorton after another test, I think it should be:
```python
        exptime = get_quantity_value(kwargs.pop(
            'exptime', observation.exptime.value), unit=u.second)
```

lspitler:
OK after more  testing on the control computer, it was clear the solution was more subtle than initially anticipated.

Eventually realised that pop'ing from `**kwargs` inside a function doesn't impact the `**kwargs` outside of that function, e.g.:
image (2)

This one works and can be merged.

pocs/camera/camera.py
```diff
@@ -300,6 +300,8 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
                                                                          filename,
                                                                          **kwargs)
 
+        # pop exptime from kwarg as its now in exptime
+        exptime = kwargs.pop('exptime', observation.exptime.value)
```
wtgee: Sorry for being so late to the game. `exptime` here is a keyword argument that can be used to override the `observation.exptime.value` and `take_observation` is the only consumer of that, so it probably just shouldn't be hidden away in `kwargs` but as an actual documented parameter.

PR #936, opened 2020-01-17
Set state horizon level

wtgee:
Allow states to specify the required horizon level. If the state requires a specific horizon then the `run` method will sleep until the sun is at that horizon. The horizon level is specified in the state yaml configuration.

## Related Issue
Replaces #932


wtgee:
I don't like that the coverage is down. Ideally this would have a test for it.

pocs/core.py
```diff
@@ -478,7 +481,7 @@ def has_ac_power(self, stale=90):
 # Convenience Methods
 ##################################################################################################
 
-    def sleep(self, delay=2.5, with_status=True):
+    def sleep(self, delay=2.5, with_status=True, **kwargs):
```
danjampro: Does this **kwargs do anything?

wtgee: Yes, in the sense that it allows `sleep` to ignore other kwargs it doesn't care about. On line 593 I want to pass a series of kwargs to `wait_until_safe`, which in turn passes those to _both_ `is_safe` and `sleep`. If `is_safe` and `sleep` receive a kwarg they don't recognize they will throw an error. I could pull the sepcific parameters out of the `kwargs` in `wait_until_safe` and then pass them to `is_safe` and `sleep` respectively. But I'm lazy.

Really though this is just an effective way to say, "take the parameters you need and ignore all the others". I believe @jamessynge and I had debates about this a long time ago, as I used to include `*args, **kwargs` automatically on every method (although note that this is a different usage from subclassing).

wtgee:
> LGTM. Are any modifications required within Huntsman-POCS, other than in the state yaml config?

@danjampro I split out some of the changes to #937 (which I then merged). That requires an update to the requirements (via `pip install -Ur requirements.txt`). It also changes a bit how we do our requirements although that doesn't immediately affect Huntsman-POCS.

That would be in addition to the state yaml config, correct.

Issue #901, opened 2019-10-03
Add support for ZWO filterwheels


AnthonyHorton:
*<Enter description of change or enhancement here>*
Add support for the ZWO EFW/EFWmini filterwheels: https://astronomy-imaging-camera.com/product-category/filterfilter-wheel

**Describe the solution you'd like**
Implement a `pocs.filterwheel.zwo.FilterWheel` class based using the EFW SDK provided by ZWO.

**Related Issues/PRs**
Issue #768 and related PRs introduced filterwheel support to POCS, and added support for filterwheels from Diffraction Limited (SBIG).

AnthonyHorton:
Closed by #768

PR #940, opened 2020-02-03
Add support for filterwheels from ZWO

AnthonyHorton:
Adds support for filterwheels from ZWO

## Description

Add support for the ZWO EFW/EFWmini filterwheels: https://astronomy-imaging-camera.com/product-category/filterfilter-wheel

## Related Issue
#901

## How Has This Been Tested?
Testing functionality on actual hardware.

## Types of changes

## Checklist:


pocs/utils/library.py
```diff
@@ -14,6 +14,10 @@ def load_library(name, path=None, logger=None):
     Args:
         name (str): name of the library (without 'lib' prefix or any suffixes, e.g. 'fli').
         path (str, optional): path to the library e.g. '/usr/local/lib/libfli.so'.
+        mode (int, optional): mode in which to load the library, see dlopen(3) man page for
+            details. Should be one of ctypes.RTLD_GLOBAL, ctypes.RTLD_LOCAL, or
+            ctypes.DEFAULT_MODE. Default is ctypes.DEFAULT_MODE.
+        logger (logging.Logger, optional): logger to use.
```
wtgee: What happens if `mode=None`?

AnthonyHorton: If you explicitly set it to `None` you get something like
```
In [7]: libc = ctypes.CDLL("libc.so.6", mode=None)                                                                                     
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-9f402658a1ee> in <module>
----> 1 libc = ctypes.CDLL("libc.so.6", mode=None)

~/miniconda3/envs/huntsman/lib/python3.8/ctypes/__init__.py in __init__(self, name, mode, handle, use_errno, use_last_error, winmode)
    371 
    372         if handle is None:
--> 373             self._handle = _dlopen(self._name, mode)
    374         else:
    375             self._handle = handle

TypeError: an integer is required (got type NoneType)
```

If you simply don't set it it defaults to `ctypes.DEFAULT_MODE`, which is equivalent to `RTLD_LOCAL` in Linux and `RTLD_GLOBAL` in OS X.

pocs/filterwheel/libefw.py
```diff
@@ -0,0 +1,181 @@
...
+import enum
+
+from pocs.camera.sdk import AbstractSDKDriver
+from pocs.utils import error
+from pocs.utils.library import load_library
+
+
+class EFWDriver(AbstractSDKDriver):
+    # Because ZWO EFW library isn't linked properly have to manually load libudev
+    # in global mode first, otherwise get undefined symbol errors.
```
wtgee: Is that something on the fault of ZWO that might have a link we could include here for more information?

pocs/filterwheel/libefw.py
```diff
@@ -0,0 +1,181 @@
...
+        """Main class representing the ZWO EFW library interface.
+
+        On construction loads the shared object/dynamically linked version of the EFW SDK library,
+        which must be already installed (see https://astronomy-imaging-camera.com/software-drivers).
+
+        The name and location of the shared library can be manually specified with the library_path
+        argument, otherwise the ctypes.util.find_library function will be used to try to locate it.
+
+        Args:
+            library_path (str, optional): path to the libary e.g. '/usr/local/lib/libEFWFilter.so'
```
wtgee:
```suggestion
            library_path (str, optional): path to the library e.g. '/usr/local/lib/libEFWFilter.so'
```

Lots of' those `libary` copy/paste errors. :)

pocs/filterwheel/libefw.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_devices(self):
+        """Get connected device UIDs and corresponding device nodes/handles/IDs.
+        """
+        # EFW SDK has no way to access any unique identifier for connected filterwheels.
+        # Construct an ID (probably not deterministic, in general) from combination of
+        # product code and filterwheel ID.
+        n_filterwheels = self.get_num()  # Nothing works if you don't call this first.
+        products = self.get_product_ids()  # Will raise error.NotFound if no filterwheels
+        ids = [self.get_ID(i) for i in range(n_filterwheels)]
+        filterwheels = {f"{product}_{id}": id for product, id in zip(products, ids)}
```
wtgee:
```suggestion
        filterwheels = {f"{product}_{fw_id}": fw_id for product, fw_id in zip(products, ids)}
```

`id` is a python bulit-in so probably best to avoid using it even in situations where it probably won't matter.

pocs/filterwheel/libefw.py
AnthonyHorton: ISTR there being discussion of the undefined udev symbols problem on the ZWO forums, and me getting the workaround from there. Good call about including a link, I'll see if I can find the relevant thread again.

AnthonyHorton: OK, don't seem to be able to find it in ZWO forums. Not sure how I came up with the workaround.

pocs/utils/library.py
AnthonyHorton: Hmmm. I should probably fix this. `pocs.utils.library.load_library("c")` works just fine, loading `libc` with the default mode, but `pocs.utils.library.load_library("c", mode=None)` fails. I think I should just have this function replace a `mode` of `None` with `ctypes.DEFAULT_MODE`.

pocs/filterwheel/filterwheel.py
```diff
@@ -137,6 +140,10 @@ def current_filter(self):
     def current_filter(self, filter_name):
         self.move_to(filter_name, blocking=True)
 
+    @property
+    def is_unidirectional(self):
+        return self._unidirectional
```
wtgee: I don't see `._unidirectional` defined anywhere.

pocs/filterwheel/filterwheel.py
```diff
@@ -176,26 +183,31 @@ def move_to(self, position, blocking=False):
...
         assert self.is_connected, self.logger.error("Filter wheel must be connected to move")
+
         if self.camera and self.camera.is_exposing:
             msg = f'Attempt to move filter wheel {self} while camera is exposing, ignoring.'
             raise error.PanError(msg)
 
+        if self.is_moving:
+            msg = f'Attempt to move filter wheel {self} while already moving. Waiting.'
+            self.logger.warning(msg)
+            self._move_event.wait()
```
wtgee: I think a default timeout is a good idea for any hardware movement, even if really long. In situations where the hardware gets disconnected while in action then these loops can go forever.

pocs/filterwheel/filterwheel.py
```diff
@@ -176,26 +183,31 @@ def move_to(self, position, blocking=False):
...
+            # Already at requested position, don't go nowhere.
+            return self._move_event
 
-        self._move_to(position, move_event)  # Private method to actually perform the move.
+        self._move_event.clear()
+        self._move_to(position)  # Private method to actually perform the move.
 
         if blocking:
-            move_event.wait()
+            self._move_event.wait()
```
wtgee: Ditto on the timeout.

pocs/filterwheel/libefw.py
wtgee: No worries. I'll leave it up to you whether or not you think it a good enough idea to merit creating an issue out of it for the future in case it's found.

pocs/filterwheel/libefw.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        n_filterwheels = self.get_num()  # Nothing works if you don't call this first.
+        filterwheels = {}
+        for i in range(n_filterwheels):
+            fw_id = self.get_ID(i)
+            self.open(fw_id)
+            info = self.get_property(fw_id)
+            self.close(fw_id)
+            filterwheels[f"{info['name']}_{info['slot_num']}_{fw_id}"] = fw_id
+        return filterwheels
```
wtgee: Some amount of exception handling seems warranted. And debug logging.

pocs/filterwheel/libefw.py
```diff
@@ -0,0 +1,281 @@
...
+    def get_num(self):
+        """Get the count of connected EFW filterwheels."""
+        count = self._CDLL.EFWGetNum()
+        self.logger.debug(f"Found {count} connected EFW filterwheels.")
+        return count
+
+    def get_product_ids(self):
+        """Get product IDs of supported(?) EFW filterwheels.
+
+        The SDK documentation does not explain that the product IDs returned by this function are,
```
wtgee:
```suggestion
        The SDK documentation does not explain what the product IDs returned by this function are,
```

pocs/filterwheel/libefw.py
```diff
@@ -0,0 +1,281 @@
...
+        but from experiment and analogy with a similar function in the ASI camera SDK it appears
+        this is a list of the product IDs of the filterwheels that the SDK supports, not the
+        product IDs of the connected filterwheels. There appears to be no way to obtain the
+        product IDs of the connected filterwheel(s).
+        """
+        n_filterwheels = self._CDLL.EFWGetProductIDs(0)
+        if n_filterwheels > 0:
+            product_ids = (ctypes.c_int * n_filterwheels)()
```
wtgee: A function call here?

pocs/filterwheel/sbig.py
```diff
@@ -44,6 +44,9 @@ def __init__(self,
                          timeout=timeout,
                          serial_number=serial_number,
                          *args, **kwargs)
+        # All SBIG filterwheels unidirectional?
+        self._unidirectional = True
```
wtgee: Maybe an init or kwarg param.

pocs/filterwheel/zwo.py
```diff
@@ -0,0 +1,173 @@
...
+        return not self._move_event.is_set()
+
+    @property
+    def is_unidirectional(self):
+        return self._driver.get_direction(self._handle)
+
+    # ZWO filterwheels can be set to be unidirectional or bidirectional.
+    @is_unidirectional.setter
+    def is_unidirectional(self, unidirectional):
+        self._driver.set_direction(self._handle, bool(unidirectional))
```
wtgee: I don't see `set_direction`

pocs/filterwheel/zwo.py
```diff
@@ -0,0 +1,173 @@
...
+        if self._device_name:
+            self.logger.debug(f"Looking for filterwheel with device name '{self._device_name}'.")
+            if self._device_name not in FilterWheel._filterwheels:
+                msg = f"No filterwheel with device name '{self._device_name}'."
+                self.logger.error(msg)
+                raise error.NotFound(msg)
+            if self._device_name in FilterWheel._assigned_filterwheels:
+                msg = f"Filterwheel '{self._device_name}' already in use."
+                self.logger.error(msg)
+                raise error.PanError(msg)
```
wtgee: Should this (and below) really raise an error? Would be nice if `connect` where idempotent.

pocs/filterwheel/filterwheel.py
AnthonyHorton: Yeah, I haven't set it in the abstract base class, but I have given it a value in all the derived classes, either in `__init__()` or `connect()`. Maybe better to have the method just raise `NotImplementedError` here, to make it more obvious that derived classes need to do something?

pocs/filterwheel/filterwheel.py
AnthonyHorton: I'm inclined to change this so that is raises an error instead of waiting. Previously made exactly the same change with camera exposures.

pocs/filterwheel/filterwheel.py
AnthonyHorton: There are timeouts in the moves all the derived classes, but they're buried fairly deep down. In the simulator it's in the `_move_to()` method, whereas in the SBIG and ZWO filterwheels they're even more hidden because they're in the movement methods of the driver class. In each case there is a polling thread that includes a timer, and a `finally` clause that should make sure the event always gets set.

Ideally I'd refactor this like I did with the cameras `take_exposure()`, breaking the implementation up into several private methods, with most of it in the base class and only a small amount of hardware specific code in the derived classes. A job for another PR though, I think.

pocs/filterwheel/libefw.py
AnthonyHorton: Yes. `ctypes` can allow for some some funny looking syntax, sometimes.

Here I am creating an array data type on the fly by multiplying a scalar data type (`ctypes.c_int`) by an integer (`n_filterwheels`). That's actually the recommended method: https://docs.python.org/3/library/ctypes.html#arrays. I then immediately create an instance of that array data type by calling it.

I could make this look a bit less weird by creating the array data type and assigning it to a variable on one line, then creating the instance and assigning that to another variable on the next line. I only use the data type once though, so it didn't seem worth storing.

pocs/filterwheel/libefw.py
AnthonyHorton: A fair bit of debug logging happens in this method from the other methods that get called, but I've added some checks, error logging, etc.

pocs/filterwheel/sbig.py
AnthonyHorton: I've removed this entirely from the SBIG filterwheels now.

pocs/filterwheel/zwo.py
AnthonyHorton: It's a method of the `pocs.filterwheel.libefw.EFWDriver` class, which `self._driver` is an instance of.

pocs/filterwheel/zwo.py
AnthonyHorton: I think so. This behaviour is the same as most of the other hardware classes (SBIG, FLI & ZWO cameras, & Birger focusers).  The intent is to prevent a situation where the wrong device is being controlled, and I think its reasonable to raise an error if seems like that's about to occur.

One thing I have done here is add at check at the start of `connect()` to log a warning and skip all this if the filterwheel is already 'connected', as an extra attempt to connect shouldn't really result in an error.

Issue #941, opened 2020-02-03
prepare POCS for a filterwheel mode


lspitler:
*Modify POCS to accommodate Filterwheels.*

**Describe the solution you'd like**
Here are the tasks:

- [ ] when stepping through e.g. `target.yaml` each target can have an optional `filter_name` key/name that specifies the desired filter for a target
- [ ] Modify `target.yaml` parser to accommodate the `filter_name` key.
- [ ] Modify Observation class to add an optional filter_name arg to `scheduler.observation.Observation`.
- [ ] In `camera.take_observation`, if `filter_name` was specified and if `filter_wheel_exists=True`, then run `camera.move_to(filter_name)`

- [ ] In `filter_wheel.config`, for each filterwheel require a filter_wheel configure to have a `default_filter_name` taht will be used if someone doesn't speicify `filter_name` in `targets.yaml`
- [ ] Modify the camera initialisation so it sets up a filterwheel.

**Related Issues/PRs**
If there are any related Issues or PRs list them here:

danjampro:
"Modify target.yaml parser to accommodate the filter_name key."

I believe this is done automatically in `pocs.scheduler.scheduler.add_observation`

Also by targets.yaml, we are referring to the value of `fields_file` in the config

AnthonyHorton:
> Modify the camera initialisation so it sets up a filterwheel.

Not necessary, that code has been there a while. There just needs to be a filterwheel configuration in the camera config.

wtgee:
@danjampro and @AnthonyHorton how is this Issue in relation to the work that has been done?

PR #942, opened 2020-02-03
Add filter selection to observations

danjampro:
## Description
Add's an optional argument to `pocs.scheduler.observation.Observation` called `filter_name` that can be used to specify the filter for that observation, provided a filter wheel is found. Added a call to `filterwheel.move_to` in `AbstractCamera.take_observation` to facilitate this.

## Related Issue
#941 

## Types of changes

### DIFF
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -353,7 +353,7 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
 
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
-        # Check that the filterwheel is not moving
+        # Check that the camera (and subcomponents) is ready
         if not self.is_ready:
             msg = f"Attempt to start exposure on {self} while not ready, ignoring!"
             raise error.PanError(msg)
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -659,7 +659,8 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
 
             else:
                 self.logger.info(f'Filter {observation.filter_name} requested by'
-                    ' observation but {self} has no filterwheel, using {self.filter_type}.')
+                                 ' observation but {self} has no filterwheel, using'
+                                 ' {self.filter_type}.')
 
         start_time = headers.get('start_time', current_time(flatten=True))
 
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -354,14 +354,6 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
         # Check that the filterwheel is not moving
-        """
-        if self.filterwheel is not None:
-            if self.filterwheel.is_moving:
-                msg = f'Attempt to prepare observation on {self} while' \
-                        ' filterwheel is moving, ignoring.'
-                self.logger.error(msg)
-                raise error.PanError(msg)
-        """
         if not self.is_ready:
             msg = f"Attempt to start exposure on {self} while not ready, ignoring!"
             raise error.PanError(msg)
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -715,7 +715,11 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         if (self.filterwheel is not None) and (observation.filter_name is not None):
 
             # Move filterwheel if necessary
-            self.filterwheel.move_to(observation.filter_name, blocking=True)
+            try:
+                self.filterwheel.move_to(observation.filter_name, blocking=True)
+            except Exception as e:
+                self.logger.error(f'Error moving filterwheel to {observation.filter_name}: {e}')
+                raise(e)
 
             # Store the filter name in metadata
             metadata['filter_name'] = observation.filter_name
--- a/pocs/camera/camera.py
+++ b/pocs/camera/camera.py
@@ -302,7 +302,10 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
 
         # pop exptime from kwarg as its now in exptime
         exptime = kwargs.pop('exptime', observation.exptime.value)
-        exposure_event = self.take_exposure(seconds=exptime, filename=file_path, **kwargs)
+        
+        exposure_event = self.take_exposure(seconds=exptime, filename=file_path,
+                                            filter_name=observation.filter_name,
+                                            **kwargs)
 
         # Add most recent exposure to list
         if self.is_primary:
@@ -326,6 +329,7 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
                       filename=None,
                       dark=False,
                       blocking=False,
+                      filter_name=None,
                       *args,
                       **kwargs):
         """Take an exposure for given number of seconds and saves to provided filename.
@@ -350,11 +354,6 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
 
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
-        if self.filterwheel and self.filterwheel.is_moving:
-            msg = "Attempt to start exposure on {} while filterwheel is moving, ignoring.".format(
-                self)
-            raise error.PanError(msg)
-
         if not isinstance(seconds, u.Quantity):
             seconds = seconds * u.second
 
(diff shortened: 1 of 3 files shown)


pocs/camera/camera.py
```diff
@@ -302,7 +302,13 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
 
         # pop exptime from kwarg as its now in exptime
         exptime = kwargs.pop('exptime', observation.exptime.value)
-        exposure_event = self.take_exposure(seconds=exptime, filename=file_path, **kwargs)
+        
+        # Now move the filerwheel into position
+        if (self.filterwheel is not None) & (observation.filter_name is not None):
```
lspitler:
```suggestion
        if (self.filterwheel is not None) and (observation.filter_name is not None):
```

pocs/scheduler/observation.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, field, exptime=120 * u.second, min_nexp=60,
                 (default: {10})
             priority {int} -- Overall priority for field, with 1.0 being highest
                 (default: {100})
+            filter_name {str} -- Name of the filter to be used. If specified,
+            will override the default filter name (default: {None}).
```
lspitler:
```suggestion
            will override the default filter name.
```

pocs/camera/camera.py
danjampro: @lspitler  O.K. but why?

pocs/scheduler/observation.py
danjampro: @lspitler I kept to the style of the previous argument descriptions. What is the reason for changing this?

lspitler: Oh I see, nm, rejecting this suggestion.

wtgee:
```suggestion
                will override the default filter name (default: {None}).
```

pocs/camera/camera.py
```diff
@@ -302,7 +302,13 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
 
         # pop exptime from kwarg as its now in exptime
         exptime = kwargs.pop('exptime', observation.exptime.value)
-        exposure_event = self.take_exposure(seconds=exptime, filename=file_path, **kwargs)
+
+        # Move the filerwheel into position
+        if (self.filterwheel is not None) and (observation.filter_name is not None):
+            self.filterwheel.move_to(observation.filter_name)
+
```
wtgee: I think this should actually be pulled into `_setup_observation` rather than be directly placed in `take_observation`.

Related to that, it would be nice to populate the FITS headers during this PR as well, which happens in `_setup_observation`, so also seems to make sense.

pocs/camera/camera.py
```diff
@@ -302,7 +302,13 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
 
         # pop exptime from kwarg as its now in exptime
         exptime = kwargs.pop('exptime', observation.exptime.value)
-        exposure_event = self.take_exposure(seconds=exptime, filename=file_path, **kwargs)
+
+        # Move the filerwheel into position
+        if (self.filterwheel is not None) and (observation.filter_name is not None):
+            self.filterwheel.move_to(observation.filter_name)
```
wtgee:
```suggestion
            self.filterwheel.move_to(observation.filter_name, blocking=True)
```

Need to either directly block or capture the returned event and check it. Don't want to start an exposure in the middle of a filter movement.

wtgee: Seems like we also need to do some error checking here in case movement fails.

pocs/camera/camera.py
```diff
@@ -350,6 +356,7 @@ def take_exposure(self,
 
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
+        # Check that the filterwheel is ready
         if self.filterwheel and self.filterwheel.is_moving:
             msg = "Attempt to start exposure on {} while filterwheel is moving, ignoring.".format(
                 self)
```
wtgee: See note in review message.

pocs/camera/camera.py
AnthonyHorton: Wrong operator. In Python `and` is the operator for combining `bool` values, whereas `&` is the operator for doing bitwise ANDing of binary numbers (bitmaps/bitmasks).

pocs/camera/camera.py
```diff
@@ -707,6 +705,30 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
...
+            # Check that the filterwheel is not moving
+            if self.filterwheel.is_moving:
+
+                msg = f'Attempt to prepare observation on {self} while' \
+                        ' filterwheel is moving, ignoring.'
+                self.logger.error(msg)
+                raise error.PanError(msg)
+
+            # Move filterwheel if necessary
+            if observation.filter_name is not None:
```
AnthonyHorton: Might consider logging a message if the observation has a `filter_name`, but the camera doesn't have a filter wheel. A warning might be over the top, but perhaps `self.logger.info()`.

pocs/camera/camera.py
```diff
@@ -707,6 +705,30 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
...
+
+                try:
+                    self.filterwheel.move_to(observation.filter_name, blocking=True)
+                except Exception as e:
+                    self.logger.error(f'Error moving filterwheel on {self} to' \
+                                      f' {observation.filter_name}: {e}')
+                    raise(e)
+
+                # Store the filter name in metadata
+                metadata['filter_name'] = observation.filter_name
```
AnthonyHorton: I _think_ this is also redundant.... The current filter is always written into the FITS headers of any exposure taken with a POCS camera. I don't think that manually inserting it into the `metadata` `dict` here adds anything.

AnthonyHorton:
The blocking/non-blocking thing is awkward. It does need to be blocking in `setup_observation()` because otherwise `take_observation()` will immediately continue and call `take_exposure` before the filterwheel has finished moving. That means we have a blocking operation taking place before `take_observation()` returns, which means that all the filterwheel moves will take place serially instead of in parallel, wasting time and staggering the start times of the exposures.

Ideally we'd restructure this thing so that `take_observation` calls `setup_observation` in a separate thread and immediately returns with the `Event`. Then we can have the blocking filterwheel moves in `setup_observation` taking place in parallel, but `setup_observation` would have to take on the responsibility for subsequently starting the exposure and setting up the thread for `process_exposure`.

pocs/camera/camera.py
```diff
@@ -350,11 +353,6 @@ def take_exposure(self,
 
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
-        if self.filterwheel and self.filterwheel.is_moving:
-            msg = "Attempt to start exposure on {} while filterwheel is moving, ignoring.".format(
-                self)
-            raise error.PanError(msg)
-
```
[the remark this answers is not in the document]
AnthonyHorton: The reason the tests are failing is because you took this out.

pocs/camera/camera.py
danjampro: Does it? Seems like the headers get added in `AbstractCamera._process_fits` using `info=metadata` passed to `AbstractCamera._process_exposure`. I can't see any other way filter_name (which is a new attribute of Observation) is written to the headers.

AnthonyHorton: `._process_fits()` adds additional entries to FITS headers that relate to the details of the observation. The headers have already been created and populated with all the information about camera hardware state by the `._create_fits_headers()` methods of the camera classes and the `._add_fits_keywords()` methods of the focuser and filterwheel classes (remember that `.take_exposure()` is supposed to useful by itself, that's why it doesn't rely on `._process_fits` for this).

One of the keywords that `._create_fits_headers()` adds is `FILTER`, which is set from the camera class `.filter_type` property, and if a camera has a filterwheel that property will get its value from the filterwheel's `current_filter` property.  If the camera doesn't have a filterwheel this value comes either from the camera config for fixed bandpass filter (that's how it's worked for Huntsman up to now) or, failing that, from the camera itself (things like 'M' for monochrome camera or 'RGGB' for colour cameras).

So, the FITS headers will already contain the filter that was used to take the exposure. The question is whether it is useful to separately include the requested filter name from the observation. In general these will be different, either because the camera doesn't have a filterwheel so can't use a matching filter, or because the filter name that the filterwheel is using contains additional information (e.g. batch/serial number) after the band name.

pocs/tests/test_camera.py
```diff
@@ -496,13 +496,15 @@ def test_observation(camera, images_dir):
...
-    observation = Observation(field, exptime=1.5 * u.second)
+    observation = Observation(field, exptime=1.5 * u.second, filter_name='deux')
     observation.seq_time = '19991231T235959'
     camera.take_observation(observation, headers={})
     time.sleep(7)
     observation_pattern = os.path.join(images_dir, 'fields', 'TestObservation',
                                        camera.uid, observation.seq_time, '*.fits*')
     assert len(glob.glob(observation_pattern)) == 1
+    for _ in glob.glob(observation_pattern):
+        os.remove(_)
```
AnthonyHorton: Might be able to fix the test coverage by copying the test and having one with a filter, and one without.

pocs/camera/camera.py
```diff
@@ -641,6 +644,16 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
         if headers is None:
             headers = {}
 
+        # Move the filterwheel if necessary
+        if (self.filterwheel is not None) and (observation.filter_name is not None):
```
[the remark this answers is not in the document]
danjampro: I would probably argue its clearer / more transparent when explicit. In this case, I think it is sufficient?

PR #949, opened 2020-02-14
Add UTC timezone to current_time if datetime=True

danjampro:
The conversion to a `datetime.datetime` object from an `astropy.Time` object in `current_time` was missing a timezone. This means that calling `datetime.astimezone(utc)` on the returned `datetime` object leads to an incorrect time.

## How Has This Been Tested?
Tested out the changes in the python console.

## Types of changes

## Checklist:

### DIFF
--- a/pocs/utils/__init__.py
+++ b/pocs/utils/__init__.py
@@ -4,6 +4,7 @@ import shutil
 import signal
 import time
 
+from pytz import utc
 from astropy import units as u
 from astropy.coordinates import AltAz
 from astropy.coordinates import ICRS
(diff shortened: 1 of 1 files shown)


danjampro:
> Hm. This could potentially have some far reaching consequences, I would have to look at it a bit more. What prompted the change?

We were testing out `PEAS` and found that the weather entries were incorrectly being reported as -39562.5 seconds old. This is caused by the line in `peas_shell`:

`now = current_time(datetime=True).astimezone(utc)`

wtgee:
So I would say let's go ahead and make this PR in `panoptes-utils` under the assumption we will be using that by the end of the weekend unless making this change right here right now will unblock you.

wtgee:
@danjampro I'm closing this in favor of the one in `panoptes-utils`.

danjampro:
@wtgee O.K. but this is still a potential bug in `panoptes/pocs` unless I've missed something?

wtgee:
> @wtgee O.K. but this is still a potential bug in `panoptes/pocs` unless I've missed something?

The plan is to have all of that remove from panoptes/pocs with the merge of #951 some time this week. If it's blocking you until then we can just open it and merge it.

Issue #955, opened 2020-02-20
USB Device Path Auto-Detection


zacharyt20:
**Is your feature request related to a problem? Please describe.**
This feature request is not related to a problem with POCS. It can be challenging for new Linux users to find the device path (ex. ttyUSB0) of a connected USB device, and ideally this could be avoided with automated detection of the device path.

**Describe the solution you'd like**
Have a script that automatically finds the device path and relates it to a bus and device number, possibly updating the config files as well.

**Describe alternatives you've considered**
Detailed documentation of how to manually find device paths from bus and device numbers aimed at guiding new Linux users.

wtgee:
Closed as part of #956

PR #956, opened 2020-02-20
Script for USB device path autodetection

zacharyt20:
## Description
Added a python script that relates information found in the `/sys` directory to the bash command `lsusb`. This allows for automatic detection of device paths (ex. ttyUSB0), eliminating the need to manually find the device paths through trial and error. It also allows for an opportunity to include additional information from `lsusb` in an easily readable way.

## Related Issue
#955 

## How Has This Been Tested?

The new python script has been tested on PAN015 with an intel NUC. On that system, the directory with all of the necessary information to implement this feature was located at `/sys/devices/pci0000:00/0000:00:15.0/usb1`. There is also a `usb2` directory located in the same place that only has devices internal to the intel NUC (ex. Linux Foundation Roothub), and it is unclear if this is a lucky circumstance, or part of the Linux file structure.  
## Screenshots (if appropriate):
Sample Output from the script on PAN015:
image

## Types of changes

## Checklist:

### DIFF
--- a/resources/arduino_files/identifier/DallasTemperature.cpp
+++ b/resources/arduino_files/identifier/DallasTemperature.cpp
@@ -1,0 +1,885 @@
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+#include "DallasTemperature.h"
+
+#if ARDUINO >= 100
+#include "Arduino.h"
+#else
+extern "C" {
+#include "WConstants.h"
+}
+#endif
+
+// OneWire commands
+#define STARTCONVO      0x44  // Tells device to take a temperature reading and put it on the scratchpad
+#define COPYSCRATCH     0x48  // Copy EEPROM
+#define READSCRATCH     0xBE  // Read EEPROM
+#define WRITESCRATCH    0x4E  // Write to EEPROM
+#define RECALLSCRATCH   0xB8  // Reload from last known
+#define READPOWERSUPPLY 0xB4  // Determine if device needs parasite power
+#define ALARMSEARCH     0xEC  // Query bus for devices with an alarm condition
+
+// Scratchpad locations
+#define TEMP_LSB        0
+#define TEMP_MSB        1
+#define HIGH_ALARM_TEMP 2
+#define LOW_ALARM_TEMP  3
+#define CONFIGURATION   4
+#define INTERNAL_BYTE   5
+#define COUNT_REMAIN    6
+#define COUNT_PER_C     7
+#define SCRATCHPAD_CRC  8
+
+// Device resolution
+#define TEMP_9_BIT  0x1F //  9 bit
+#define TEMP_10_BIT 0x3F // 10 bit
+#define TEMP_11_BIT 0x5F // 11 bit
+#define TEMP_12_BIT 0x7F // 12 bit
+
+#define NO_ALARM_HANDLER ((AlarmHandler *)0)
+
+DallasTemperature::DallasTemperature()
+{
+#if REQUIRESALARMS
+	setAlarmHandler(NO_ALARM_HANDLER);
+#endif
+}
+DallasTemperature::DallasTemperature(OneWire* _oneWire)
+{
+	setOneWire(_oneWire);
+#if REQUIRESALARMS
+	setAlarmHandler(NO_ALARM_HANDLER);
+#endif
+}
+
+bool DallasTemperature::validFamily(const uint8_t* deviceAddress) {
+	switch (deviceAddress[0]) {
+	case DS18S20MODEL:
+	case DS18B20MODEL:
+	case DS1822MODEL:
+	case DS1825MODEL:
+	case DS28EA00MODEL:
+		return true;
+	default:
+		return false;
+	}
+}
+
+void DallasTemperature::setOneWire(OneWire* _oneWire) {
+
+	_wire = _oneWire;
+	devices = 0;
+	ds18Count = 0;
+	parasite = false;
+	bitResolution = 9;
+	waitForConversion = true;
+	checkForConversion = true;
+
+}
+
+// initialise the bus
+void DallasTemperature::begin(void) {
+
+	DeviceAddress deviceAddress;
+
+	_wire->reset_search();
+	devices = 0; // Reset the number of devices when we enumerate wire devices
+	ds18Count = 0; // Reset number of DS18xxx Family devices
+
+	while (_wire->search(deviceAddress)) {
+
+		if (validAddress(deviceAddress)) {
+
+			if (!parasite && readPowerSupply(deviceAddress))
+				parasite = true;
+
+			bitResolution = max(bitResolution, getResolution(deviceAddress));
+
+			devices++;
+			if (validFamily(deviceAddress)) {
+				ds18Count++;
+			}
+		}
+	}
+
+}
+
+// returns the number of devices found on the bus
+uint8_t DallasTemperature::getDeviceCount(void) {
+	return devices;
+}
+
+uint8_t DallasTemperature::getDS18Count(void) {
+	return ds18Count;
+}
+
+// returns true if address is valid
+bool DallasTemperature::validAddress(const uint8_t* deviceAddress) {
+	return (_wire->crc8(deviceAddress, 7) == deviceAddress[7]);
+}
+
+// finds an address at a given index on the bus
+// returns true if the device was found
+bool DallasTemperature::getAddress(uint8_t* deviceAddress, uint8_t index) {
+
+	uint8_t depth = 0;
+
+	_wire->reset_search();
+
+	while (depth <= index && _wire->search(deviceAddress)) {
+		if (depth == index && validAddress(deviceAddress))
+			return true;
+		depth++;
+	}
+
+	return false;
+
+}
+
+// attempt to determine if the device at the given address is connected to the bus
+bool DallasTemperature::isConnected(const uint8_t* deviceAddress) {
+
+	ScratchPad scratchPad;
+	return isConnected(deviceAddress, scratchPad);
+
+}
+
+// attempt to determine if the device at the given address is connected to the bus
+// also allows for updating the read scratchpad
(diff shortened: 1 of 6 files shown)


wtgee:
Thanks @zacharyt20, I'll get into the lab tomorrow to test this.

zacharyt20:
Update:
An issue stopping the script from being adapted to directly input device path information into the config files is that there are identical device names. From the example screen shot, there is no way to determine which Prolific Technology Serial Adapter is the weather sensor, and which is the mount. I have been searching for a way to distinguish between these two adapters, and may have found a way to do so using `udevadm info --path=/sys/path/to/devpath -a`.

I was comparing the output of `udevadm info` for the mount (`ttyUSB0` on our unit) and the weather sensor ('ttyUSB1` on our unit).  Almost everything is the same, except for the removable attribute. For the mount the output is:

`ATTRS{removable}=="removable"`

And for the weather sensor it is:

` ATTRS{removable}=="unknown"`

I'm hoping that this is specific to each device, but want to confirm that before adding a parser to the detection script. If you could also let me know what the output of `udevadm info --path=/sys/devices/pci0000:00/0000:pci_leading_to_usb1/path/to/ttyUSB_of_mount_and_weather_sensor -a` on your system when you test out the script today that would be helpful. 

I am also noticing that the `udevadm info` for the Arduinos have different serial attributes, which could possibly be used to find out which one is on each board.

wtgee:
Thanks @zacharyt20. I haven't tested this in the lab yet but another approached would be to just probe each of the usb devices for the expected output.

That is, if you did something like:

```py
from pocs.mount import create_mount_from_config
from panoptes.utils.error import MountNotFound

from panoptes.utils.config.client import set_config

# However many you wanted to test or could somehow discover what
# was plugged in and only test those.
for port_num in [0, 1, 2]:  
    usb_port = f'/dev/ttyUSB{port_num}'

    mount_info = {
        'driver': 'ioptron',
        'serial': {
            'port': usb_port
        }
    }

    mount = create_mount_from_config(mount_info=mount_info)

    try:
        mount.initialize()
    except MountNotFound:
        print(f'Port {usb_port} invalid')
        continue

    print(f'Found mount on {usb_port}, saving to config')    
    set_config('mount.serial.port', usb_port)
```

That's totally untested but something similar to that. Could do the same with the AAG weather sensor.

We could also feasibly parse the output for the arduinos and do the same thing.

zacharyt20:
That is definitely a better approach. I'm testing it out, but I'm not even able to import the methods successfully. The first time, the script couldn't find the pocs module, which I fixed by adding `sys.path.insert(0, "/var/panoptes/POCS"` before importing. However this led to another ModuleNotFoundError:
```
Traceback (most recent call last):
  File "lsusbplus.py", line 8, in <module>
    from pocs.mount import create_mount_from_config
  File "/var/panoptes/POCS/pocs/mount/__init__.py", line 4, in <module>
    from pocs.mount.mount import AbstractMount  # pragma: no flakes
  File "/var/panoptes/POCS/pocs/mount/mount.py", line 3, in <module>
    from astropy import units as u
ModuleNotFoundError: No module named 'astropy'
```
Do you know if this is an issue with not importing something in the python script or an issue with how the working directory or python environment is setup?

wtgee:
> That is definitely a better approach. I'm testing it out, but I'm not even able to import the methods successfully. The first time, the script couldn't find the pocs module, which I fixed by adding `sys.path.insert(0, "/var/panoptes/POCS"` before importing. However this led to another ModuleNotFoundError:
> 
> ```
> Traceback (most recent call last):
>   File "lsusbplus.py", line 8, in <module>
>     from pocs.mount import create_mount_from_config
>   File "/var/panoptes/POCS/pocs/mount/__init__.py", line 4, in <module>
>     from pocs.mount.mount import AbstractMount  # pragma: no flakes
>   File "/var/panoptes/POCS/pocs/mount/mount.py", line 3, in <module>
>     from astropy import units as u
> ModuleNotFoundError: No module named 'astropy'
> ```
> 
> Do you know if this is an issue with not importing something in the python script or an issue with how the working directory or python environment is setup?

@zacharyt20 I believe you are running with the docker setup so everything needs to be run inside the docker container.  You can put your script in `scripts/your-script.py` and do:

```bash
bin/pocs-cmd "scripts/your-script.py"
```

Or if you want to do it interactively in a jupyter notebook do:

```
bin/pocs-cmd jupyter-console
```

You should then be running jupyter within the running container with all the dependencies installed.

zacharyt20:
Running the script through the the `pocs-cmd` fixed the import issues. It also caused another issue since the command `lsusb` is not available inside a docker container. This stops the script from relating the device path to the bus number and device number. Luckily, it should be an easy fix. If [this](https://stackoverflow.com/questions/48884854/lsusb-command-brings-no-results-on-docker-container-ubuntu) stack overflow post is correct, starting/creating the docker containers with privileged access should work, although it could open security risks. The other option is that the auto-detection script doesn't necessarily need `lsusb`. The information in the `usb1` folder is very similar to the output of `lsusb`, so that can be stored by the script and displayed. Looking back at the code, it might actually be simpler to collect all of this information in the script rather than parsing `lsusb` and inserting specific information. The output won't be exactly the same, but close enough to be effective. Can we run privileged docker containers or should the script get information?

As for the test code for distinguishing the mount adapter from the weather adapter, I merged it into the `lsusbplus.py` script and had it testing device paths starting with `ttyUSB`. However, the test code thought that the weather sensor was the mount.  (But it did change the device path in `pocs_local.yaml` perfectly!) Weirdly, it did this once the first time I tested the script, and then it kept on saying that both USB device paths were invalid. For trouble shooting, where is the `initialize()` method defined? Any ideas on how this happened?

wtgee:
> Can we run privileged docker containers or should the script get information?

The docker containers should already be running with privileged. If you look in the `docker/docker-compose.yaml` you should see `privileged: true` defined for each service.

I think the issue is that I was originally trying to keep the docker images small so `usbutils`, which provides `lsusb` is not installed.  It could be installed with `apt` temporarily and we could build it into the machines if needed.

Note also that the docker containers are only aware of the usb devices that exist when it is started. So if you have, for instance, the mount plugged in as `/dev/ttyUSB0` and then unplug it and plug back in, the docker container won't be aware of that change until you do a `docker restart pocs-shell`.  That's annoying and I hope we can overcome that somehow but might be a limitation.

`initialize` is part of `pocs.mount.ioptron.py` but I'm not sure debugging there would help. The issue might be related to the above docker restart issue.

I'm running around so sorry for lack of detailed answer. Let me know if that helps you any otherwise feel free to ask more. 

Thanks for all the work!

wtgee:
@zacharyt20 sorry, I still haven't had a chance to look at this thoroughly and I will be flying later today so probably won't get to it for a day or two.

However, please take a look at the [Code Formatting](https://github.com/panoptes/POCS/blob/develop/CONTRIBUTING.md#code-formatting) section of the CONTRIBUTING doc.  We pretty much follow python standards so it's good to get in the habit of coding that way. It's a bit confusing because we are switching languages from something like the Arduino that you have also been working on, which tends to have a style more like this script.

I don't actually use the `yapf` as described there but have `pycodestyle` and `autopep8` plugins for my editor (they exist for all editors).

Let me know if you have any questions!

zacharyt20:
It seems that the `initialize()` function uses the device path set in `pocs_local.yaml` to help confirm if it is trying to initialize the mount. I didn't see this in the `initialize()` code, but found out through testing. I saw that whenever the mount's serial port in `pocs_local.yaml` was incorrect the script said that the USB device paths being tested were invalid. To get around this, I had the script set the mount's serial port in the config file to the port that is being tested. The begging of the script that @wtgee suggested is now:

```
for port_num in [0, 1, 2]:  
    usb_port = f'/dev/ttyUSB{port_num}'
    set_config('mount.serial.port', usb_port)
    mount_info = {
        'driver': 'ioptron',
        'serial': {
            'port': usb_port
        }
    }
```
On our system, it correctly identifies that `/dev/ttyUSB0` is the mount.

That leaves detection of the weather sensor and Arduinos. 
### Weather Sensor
Does the peas-shell have a `weather.initialize()` function similar to this or does that need to be developed from scratch?
### Arduinos
The `peas-shell` already gets serial output from the Arduinos. Could the `last_reading` or `load_sensor_board` methods be used to find which Arduino is the one for the camera/control board? If not, the other approach I can think of is uploading a sketch to a board that tries getting input from the `control_board`'s 3 temperature sensors. If the input is in the right format, that sketch is the control board and if not it is the camera board. 

A build like PAN012, would use an Arduino Micro instead of Uno in the camera box. Hopefully this makes things easier and the product ID, description, or output of `arduino-cli board list` can be used to distinguish between the camera box and control box Arduinos.

zacharyt20:
@wtgee 
I have been trying to get output from the weather sensor but have had no real success. I started just by looking at serial output from the weather sensor's usb path (`/dev/ttyUSB1`) with a very simple python script, but the output was not human readable. I know that this information is somehow being parsed and displayed in the `peas-shell`, and I am assuming that the parsing/organizing of the serial output is going on somewhere in the `aag-weather` repo/containers.

I looked there to try and find a way of getting real-time output from the weather sensor and came across the `read-aag` script.  I tried using it from within the `aag-weather-reader` container, but did not get any output in terminal and was kicked out of the container. This was the behavior:
```
root@pan015:/app# scripts/read-aag.py --config-file config.yaml --serial-address /dev/ttyUSB1 --verbose               
WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError. [astropy.units.quantity]                                                                                        
panoptes@pan015:/var/panoptes/aag-weather/aag$
```
In the help section I see that there is an option to store information to a db file. Is that where weather/sensor information is stored as opposed to being printed in terminal? If so, where is it? Also, is it even possible to execute commands in one of the aag-weather containers from a single script that is called in the `pocs-shell` container with `pocs-cmd`?

zacharyt20:
@wtgee 
Auto USB detection is done! 

The only thing left to do is to make the script work with other builds. I know one of these will be the setup using an Arduino Micro in the camera box. Is there a way to see the output of `lsusb` on a unit with the Micro setup? I can't think of another build where this script would not work.

wtgee:
Thanks for the work on this @zacharyt20! 

The newer design has things much simplified. Some aspects of this were incorporated although we now mostly use firmata on the arduino.

Closing as stale.

PR #959, opened 2020-03-03
Increase default temperature_tolerance for ZWO cameras

danjampro:
## Description
The temperature resolution for ZWO cameras is 0.5 deg, meaning that the current default of  `temperature_tolerance=0.5` can cause `camera.is_temperature_stable` to undesirably return `False`. Increasing the default to `temperature_tolerance=0.6` for the ZWO cameras solves this problem.

## How Has This Been Tested?
Tested running POCS on actual hardware with default value overridden.

## Types of changes

## Checklist:

### DIFF
--- a/pocs/camera/zwo.py
+++ b/pocs/camera/zwo.py
@@ -50,6 +50,9 @@ class Camera(AbstractSDKCamera):
 
         super().__init__(name, ASIDriver, *args, **kwargs)
 
+        # Increase default temperature_tolerance for ZWO cameras
+        self.temperature_tolerance = kwargs.get('temperature_tolerance', 0.6 * u.Celsius)
+
         if gain:
             self.gain = gain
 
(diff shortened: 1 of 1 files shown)


AnthonyHorton:
> Why not change the default for all cameras in `camera.py`? Or is it just because of the 0.5 tolerance on ZWO and the others presumably won't have those?

Yes, the default value of 0.5C set in the camera base class is only causing a problem for the ZWO cameras, because the resolution of their temperature readings is also 0.5C. Both the FLI and SBIG cameras have higher resolution temperature readings and can easily achieve better than 0.5C stability. Increasing the tolerance for ZWO only is, I think, the right thing to do.

pocs/camera/zwo.py
```diff
@@ -50,6 +50,9 @@ def __init__(self,
 
         super().__init__(name, ASIDriver, *args, **kwargs)
 
+        # Increase default temperature_tolerance for ZWO cameras
```
AnthonyHorton: Might be nice to expand this comment to say _why_ this needed to be done too.

PR #965, opened 2020-04-16
Fix skyx timeout

danjampro:
## Description
This PR removes the redundant timeout functionality in `pocs.mount.bisque.read` to instead rely on the equivalent funtionality in `pocs.utils.theskyx.read`. It also adds functionality to pass the `timeout` key word argument to `pocs.utils.theskyx.read` via `pocs.mount.mount.query`, and increases the default timeout for movements in `pocs.mount.bisque`.

This fixes TheSkyXTimeout errors encountered after running:
```
from pocs.mount import create_mount_from_config 
from huntsman.utils import load_config 
config = load_config()  
mount = create_mount_from_config(config)   
mount.initialize() 
mount.unpark()
mount.home_and_park() 
```

I think the problem is twofold:
a) in `pocs.mount.bisque.read` there is a timeout countdown. This calls `theskyx.read` at each time step, without catching any errors. It incorrectly expects `theskyx.read` to return None if there is no response, but instead it implicitly raises a TheSkyXTimeout error , causing the timeout functionality in `pocs.mount.bisque.read` to be ignored.
b) The default timeout in `pocs.mount.bisque.read` is 5s, which is too short for typical slewings.

## How Has This Been Tested?
Tested on hardware (Huntsman telscope).

## Types of changes

## Checklist:

### DIFF
--- a/pocs/mount/bisque.py
+++ b/pocs/mount/bisque.py
@@ -157,7 +157,7 @@ class Mount(AbstractMount):
 # Movement methods
 ##########################################################################
 
-    def slew_to_target(self):
+    def slew_to_target(self, timeout=60):
         """ Slews to the current _target_coordinates
 
         Args:
@@ -182,7 +182,7 @@ class Mount(AbstractMount):
                 response = self.query('slew_to_coordinates', {
                     'ra': mount_coords[0],
                     'dec': mount_coords[1],
-                })
+                }, timeout=timeout)
                 success = response['success']
                 if success:
                     self.status()
@@ -203,7 +203,7 @@ class Mount(AbstractMount):
 
         return success
 
-    def slew_to_home(self, blocking=False):
+    def slew_to_home(self, blocking=False, timeout=60):
         """ Slews the mount to the home position.
 
         Note:
@@ -220,7 +220,7 @@ class Mount(AbstractMount):
 
         if not self.is_parked:
             self._target_coordinates = None
-            response = self.query('goto_home')
+            response = self.query('goto_home', timeout=timeout)
         else:
             self.logger.info('Mount is parked')
 
@@ -230,7 +230,7 @@ class Mount(AbstractMount):
         """ Calls `slew_to_home` in base class. Can be overridden.  """
         self.slew_to_home(blocking=blocking)
 
-    def park(self):
+    def park(self, timeout=60):
         """ Slews to the park position and parks the mount.
 
         Note:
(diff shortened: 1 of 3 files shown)


pocs/mount/bisque.py
```diff
@@ -301,14 +302,14 @@ def write(self, value):
...
-            if response is not None or timeout == 0:
-                break
-            else:
-                time.sleep(1)
-                timeout -= 1
+
+        try:
+            response = self.theskyx.read(timeout=timeout)
+        except Exception as err:
+            self.logger.error(err)
```
danjampro, line 311: @wtgee, Not sure whether you'd rather the error be raised here.

pocs/mount/mount.py
```diff
@@ -698,7 +698,7 @@ def get_ms_offset(self, offset, axis='ra'):
 
         return (offset / (self.sidereal_rate * guide_rate)).to(u.ms)
 
-    def query(self, cmd, params=None):
+    def query(self, cmd, params=None, **kwargs):
```
danjampro, line 701: I used **kwargs here so that the other mounts are unaffected.

pocs/mount/bisque.py
wtgee: I think this was put in mostly as a debugging check because we were getting exceptions but didn't know why but wanted to.  Could be removed now.

pocs/mount/bisque.py
```diff
@@ -301,14 +302,14 @@ def write(self, value):
         return self.theskyx.write(value)
 
     def read(self, timeout=5):
```
wtgee, line 304: Why not just change the default to 120 seconds here?

We could still feasibly allow for timeouts for each method via the keywords you have added or just pass the `kwargs` through for each command method.

pocs/mount/bisque.py
danjampro: O.K., I'll remove it.

pocs/mount/bisque.py
danjampro: Basically because a 5s timeout is all we really need for the other (non-movement related) SkyX operations.

danjampro: > I don't see the "removes the redundant timeout functionality in `pocs.mount.bisque.read`" part of this.
> 
> I'm approving as well but see my note about the default timeout. Could probably just change the default timeout and not bother with the rest of the PR or do as note suggests.

@wtgee Did you mean to approve?

PR #967, opened 2020-04-22
Horizon ranges

danjampro:
Add support for a state to specify a "bright horizon", meaning that `pocs.is_safe()=True` only when the Sun is between `horizon` and `horizon_bright`. Useful for e.g. taking twilight flat fields before it gets too dark.

## Types of changes

## Checklist:

### DIFF
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -291,6 +291,8 @@ class POCS(PanStateMachine, PanBase):
                 defaults to False.
             horizon (str, optional): For night time check use given horizon,
                 default 'observe'.
+            horizon_bright (str, optional): Horizon spcifying minimum brightness,
+                default None.
         Returns:
             bool: Latest safety flag
 
--- a/pocs/core.py
+++ b/pocs/core.py
@@ -276,7 +276,8 @@ class POCS(PanStateMachine, PanBase):
 # Safety Methods
 ##################################################################################################
 
-    def is_safe(self, no_warning=False, horizon='observe', **kwargs):
+    def is_safe(self, no_warning=False, horizon='observe', horizon_bright=None,
+                **kwargs):
         """Checks the safety flag of the system to determine if safe.
 
         This will check the weather station as well as various other environmental
@@ -308,6 +309,10 @@ class POCS(PanStateMachine, PanBase):
 
         # Check if night time
         is_safe_values['is_dark'] = self.is_dark(horizon=horizon)
+
+        # Check if its bright enough (e.g. for twilight flats)
+        if horizon_bright is not None:
+            is_safe_values['is_bright'] = not self.is_dark(horizon=horizon_bright)
 
         # Check weather
         is_safe_values['good_weather'] = self.is_weather_safe()
--- a/pocs/state/machine.py
+++ b/pocs/state/machine.py
@@ -43,7 +43,7 @@ class PanStateMachine(Machine):
                         for transition in state_machine_table['transitions']]
 
         # States can require the horizon to be at a certain level.
-        self._horizon_lookup = dict()
+        self._horizon_lookup = {'dark': dict(), 'bright': dict()}
 
         # Setup States.
         states = [
@@ -135,12 +135,19 @@ class PanStateMachine(Machine):
 
             # If we are processing the states
             if self.do_states:
+
                 # Wait for horizon level if state requires.
                 self.logger.warning(f'Checking horizon limits for next state: {self.next_state}')
+                _horizon_kwargs = {}
                 with suppress(KeyError):
-                    required_horizon = self._horizon_lookup[self.next_state]
-                    self.logger.info(f'Horizon limit for {self.state}: {required_horizon}')
-                    self.wait_until_safe(horizon=required_horizon)
+                    _horizon = self._horizon_lookup['dark'][self.next_state]
+                    _horizon_kwargs['horizon'] = _horizon
+                    self.logger.info(f'Horizon limit for {self.state}: {_horizon}')
+                with suppress(KeyError):
+                    _horizon = self._horizon_lookup['bright'][self.next_state]
+                    _horizon_kwargs['horizon_bright'] = _horizon
+                    self.logger.info(f'Bright horizon limit for {self.state}: {_horizon}')
+                self.wait_until_safe(**_horizon_kwargs)
 
                 self.logger.warning('Going to next state')
                 try:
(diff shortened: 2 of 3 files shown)


wtgee:
I'm not terribly keen on adding the `horizon_bright` to the `is_safe` and related methods, although I'm not sure why. Doesn't seem very clean to me.

How would the bright horizon differ from the flat, focus, and observe horizons?  I realize we talked about this some in the slack channel but it's not clear to me that this does immediately solve the problem.

If we really need to finely distinguish between horizon levels I guess I would rather build in more thorough support so that a `pocs` instance (really the `pocs.observatory.observer`) can directly figure out horizon levels rather than have some side state logic.  

Thoughts? @AnthonyHorton @danjampro

danjampro:
> I'm not terribly keen on adding the `horizon_bright` to the `is_safe` and related methods, although I'm not sure why. Doesn't seem very clean to me.
> 
> How would the bright horizon differ from the flat, focus, and observe horizons? I realize we talked about this some in the slack channel but it's not clear to me that this does immediately solve the problem.
> 
> If we really need to finely distinguish between horizon levels I guess I would rather build in more thorough support so that a `pocs` instance (really the `pocs.observatory.observer`) can directly figure out horizon levels rather than have some side state logic.
> 
> Thoughts? @AnthonyHorton @danjampro

Thanks for looking at this. We are basically just trying to get the state machine to wait until it is bright enough to take flats. This is particularly important when entering the morning twilight flats state after finishing observing. So in this case, we could have something like `bright_horizon='focus'`.

I'm not sure what you mean by "clean" in this case. I would say it is, because it leaves current use cases unaffected while implementing (I think) our desired behaviour, with not too much additional code - most of it just mirrors existing code for the normal horizon. But I'm interested to hear your suggestions!

AnthonyHorton:
@wtgee The re-use of the `focus` horizon as the `bright_horizon` for flat fielding is possibly obscuring the intent here.

The rationale behind these changes is that the existing `is_dark` and `is_safe` checks, and the corresponding wait until dark mechanism in the state machine, only work with a single Sun altitude limit (horizon), a maximum Sun altitude. For twilight flats we need the Sun altitude to be within a specific range, i.e. a single state needs to have both a minimum and maximum Sun altitude associated with it. This is especially important for morning twilight flats because without this the state machine will transition to the twilight flats state as soon as it gets too bright for observing, when it is still far too dark for flats. We could implement a 'wait until morning twilight' function in  `huntsman-pocs` but that would involve duplicating some of the existing functionality in POCS, so it seems cleaner to add support for a minimum Sun altitude to POCS instead.

wtgee:
> @wtgee The re-use of the `focus` horizon as the `bright_horizon` for flat fielding is possibly obscuring the intent here.
> 
> The rationale behind these changes is that the existing `is_dark` and `is_safe` checks, and the corresponding wait until dark mechanism in the state machine, only work with a single Sun altitude limit (horizon), a maximum Sun altitude. For twilight flats we need the Sun altitude to be within a specific range, i.e. a single state needs to have both a minimum and maximum Sun altitude associated with it. This is especially important for morning twilight flats because without this the state machine will transition to the twilight flats state as soon as it gets too bright for observing, when it is still far too dark for flats. We could implement a 'wait until morning twilight' function in `huntsman-pocs` but that would involve duplicating some of the existing functionality in POCS, so it seems cleaner to add support for a minimum Sun altitude to POCS instead.

It's not the rationale that is confusing me at all or that I don't like, it's the implementation.

Our horizons already have "borders" so I was thinking we could just use the knowledge of what horizon we are in to define those borders. I guess I was wondering if those bright limits need to take place between established horizon limits or not.  That is, is `horizon_bright` always going to be equal to the `focus` altitude? Even if we need an in-between it seems like we could add another horizon level, then just tag the state with the existing mechanism to require certain horizons.

In particular it is adding the `bright_horizon` as a parameter to `is_safe` that seems off to me for some reason.

I don't have a ton of better suggestions however.

danjampro:
> > @wtgee The re-use of the `focus` horizon as the `bright_horizon` for flat fielding is possibly obscuring the intent here.
> > The rationale behind these changes is that the existing `is_dark` and `is_safe` checks, and the corresponding wait until dark mechanism in the state machine, only work with a single Sun altitude limit (horizon), a maximum Sun altitude. For twilight flats we need the Sun altitude to be within a specific range, i.e. a single state needs to have both a minimum and maximum Sun altitude associated with it. This is especially important for morning twilight flats because without this the state machine will transition to the twilight flats state as soon as it gets too bright for observing, when it is still far too dark for flats. We could implement a 'wait until morning twilight' function in `huntsman-pocs` but that would involve duplicating some of the existing functionality in POCS, so it seems cleaner to add support for a minimum Sun altitude to POCS instead.
> 
> It's not the rationale that is confusing me at all or that I don't like, it's the implementation.
> 
> Our horizons already have "borders" so I was thinking we could just use the knowledge of what horizon we are in to define those borders. I guess I was wondering if those bright limits need to take place between established horizon limits or not. That is, is `horizon_bright` always going to be equal to the `focus` altitude? Even if we need an in-between it seems like we could add another horizon level, then just tag the state with the existing mechanism to require certain horizons.
> 
> In particular it is adding the `bright_horizon` as a parameter to `is_safe` that seems off to me for some reason.
> 
> I don't have a ton of better suggestions however.

I'm not sure that we will end up using `horizon='focus'` as maybe even that is too dark for the flats, we will have to see during commissioning.  

If you don't like the extra horizon being passed into `is_safe` via `wait_until_safe` (as far as I can see, the simplest solution), then there will have to be an additional call to some sort of wait function somewhere else in the code, which is the sort of thing we were trying to avoid doing in huntsman-pocs. I don't really get what "tagging" the states means though, so maybe there is a better solution - if you can suggest one? Maybe a function like `wait_for_state_conditions` can replace the call to `wait_until_safe` or something...

AnthonyHorton:
> Our horizons already have "borders" so I was thinking we could just use the knowledge of what horizon we are in to define those borders. I guess I was wondering if those bright limits need to take place between established horizon limits or not. That is, is `horizon_bright` always going to be equal to the `focus` altitude? Even if we need an in-between it seems like we could add another horizon level, then just tag the state with the existing mechanism to require certain horizons.

No, in general the minimum Sun altitude limit for twilight flats will not be the same as the maximum Sun altitude limit for focusing. Even if they do have the same numerical value for now they should certainly not be sharing the same name because they have distinct purposes Re-using the `focus` horizon instead of adding a new one was a mistake here, & in general the naming could be improved to make things clearer.

I guess I missed the horizons' "borders".  How do they work?

A note about relevant future plans. The twilight flats autoexposure algorithm is great but I think it's clear that when using twilight flats for wide field imaging systems you really want to be using consistent Sun altitudes for best results, especially between evening and morning twilight flats so that you can cancel sky gradients. Once we've actually got a bit more twilight flat data I think we can pretty quickly settle on a very narrow range of Sun altitudes that work best. These will be different for each filter, of course. Most of the implementation for all of that will naturally go into the `twilight_flat_fielding` state, but having a good 'safely wait until the sun altitude is between these two limits` function is going to be very useful.

wtgee:
@danjampro is this PR still relevant?

wtgee:
Closing in favor of https://github.com/AstroHuntsman/huntsman-pocs/pull/183

PR #973, opened 2020-06-04
Fixes the install-pocs script

jlibermann:
## Description
This fixes the issue of not being able to select the `--developer` option in the latest `install-pocs` script. It also fixes the issue of the `install-pocs.log` and `env` files not being created in PANDIR. 

## Related Issue

Fixes #972 

## Types of changes

## Checklist:

### DIFF
--- a/docker/setup-local-environment.sh
+++ b/docker/setup-local-environment.sh
@@ -7,7 +7,6 @@ cd "${POCS}"
 
 echo "Building local panoptes-pocs:latest"
 docker build \
-    --quiet \
     -t "panoptes-pocs:latest" \
     -f "${POCS}/docker/latest.Dockerfile" \
     "${POCS}"
@@ -17,7 +16,6 @@ sed -i s'/^\.git$/\!\.git/' .dockerignore
 
 echo "Building local panoptes-pocs:develop"
 docker build \
-    --quiet \
     --build-arg IMAGE_URL="panoptes-pocs:latest" \
     -t "panoptes-pocs:develop" \
     -f "${POCS}/docker/develop.Dockerfile" \
@@ -25,7 +23,6 @@ docker build \
 
 echo "Building local panoptes-pocs:developer-env"
 docker build \
-    --quiet \
     --build-arg IMAGE_URL="panoptes-pocs:develop" \
     -t "panoptes-pocs:developer-env" \
     -f "${POCS}/docker/developer-env.Dockerfile" \
--- a/docker/setup-local-environment.sh
+++ b/docker/setup-local-environment.sh
@@ -7,6 +7,7 @@ cd "${POCS}"
 
 echo "Building local panoptes-pocs:latest"
 docker build \
+    --quiet \
     -t "panoptes-pocs:latest" \
     -f "${POCS}/docker/latest.Dockerfile" \
     "${POCS}"
@@ -16,6 +17,7 @@ sed -i s'/^\.git$/\!\.git/' .dockerignore
 
 echo "Building local panoptes-pocs:develop"
 docker build \
+    --quiet \
     --build-arg IMAGE_URL="panoptes-pocs:latest" \
     -t "panoptes-pocs:develop" \
     -f "${POCS}/docker/develop.Dockerfile" \
@@ -23,6 +25,7 @@ docker build \
 
 echo "Building local panoptes-pocs:developer-env"
 docker build \
+    --quiet \
     --build-arg IMAGE_URL="panoptes-pocs:develop" \
     -t "panoptes-pocs:developer-env" \
     -f "${POCS}/docker/developer-env.Dockerfile" \
--- a/scripts/install/install-pocs.sh
+++ b/scripts/install/install-pocs.sh
@@ -141,7 +141,6 @@ function make_directories {
 }
 
 function setup_env_vars {
-
     echo "Writing environment variables to ${ENV_FILE}"
     if  [[ -f "${ENV_FILE}" ]]; then
         echo "\n**** Added by install-pocs script ****\n" >> "${ENV_FILE}"
--- a/scripts/install/install-pocs.sh
+++ b/scripts/install/install-pocs.sh
@@ -3,7 +3,7 @@ set -e
 
 usage() {
   echo -n "##################################################
-# Install POCS and friends.
+# Install POCS l friends.
 #
 # This script is designed to install the PANOPTES Observatory
 # Control System (POCS) on a cleanly installed Ubuntu system.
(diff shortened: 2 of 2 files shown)


scripts/install/install-pocs.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 usage() {
   echo -n "##################################################
-# Install POCS and friends.
+# Install POCS l friends.
```
jlibermann, line 6: I must have accidentally changed 'and' to 'l'. Disregard this.

scripts/install/install-pocs.sh
```diff
@@ -223,7 +223,7 @@ function get_or_build_images {
     if ${DEVELOPER}; then
         echo "Building local PANOPTES docker images."
 
-        cd "${POCS}"
+        cd "${POCS}"                           # TODO: Fix environment variables, 'CD' is not working
```
jlibermann, line 226: I ended up resolving this by linking the `env` file in `bashrc`.

scripts/install/install-pocs.sh
wtgee: There are a whole bunch of these and they'll need to be changed back in the PR. I can't selectively pick parts.

docker/setup-local-environment.sh
```diff
@@ -7,7 +7,6 @@ cd "${POCS}"
 
 echo "Building local panoptes-pocs:latest"
 docker build \
-    --quiet \
```
wtgee, line 10: Can we add this back if it is working now or is it helpful to see the output?  I was just worried it was too much.

scripts/install/install-pocs.sh
```diff
@@ -50,15 +50,15 @@ usage() {
  If in DEVELOPER mode, the following options are also available:
   USER      The PANUSER environment variable, defaults to current user (i.e. PANUSER=$USER).
   PANDIR    Default install directory, defaults to PANDIR=${PANDIR}. Saved as PANDIR
-            environment variable.
+            environment variable. Test $LOGFILE.
```
wtgee, line 53: There is a line somewhere below that prints all the options where this could be added. But shouldn't be needed here.

```suggestion
            environment variable.
```

scripts/install/install-pocs.sh
```diff
@@ -50,15 +50,15 @@ usage() {
...
 "
 }
 
 
 DEVELOPER=${DEVELOPER:-false}
 PANUSER=${PANUSER:-$USER}
 PANDIR=${PANDIR:-/var/panoptes}
-LOGFILE="${PANDIR}/install-pocs.log"
+LOGFILE=${PANDIR}/install-pocs.log
```
wtgee, line 61: Was this necessary to make it work? Just curious.

scripts/install/install-pocs.sh
wtgee: So does the comment need to be here?

scripts/install/install-pocs.sh
```diff
@@ -248,15 +248,19 @@ function do_install {
     echo "OS: ${OS}"
     echo "Logfile: ${LOGFILE}"
 
-    exit 0;
```
wtgee, line 251: Oh, oops.  No wonder it wasn't running. :laughing:

scripts/install/install-pocs.sh
```diff
@@ -248,15 +248,19 @@ function do_install {
...
+  
 
     echo "Creating directories in ${PANDIR}"
     make_directories
 
+    echo "Setting up environment variables"
+    setup_env_vars
```
wtgee, line 257: Looks like this function didn't get included?

docker/setup-local-environment.sh
jlibermann: I found it helpful in case the build were to throw any errors, but I can add it back.

scripts/install/install-pocs.sh
jlibermann: No, it actually doesn't seem to make a difference.

jlibermann:
Comments to PR have been addressed.

scripts/install/install-pocs.sh
```diff
@@ -223,7 +223,7 @@ function get_or_build_images {
     if ${DEVELOPER}; then
         echo "Building local PANOPTES docker images."
 
-        cd "${POCS}"
+        cd "${POCS}"
```
wtgee, line 226: Your editor should be set to automatically strip extra white space.

scripts/install/install-pocs.sh
```diff
@@ -92,6 +93,16 @@ case ${key} in
...
 done
 
+if ! ${DEVELOPER}; then
+    echo "WARNING "developer" mode not detected. You can exit and install in "developer" mode with --developer or continue."
+    select yn in "Yes" "No"; do
+        case ${yn} in
+            Yes ) echo "Proceeding with existing installation settings"; break;;
+            No ) echo "Exiting"; exit 1;;
+        esac
+    done
```
wtgee, line 103: Mostly I meant to copy the structure for asking a yes/no question in make_directories.  Here it's a bit different.

This doesn't have to be a warning and that would scare the people who are installing this on a unit (i.e. not in developer mode).

We also don't need to exit and start over, we can just set the variable based on their answer. `make_directories` was different because they would presumably have to go deal with the directory before installing.

```suggestion
    echo -n "Are you installing POCS as a developer? (for PANOPTES units, select No)"
    select yn in "Yes" "No"; do
        case ${yn} in
            Yes ) echo "Enabling developer mode. Note that you will need your GitHub username to proceed"; DEVELOPER=true; break;;
            No ) echo "Installing POCS in production mode"; break;;
        esac
    done
```

scripts/install/install-pocs.sh
```diff
@@ -132,7 +143,7 @@ function setup_env_vars {
     ENV_FILE="${PANDIR}/env"
```
wtgee, line 143: Let's move this line to the top of the file with the other variables.  See also the note below.

scripts/install/install-pocs.sh
```diff
@@ -248,11 +259,15 @@ function do_install {
     echo "OS: ${OS}"
     echo "Logfile: ${LOGFILE}"
 
-    exit 0;
+
 
     echo "Creating directories in ${PANDIR}"
     make_directories
 
+    echo "Setting up environment variables"
```
wtgee, line 267: Let's tell them where they end up.

```suggestion
    echo "Setting up environment variables in ${ENV_FILE}"
```

jlibermann:
I successfully ran the `install-pocs` script on a "dirty" installation of Ubuntu 20.03.

PR #984, opened 2020-07-08
Update install script (again)

wtgee:
* Test for ssh access for the github user.
* Better install instructions in readme and script.

### DIFF
--- a/scripts/install/install-pocs.sh
+++ b/scripts/install/install-pocs.sh
@@ -10,7 +10,7 @@ usage() {
 #
 # This script is meant for quick & easy install via:
 #
-#   $ curl -L https://install.projectpanoptes.org | bash
+#   $ curl -fsSL https://install.projectpanoptes.org | bash
 #   or
 #   $ wget -O - https://install.projectpanoptes.org | bash
 #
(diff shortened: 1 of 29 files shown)


scripts/install/install-pocs.sh
```diff
@@ -255,7 +269,7 @@ function get_or_build_images {
     if ${DEVELOPER}; then
         echo "Building local PANOPTES docker images."
 
-        cd "${POCS}"
+        cd "${PANDIR}/POCS"
         ./docker/setup-local-environment.sh
```
jlibermann, line 273: Returned `Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/prune: dial unix /var/run/docker.sock: connect: permission denied` when testing the script in a VM.

I fixed this by running `sudo chmod 666 /var/run/docker.sock`.

This may just be a VM issue.

wtgee: Was that after the full install or during the docker pull?

jlibermann: This was during the docker pull.

jlibermann: When the `setup-local-environment` script ran the `docker system prune --force` command.

Issue #988, opened 2020-07-18
Add support for taking dark exposures using a filterwheel equipped with an opaque filter


AnthonyHorton:
**Is your feature request related to a problem? Please describe.**
Some camera types (e.g. CMOS image sensor, interline CCD) lack a internal mechanical shutter or lack the control over the shutter required to take dark exposures. If these cameras are equipped with a filterwheel containing an opaque filter then it is possible to use this block the light in order to take dark exposures, however at present this requires manually instructing the filterwheel to rotate to the opaque filter, taking the dark exposures, then manually instructing the filterwheel to rotate to another filter. This is in contrast to dark exposures taken with a camera's internal shutter, which can be taken by simply passing `dark=true` to the camera's `take_exposure()` or `take_observation()` methods.

**Describe the solution you'd like**
It should be possible to take dark exposures using an opaque filter simply by passing `dark=True` to `take_exposure()`, the same as for dark exposures taken using a camera's internal shutter. Setting `dark=True` should automatically rotate the camera's filterwheel to a named opaque filter. Taking a subsequent exposure with `dark=False` should automatically rotate the camera's filterwheel to the most recently selected non-opaque filter.

**Additional context**
This change has been prompted by the upgrade of the Huntsman Telescope to ZWO CMOS cameras which do not have internal shutters, but do have filterwheels equipped with opaque fllters.

**Related Issues/PRs**
If there are any related Issues or PRs list them here:

wtgee:
Thanks @AnthonyHorton. Looking at the PR, I'm wondering if we shouldn't have some kind of expanded `pre-` and `post-take_exposure` mechanism that is easy for the cameras to override.  Ideally we could get this into some kind of decorator or plugin and ideally that would work across all our hardware types (I've been wanting to create a `PanHardwareBase` that inherits from `PanBase` and that all actual hardware inherits from, and just as soon as I have time...).

That might not be possible immediately but would be good to at least discuss what that might look like.

AnthonyHorton:
I like the idea, but it will take some thought. One thing that would complicate the implementation is that different ways of taking dark exposures require actions at different points, e.g. using an internal mechanical shutter requires setting some shutter control parameter in the camera driver during the camera's `_start_exposure()` method, while for filterwheel darks we need to call filterwheel methods during the camera's `take_exposure()` method before the call to `_start_exposure()`.

AnthonyHorton:
Closed by #989. For now. Will probably revisit this as part of a more general Camera refactorisation.

PR #989, opened 2020-07-18
Add support for filterwheel dark exposures

AnthonyHorton:
Makes it possible to take dark exposures using a camera's filterwheel by passing `dark=True` to the camera's `take_exposure` method, the same as for camera's that take darks using an internal mechanical shutter.

## Description

1. Adds a `can_take_internal_darks` property to the camera classes. If `True` the camera has an internal mechanical shutter and has sufficient control over that shutter to take dark exposures using it. For cameras that can take internal darks this will be the preferred method.
1. Add an optional `dark_position` property to the filterwheel classes, which can be used to specify which filter (if any) is an opaque one suitable for taking dark exposures.
1. Adds a `move_to_dark_position()` method to filterwheels that will rotate to the opaque filter, and a `move_to_light_position()` to revert to the most recently selected non-opaque filter. The `move_to()` method has been modified to keep a record of the most recent non-opaque filter.
1. Modify camera `take_exposure()` method so that if `dark=True` it will first check the `can_take_internal_darks` property, and if that is `False` it will then look for a filterwheel, and check for the `dark_position` property of that. If a filterwheel with opaque filter is present the filterwheel `move_to_dark_position()` method will be called before starting the exposure.
1. Modify camera `take_exposure()` method so that if `dark=False` but there is an attached filterwheel with an opaque filter in position the filterwheel's `move_to_light_position()` method will be called before starting the exposure.

## Related Issue
#988 

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

- [X] I have added tests to cover my changes.


src/panoptes/pocs/camera/camera.py
```diff
@@ -68,28 +68,40 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
...
         readout_time (float): approximate time to readout the camera after an exposure.
         file_extension (str): file extension used by the camera's image data, e.g. 'fits'
         library_path (str): path to camera library, e.g. '/usr/local/lib/libfli.so' (SBIG, FLI, ZWO)
         properties (dict): A collection of camera properties as read from the camera.
+        is_connected (bool): True if camera is connected.
         is_cooled_camera (bool): True if camera has image sensor cooling capability.
         is_temperature_stable (bool): True if image sensor temperature is stable.
         is_exposing (bool): True if an exposure is currently under way, otherwise False.
+        is_ready (bool): True if the camera is ready to take an exposure.
+        can_take_internal_darks (bool): True if the camera can take internal dark exposures.
```
wtgee, line 104: Can we do `darks_enabled` to keep with the `cooling_enabled`?

My first thought was that wouldn't work because what we are sort of saying is we want every camera within POCS to be dark-capable by providing some mechanism (lens cap, slew to home, opaque filter) that allows the to take dark shots.  

However, I still think a `ZWO` camera class would have a `dark_enabled=False` by default, and then when you create particular instances of `ZWO` with an opaque filter, you then can set `dark_enabled=True` (could be set automatically on `init` by testing for filterwheel, etc).

This would change some of the structure below, but I think would be more consistent with the cooling feature of a camera (and again points to that plugin/decorator mechanism).

src/panoptes/pocs/filterwheel/filterwheel.py
```diff
@@ -48,6 +52,11 @@ def __init__(self,
         else:
             self._timeout = timeout
         self._serial_number = serial_number
+        if dark_position:
+            self._dark_position = self._parse_position(dark_position)
+        else:
+            self._dark_position = None
+        self._light_position = None
```
wtgee, line 59: `_dark_position` is static for the instance but `_light_position` is dynamic.  Maybe `_last_light_position` or something to reflect the difference?

src/panoptes/pocs/filterwheel/filterwheel.py
```diff
@@ -212,6 +236,23 @@ def move_to(self, position, blocking=False):
...
+            msg = f"Request to move to dark position but {self} has no dark_position set."
+            self.logger.warning(msg)
+
+    def move_to_light_position(self):
+        """ Return to previous filterwheel position from before taking darks. """
+        if self._light_position:
+            self.move_to(self._light_position)
+        else:
+            msg = f"Request to revert to previous light position but {self} has" + \
+                "no _light_position stored."
```
wtgee, line 253: Let's change to not expose the private variable name in the log message, especially since the user doesn't have the option to change or set that in any way.

src/panoptes/pocs/camera/camera.py
```diff
@@ -68,28 +68,40 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
...
+        uid (str): Unique identifier of the camera.
         is_primary (bool): If this camera is the primary camera for the system, default False.
         model (str): The model of camera, such as 'gphoto2', 'sbig', etc. Default 'simulator'.
         name (str): Name of the camera, default 'Generic Camera'.
         port (str): The port the camera is connected to, typically a usb device, default None.
+        temperature (astropy.units.Quantity): Current temperature of the image sensor.
         target_temperature (astropy.units.Quantity): image sensor cooling target temperature.
         temperature_tolerance (astropy.units.Quantity): tolerance for image sensor temperature.
+        cooling_enabled (bool): True if image snsor cooling is active.
+        cooling_power (float): Current image sensor cooling power level.
```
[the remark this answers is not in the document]
AnthonyHorton: All of the cameras with cooling that we have support for so far has returned this is a percentage, so it would be 0-100.

src/panoptes/pocs/camera/camera.py
```diff
@@ -418,6 +444,21 @@ def take_exposure(self,
 
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
+        if not self.can_take_internal_darks:
+            if dark and self.filterwheel and self.filterwheel.dark_position:
```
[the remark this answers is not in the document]
AnthonyHorton: Right, so we could have several different implementations for "prepare camera to take dark exposures", and a mechanism to select which one should be used for a given set of hardware, so that the base class only needs to make a generic method call instead of covering all possible hardware configurations itself?

src/panoptes/pocs/camera/camera.py
AnthonyHorton: I like the principle, but not entirely happy with the nomenclature.

From my point of view all cameras can take dark exposures, it's just that some need more help than others... It's (almost) always possible to have a human manually put on a lens cap, or equivalent. That's why I did this so that it's still possible to take an exposure with `dark=True` even when there is no supported mechanism for automatically blocking the light. In that case all that `dark=True` does is set the `IMAGETYP` FITS header accordingly, under the assumption that something external is taking care of keeping light out.

wtgee: Right, it's mostly the nomenclature that is a bit odd to me. I definitely like the idea the all cameras could somehow provide darks. I almost suggested something like `has_shutter_control` but then you also have to build some things around it to make it work like you want.

Let's not worry about it too much right now and maybe with some usage and implementation it will evolve.

danjampro:
@AnthonyHorton Looks good! Do we really want to move to call `move_to_light_position()` in take_exposure if `dark=False`? Might a warning be more appropriate?

wtgee:
@AnthonyHorton I pulled this down to my local, merged the upstream, and tests are passing fine.  I didn't actually look over the code again but apparently I was okay with it before. :)

wtgee:
If we could get some coverage on `move_to_dark_position` and `move_to_light_position` should pass the coverage tests.

src/panoptes/pocs/filterwheel/filterwheel.py
wtgee, line 253: Is `dark_position` public and `_light_position` private for any reason?  If not let's make consistent.

src/panoptes/pocs/camera/camera.py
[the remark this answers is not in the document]
AnthonyHorton: All cameras so far actually return this as a `Quantity` with `percent` unit. Correcting the docstring for this.

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: I made `dark_position` into a read only property so there is both `dark_position` and `_dark_position`. At the moment though the only code outside the `FilterWheel` class that accesses this is the parent `Camera`, and I can't think of any other reasons why someone else would need to access it so I guess I'll remove that property.

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: Done.

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: Done.

src/panoptes/pocs/camera/camera.py
```diff
@@ -418,6 +444,21 @@ def take_exposure(self,
...
+                # Can't take internal darks, but do have an opaque filter for that.
+                msg = "Taking dark exposure using filter '" + \
+                    f"{self.filterwheel.filter_name(self.filterwheel.dark_position)}'."
+                self.logger.debug(msg)
+                self.filterwheel.move_to_dark_position(blocking=True)
```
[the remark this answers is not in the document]
AnthonyHorton: Changed all of this to be based entirely on `try .. except` instead of testing truthiness of lots of attributes. Needed to modify both `move_to_dark_position` and `move_to_light_position` to raise exceptions (`error.NotFound`) instead of merely logging a warning about the absence of the set dark position or saved last light position.

src/panoptes/pocs/camera/camera.py
```diff
@@ -447,6 +473,24 @@ def take_exposure(self,
...
         assert filename is not None, self.logger.error("Must pass filename for take_exposure")
 
+        if not self.can_take_internal_darks:
+            if dark:
+                try:
+                    # Can't take internal dark, so try using an opaque filter in a filterwheel
+                    self.filterwheel.move_to_dark_position(blocking=True)
+                    msg = "Taking dark exposure using filter '" + \
+                          f"{self.filterwheel.filter_name(self.filterwheel._dark_position)}'."
+                    self.logger.debug(msg)
```
wtgee, line 483: Not sure how this `msg` syntax keeps slipping in ;)

```suggestion
                    self.logger.debug("Taking dark exposure using filter '"
                                      f"{self.filterwheel.filter_name(self.filterwheel._dark_position)}'.")
```

src/panoptes/pocs/camera/camera.py
```diff
@@ -447,6 +473,24 @@ def take_exposure(self,
...
+                except (AttributeError, error.NotFound):
+                    # No filterwheel, or no opaque filter (dark_position not set)
+                    msg = "Taking dark exposure without shutter or opaque filter. " + \
+                        "Is the lens cap on?"
+                    self.logger.warning(msg)
```
wtgee, line 488:
```suggestion
                    self.logger.warning("Taking dark exposure without shutter or opaque filter. Is the lens cap on?")
```

src/panoptes/pocs/filterwheel/filterwheel.py
```diff
@@ -196,28 +213,53 @@ def move_to(self, position, blocking=False):
             self.logger.error(msg)
             raise error.PanError(msg)
 
-        position = self._parse_position(position)
-        self.logger.info("Moving {} to position {} ({})".format(
-            self, position, self.filter_names[position - 1]))
+        # Will raise a ValueError at this point if new_position is not a valid position
```
wtgee, line 216: Maybe put that in the docstring

src/panoptes/pocs/filterwheel/filterwheel.py
```diff
@@ -196,28 +213,53 @@ def move_to(self, position, blocking=False):
...
+        if new_position == self.position:
             # Already at requested position, don't go nowhere.
+            self.logger.debug(f"{self} already at position {new_position}" + \
+                              f" ({self.filter_name(new_position)})")
             return self._move_event
 
+        if new_position == self._dark_position:
+            # Moving from light into darkness... Store current position so we can revert
+            # back to it if requested with move_to_light_position()
+            self._last_light_position = self.position
```
wtgee, line 228: I just wondering if we should have a `previous_filter` or even `filter_history`. :thinking: A different PR though.

src/panoptes/pocs/filterwheel/filterwheel.py
```diff
@@ -196,28 +213,53 @@ def move_to(self, position, blocking=False):
...
 
         if blocking:
             self._move_event.wait()
 
         return self._move_event
 
+    def move_to_dark_position(self, blocking=False):
+        """ Move to filterwheel position for taking darks. """
+        try:
+            return self.move_to(self._dark_position, blocking=blocking)
```
wtgee, line 243:
```suggestion
            self.logger.info(f'Moving filter to dark position')
            return self.move_to(self._dark_position, blocking=blocking)
```

src/panoptes/pocs/filterwheel/filterwheel.py
```diff
@@ -196,28 +213,53 @@ def move_to(self, position, blocking=False):
...
+        except ValueError:
+            msg = f"Request to move to dark position but {self} has no dark_position set."
+            raise error.NotFound(msg)
+
+    def move_to_light_position(self, blocking=False):
+        """ Return to last filterwheel position from before taking darks. """
+        try:
+            return self.move_to(self._last_light_position, blocking=blocking)
```
wtgee, line 251:
```suggestion
            self.logger.info(f'Moving filter to light position after dark')
            return self.move_to(self._last_light_position, blocking=blocking)
```

src/panoptes/pocs/camera/camera.py
AnthonyHorton: Ah, `msg =` is 18 characters shorter than `self.logger.debug(`, which helps a bit with getting these big format strings within the line limits with fewer line breaks...

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: Added to docstring as well.

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: I think the nomenclature is OK, for now.

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: For a camera that used filterwheel darks this gets called before every single dark exposure (regardless of whether the filterwheel is already at the darkposition), so I'm not sure we really want an INFO level log message here.

src/panoptes/pocs/filterwheel/filterwheel.py
AnthonyHorton: As above, this gets called regardless of whether the filterwheel is already at a light position.

PR #990, opened 2020-07-24
Temperature checking

danjampro:
Cooled cameras need time for their temperature to stabilise. This PR improves how this is handled by POCS by introducing a by-default non-blocking method to `AbstractCamera`, `wait_for_stable_temperature`, which monitors the temperature periodically and sets `camera._is_temperature_stable` to `True` if the temperature is stable for a given minimum period of time. The camera method `is_ready` requires  `camera._is_temperature_stable=True` to return `True` if the camera has cooling enabled.

Also fixes a small bug where `dir` was being used in place of `directory` as a kwarg in `get_free_space`, which was causing tests to fail.

## How Has This Been Tested?
POCS unit tests. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #992, opened 2020-07-27
Bump panoptes-utils

wtgee:
* Fixes the usage of the `dir` built-in in `get_free_space`.

This fix exists in #991 #990 #987 but I'm pushing it through here first.


danjampro:
LGTM.

PR #999, opened 2020-08-28
Gets portal observations from cloud function

jlibermann:
## Description

This is a script that will retrieve scheduled observations from the [PANOPTES Observation Portal](https://github.com/panoptes/panoptes-tom) and output the targets to a YAML file.

## Types of changes

## Checklist:

- [ x] My code follows the code style of this project.
- [ x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/scripts/get-cloud-observations.py
+++ b/scripts/get-cloud-observations.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+
+import click
+import yaml
+
+import subprocess
+
+from panoptes.pocs.utils.logger import get_logger
+from panoptes.utils.serializers import from_json, to_json
+
+logger = get_logger()
+
+
+@click.command()
+@click.option("--cloud-function-name", default="get-tom-observations")
+@click.option("--targets-file", required=True, help="File for writing observation debug.")
+@click.option("--facility", required=True, help="Observation facility name (ie. PAN001)")
+@click.option(
+    "--overwrite", is_flag=True, help="Should targets file be appended to or overwritten."
+)
+def main(
+    targets_file=None, facility=None, cloud_function_name="get-tom-observations", overwrite=False,
+):
+    # Convenience function for printing if verbose.
+    logger.info(f"Calling {cloud_function_name}, writing new targets to {targets_file}")
+
+    # Call the cloud function get json data.
+    new_targets = call_cloud_function(cloud_function_name, facility)
+
+    # Create targets.
+    target_list = list()
+    for target in new_targets:
+        if target["facility"] == facility:
+            logger.info(f"Adding {target}")
+            # Pull out relevant fields.
+            data = target["parameters"]
+
+            ra = data.get("field_ra")
+            dec = data.get("field_dec")
+            position = f"{ra}deg {dec}deg"
+
+            target_info = dict(
+                name=data.get("name"),
+                position=position,
+                priority=data.get("priority", 100),
+                exp_time=data.get("exp_time"),
+                min_nexp=data.get("min_nexp"),
+                exp_set_size=data.get("exp_set_size"),
+            )
+
+            target_list.append(target_info)
+
+    # Write that to fields file.
+    file_mode = "a"
+    if overwrite:
+        file_mode = "w"
+
+    with open(targets_file, file_mode) as f:
+        f.write(yaml.safe_dump(target_list, sort_keys=False,))
+
+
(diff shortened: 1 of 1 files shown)


scripts/get-cloud-observations.py
```diff
@@ -0,0 +1,89 @@
...
+
+from panoptes.pocs.utils.logger import get_logger
+from panoptes.utils.serializers import from_json, to_json
+
+logger = get_logger()
+
+
+@click.command()
+@click.option("--cloud-function-name", default="get-tom-observations")
+@click.option("--targets-file", required=True, help="File for writing observation debug.")
```
wtgee, line 28:
```suggestion
@click.option("--targets-file", required=True, help="Output file for returned observations")
```

scripts/get-cloud-observations.py
```diff
@@ -0,0 +1,89 @@
...
+@click.option("--facility", required=True, help="Observation facility name (ie. PAN001)")
+@click.option(
+    "--overwrite", is_flag=True, help="Should targets file be appended to or overwritten."
```
wtgee, line 31:
```suggestion
    "--overwrite", is_flag=True, help="Should targets file be appended to or overwritten, default False"
```

scripts/get-cloud-observations.py
```diff
@@ -0,0 +1,89 @@
...
+
+            target_list.append(target_info)
+
+    # Write that to fields file.
+    file_mode = "a"
+    if overwrite:
+        file_mode = "w"
+
+    with open(targets_file, file_mode) as f:
+        f.write(yaml.safe_dump(target_list, sort_keys=False,))
```
wtgee, line 71:
```suggestion
        f.write(yaml.safe_dump(target_list, sort_keys=False))
```

scripts/get-cloud-observations.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def call_cloud_function(cloud_function_name, facility):
+    request_info = to_json(dict(facility=facility))
+    cmds = ["gcloud", "functions", "call", cloud_function_name, "--data", request_info]
```
wtgee, line 76: I did this a bit quick and dirty while we were live-coding, but what you would want to do for thoroughness is first check for the correct `gcloud` command using `shutil` from the standard library:

```suggestion
    # at top of file
    import shutil
    ...
    gcloud_cmd = shutil.which('gcloud')
    if gcloud_cmd is None:
        logger.error(f"gcloud command not found on system, can't get observations from the network")
    
    cmds = [gcloud_cmd, "functions", "call", cloud_function_name, "--data", request_info]
```

scripts/get-cloud-observations.py
```diff
@@ -0,0 +1,89 @@
...
+
+    logger.info(f"Calling {cloud_function_name} with {cmds=}")
+
+    completed_process = subprocess.run(cmds, check=True, capture_output=True)
+
+    target_info = completed_process.stdout.decode().lstrip().rstrip().split("\n")
+    new_targets = from_json(target_info[-1])
```
wtgee, line 83: Let's explain what's going on here and hide all the parsing in one line:

```suggestion
    # Get the targets as the last item from the gcloud output.
    target_info = completed_process.stdout.decode().strip().split("\n")[-1]
    new_targets = from_json(target_info)
```

scripts/get-cloud-observations.py
```diff
@@ -0,0 +1,89 @@
...
+            dec = data.get("field_dec")
+            position = f"{ra}deg {dec}deg"
+
+            target_info = dict(
+                name=data.get("name"),
+                position=position,
+                priority=data.get("priority", 100),
+                exp_time=data.get("exp_time"),
+                min_nexp=data.get("min_nexp"),
+                exp_set_size=data.get("exp_set_size"),
```
wtgee, line 60: This might handle the default values oddly.  Might be better to just not include them in the dictionary unless there is a value.

wtgee: We could check this by trying to actually create `astropy.coordinates.SkyCoord` object (or a POCS `Field` and `Observation`) from each `target_info` before you `append` it.  If it fails then just skip it?

It will handle that during the intake on the scheduler side, but it might be nice for the script user to know they downloaded ones that won't work.

wtgee:
Closing as stale at this point. While reintroduce similar concept later.

PR #1003, opened 2020-09-12
WIP: Improved camera error handling for take_exposure & autofocus

### LINKED ISSUE #925: Not possible to enable autofocus plots by config file
**Describe the bug**
At some point the behaviour of autofocus was changed from producing plots to not producing plots by default. This can be controlled via the `make_plots` argument to `pocs.focuser.focuser.AbstractFocuser.autofocus()` but unfortunately no corresponding member variable was added to `autofocus()` and `__init__()`, so unlike all the other autofocus parameters this behaviour cannot be set in the config file. This makes it effectively impossible to generate focus plots during normal observing.

AnthonyHorton:
Improved error handling for the main asynchronous operations of cameras and focusers: taking an exposure and doing autofocus. This has been prompted by the problems caused by unreliable cameras in the Huntsman Telescope. Calling code is unable to detect problems that occur after an exposure or autofocus sequence have started, leading to attempts to continue using a failed camera and, in the case of autofocus, possible deadlock.

## Description

1. Add an `exposure_error` property to the `pocs.camera.camera.AbstractCamera` class. This will be `None` unless an exception is raised during an exposure, at which point it will be set to the string representation of the exception that occurred. Note that exceptions that occur after an exposure has started occur in a `threading.Thread` and so are otherwise invisible to the calling code in the main thread. This property allows code in the main thread to check whether an exception did occur, and respond accordingly.
1. Changes to `AbstractCamera.take_exposure()` to ensure that `exposure_error` gets set if exceptions are raised when starting the exposure.
1. Changes to `AbstractCamera._poll_exposure()` to ensure that `exposure_error` gets set if exceptions are raised either while waiting for the exposure to complete, or if the subsequent call to `._readout()` raises an exception. `._poll_exposure()` and `._readout()` run in a `Thread` so these exceptions would otherwise be invisible to the main thread.
1. Changes to `AbstractCamera.get_thumbnail()` so that it checks `exposure_error` after taking an exposure and raises an `error.PanError` if the exposure failed. Also removed the suppression of exceptions from the call to `img_utils.crop_data()`. Between them these changes are intended to ensure that if anything goes wrong during `get_thumbnail()` a relevant, descriptive exception will be raised.
1. Add an `autofocus_error` property to the `pocs.focuser.focuser.AbstractFocuser` class. This will be `None` unless an exception is raised during an autofocus sequence, at which point it will be set to the string representation of the exception that occurred. Note that exceptions that occur during an autofocus sequence occur in a `threading.Thread` and so are otherwise invisible to the calling code in the main thread. This property allows code in the main thread to check whether an exception did occur, and responding accordingly.
1. Changes to `AbstractFocuser._autofocus()` so that all calls to `self._camera.get_thumbnail()` (which will now raise an exception in the case of a failed exposure) are wrapped in `try ... except` blocks which will set `autotocus_error`, set the focus `threading.Event` and re-raise the exception, terminating the `Thread` that `_autofocus()` runs in. Previously exceptions from `get_thumbnail()` were not handled and would terminate the thread without setting the `Event`, relying on the calling code to implement a timeout to prevent deadlocks.
1. Make it possible to set the autofocus option `make_plots` when creating the focuser via a new `autofocus_make_plots` constructor argument. This makes it possible to enable or disable plots using the config. These plots are very useful for troubleshooting but currently there is no way to turn them on.

## Related Issue

Incidentally fixes #925 along the way, but the main focus of this PR relates to https://github.com/AstroHuntsman/huntsman-pocs/issues/205

## How Has This Been Tested?

So far, not at all.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/panoptes/pocs/camera/camera.py
+++ b/src/panoptes/pocs/camera/camera.py
@@ -147,6 +147,7 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         self._exposure_event = threading.Event()
         self._exposure_event.set()
         self._is_exposing = False
+        self._exposure_error = None
 
         # By default assume camera isn't capable of internal darks.
         self._internal_darks = kwargs.get('internal_darks', False)
@@ -384,6 +385,11 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         """
         return self._internal_darks
 
+    @property
+    def exposure_error(self):
+        """ Error message from the most recent exposure or None, if there was no error."""
+        return self._exposure_error
+
     ##################################################################################################
     # Methods
     ##################################################################################################
@@ -469,9 +475,19 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
             threading.Event: Event that will be set when exposure is complete.
 
         """
-        assert self.is_connected, self.logger.error("Camera must be connected for take_exposure!")
+        self._exposure_error = None
 
-        assert filename is not None, self.logger.error("Must pass filename for take_exposure")
+        if not self.is_connected:
+            err = AssertionError("Camera must be connected for take_exposure!")
+            self.logger.error(str(err))
+            self._exposure_error = repr(err)
+            raise err
+
+        if not filename:
+            err = AssertionError("Must pass filename for take_exposure")
+            self.logger.error(str(err))
+            self._exposure_error = repr(err)
+            raise err
 
         if not self.can_take_internal_darks:
             if dark:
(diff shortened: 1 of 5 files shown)


AnthonyHorton:
I'm not actually expecting this to get merged, at least not in its current form. The `exposure_error` and `autofocus_error` properties are workarounds for the error handling limitations resulting from using `threading.Thread` and `threading.Event` for asynchronous operations. I think that it would be better to rework all of this to use `concurrent.futures` instead of `threading`which would, amongst other benefits, enable exceptions that occur in threads to be automatically re-raised in the main thread. That's a much cleaner solution than adding lots of extra code to catch, store the string representation, and re-raise exceptions. Switching to `concurrent.futures` would be a much more invasive change though, and Huntsman needs some quick fixes, so I created this branch to implement the ugly hack solution. This draft PR exists to get some feedback.

AnthonyHorton:
Should probably make most if not all of `pocs.camera.camera.AbstractCamera.process_exposure()` conditional on `exposure_error` and/or the FITS file actually existing at the expected path. Currently, following a failed exposure, `process_exposure()` will write a number of warnings to the logs as a result of trying to do things with a non-existent FITS files that really only obscure the root cause of the problem.

wtgee:
Most of this is making it's way into https://github.com/panoptes/POCS/pull/1007

PR #1009, opened 2020-10-07
Updates to take_observation

### LINKED ISSUE #1002: Camera take_observation() method fails if optional headers argument not given
**Describe the bug**

`pocs.camera.camera.AbstractCamera.take_observation()` raises a `TypeError` if called on the primary camera without setting the `headers` argument. `headers` is supposed to be an optional argument, & defaults to `None`, but `take_exposure()` contains https://github.com/panoptes/POCS/blob/1f74a94a7fc5e81745eb786b8e5789d8c7da8f6c/src/panoptes/pocs/camera/camera.py#L432, which fails if `headers` is not an iterable.

This could be fixed either by making `headers` default to an empty dictionary, or changing `if 'POINTING' in headers:` to `if 'POINTING' in metadata:` The call to `_setup_observation()` 12 lines earlier generates `metadata` and adds the contents of `headers` to it so if 'POINTING' was in `headers` it will also be in `metadata`.

**To Reproduce**
1. Create a POCS `Observation` object, and a primary `Camera` object.
1. Call the `Camera`'s `take_observation()` method with the `Observation` as the only argument.
1. `TypeError` is raised.

**Expected behavior**
`take_observation()` can be called without setting `headers` without raising an exception.

wtgee:
`take_observation` 
  * `headers` param fixed so truly optional.  The `POINTING` keyword is checked in the metadata, not original headers.
  * Passing approved headers will actually write them to file.
  * `blocking=False` param added. If `True`, will wait on `observation_event`.
* Renamed metadata variables to be consistent.
* `_process_fits` is responsible for writing the headers rather than calling out to `panoptes-utils`. Allows for easier overrides.
* dslr simulator readout time improved.
* `process_exposure` doesn't require the `exposure_event` to be passed because that is the cameras `is_exposing` property.

* Config options changed to control camera processing options and allow for `defaults` in the config that applies to all cameras: 
  * `cameras.defaults.compress_fits` if FITS files should be fpacked. Default True.
  * `cameras.defaults.record_observations` if observation metadata should be recorded. Default True.
  * `cameras.defaults.make_pretty_images` to make jpgs from images. Default True.


src/panoptes/pocs/camera/camera.py
```diff
@@ -580,31 +585,32 @@ def process_exposure(self,
...
                 img_utils.make_pretty_image(file_path,
                                             title=image_title,
                                             link_path=link_path)
             except Exception as e:  # pragma: no cover
-                self.logger.warning('Problem with extracting pretty image: {e!r}')
+                self.logger.warning(f'Problem with extracting pretty image: {e!r}')
 
         with suppress(Exception):
-            info['exptime'] = info['exptime'].value
+            metadata['exptime'] = metadata['exptime'].value
```
danjampro, line 609: Just checking that we shouldn't be using `get_quantity_value` here?

src/panoptes/pocs/camera/camera.py
```diff
@@ -921,20 +927,53 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
             metadata['filter_request'] = observation.filter_name
 
         if headers is not None:
+            self.logger.trace(f'Updating {file_path} metadata with provided headers')
             metadata.update(headers)
 
-        self.logger.debug(
-            f'Observation setup: exptime={exptime} file_path={file_path} image_id={image_id} '
-            f'metadata={metadata}')
+        self.logger.debug(f'Observation setup: {exptime=} {file_path=} {image_id=} {metadata=}')
```
danjampro, line 933: I assume this is deliberate? New Python syntax?

src/panoptes/pocs/camera/camera.py
```diff
@@ -559,19 +562,21 @@ def process_exposure(self,
...
             FileNotFoundError: If the FITS file isn't at the specified location.
         """
+        # Wait for exposure to complete. Timeout handled by exposure thread.
+        while self.is_exposing:
+            time.sleep(1)
+
+        self.logger.debug(f'Starting exposure processing for {observation_event}')
+
         # If passed an Event that signals the end of the exposure wait for it to be set
         compress_fits = compress_fits or self.get_config('observations.compress_fits')
```
danjampro, line 572: I'm not sure this way (using `or`) of doing things is such a good idea - I would expect keyword arguments to override config values. This may lead to some confusion.

src/panoptes/pocs/camera/camera.py
```diff
@@ -383,7 +383,9 @@ def take_observation(self, observation, headers=None, filename=None, **kwargs):
                 describing the observation
             headers (dict, optional): Header data to be saved along with the file.
```
danjampro, line 384: Is this true? It seems like only selected headers are written to the file. We might want to make this clear, and acceptable headers should be documented somewhere.

src/panoptes/pocs/camera/camera.py
```diff
@@ -921,20 +927,53 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
...
+            'camera_uid': {'keyword': 'INSTRUME', 'comment': 'Camera ID'},
+            'observer': {'keyword': 'OBSERVER', 'comment': 'PANOPTES Unit ID'},
+            'origin': {'keyword': 'ORIGIN'},
+            'tracking_rate_ra': {'keyword': 'RA-RATE', 'comment': 'RA Tracking Rate'},
+        }
+
+        self.logger.debug(f"Updating FITS headers: {file_path} with {metadata=}")
+        with fits.open(file_path, 'update') as f:
+            hdu = f[0]
+            for metadata_key, field_info in fields.items():
```
danjampro, line 967: So only certain fields will be written to the file. This seems like fine default behaviour, but if I passed some non-standard headers to `take_observation`, I would expect them to be written to file, rather than to be ignored without warning.

src/panoptes/pocs/camera/camera.py
```diff
@@ -921,20 +927,53 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
...
 
-        self.logger.debug(
-            f'Observation setup: exptime={exptime} file_path={file_path} image_id={image_id} '
-            f'metadata={metadata}')
+        self.logger.debug(f'Observation setup: {exptime=} {file_path=} {image_id=} {metadata=}')
+
         return exptime, file_path, image_id, metadata
 
-    def _process_fits(self, file_path, info):
+    def _process_fits(self, file_path, metadata):
```
danjampro, line 937: If the only purpose of this function is to add the fits headers, I suggest to rename to `_add_fits_headers`.

src/panoptes/pocs/camera/camera.py
```diff
@@ -559,19 +562,21 @@ def process_exposure(self,
...
-        # Wait for exposure to complete.
-        while self.is_exposing:
-            time.sleep(1)
-
-        image_id = info['image_id']
-        seq_id = info['sequence_id']
-        file_path = info['file_path']
-        exptime = info['exptime']
-        field_name = info['field_name']
+        image_id = metadata['image_id']
```
danjampro, line 575: One way to clean this up a bit (and make it more flexible) would be to offload these mappings to a `header_mappings.yaml` file. Same goes for `_process_exposure`.

src/panoptes/pocs/camera/camera.py
```diff
@@ -921,20 +927,53 @@ def _setup_observation(self, observation, headers, filename, **kwargs):
...
+            'filter': {'keyword': 'FILTER'},
+            'latitude': {'keyword': 'LAT-OBS', 'comment': 'Degrees'},
+            'longitude': {'keyword': 'LONG-OBS', 'comment': 'Degrees'},
+            'elevation': {'keyword': 'ELEV-OBS', 'comment': 'Meters'},
+            'moon_separation': {'keyword': 'MOONSEP', 'comment': 'Degrees'},
+            'moon_fraction': {'keyword': 'MOONFRAC'},
+            'creator': {'keyword': 'CREATOR', 'comment': 'POCS Software version'},
+            'camera_uid': {'keyword': 'INSTRUME', 'comment': 'Camera ID'},
+            'observer': {'keyword': 'OBSERVER', 'comment': 'PANOPTES Unit ID'},
+            'origin': {'keyword': 'ORIGIN'},
```
danjampro, line 960: I wonder if we can add the alt/az of the mount here, `ALT-MNT` and `AZ-MNT`? This would be super useful for vignetting tests.

danjampro (commented):
@wtgee Some comments for your consideration below.  Looks good!

src/panoptes/pocs/camera/camera.py
wtgee: I've created #1013 to address this, but yes, it's an annoying problem.

src/panoptes/pocs/camera/camera.py
wtgee: That's what this does.  The value from the first argument (`compress_fits`) will be used unless it's value is `None`, in which case the `or` will use the second value (the config lookup).

I've started using this `param=None` then `param = param or something_else` syntax quite a bit and think it's actually a lot cleaner but am open to discussion if others find it confusing.

src/panoptes/pocs/camera/camera.py
wtgee: Yeah, I was just being lazy with this PR and didn't want it to grow too much. Will make issue.

src/panoptes/pocs/camera/camera.py
wtgee: This is from python > 3.7 and yes, have been using in a lot of places.

src/panoptes/pocs/camera/camera.py
wtgee: All of the concrete camera classes are supposed to override this and they do more elaborate things than just adding FITS (potentially at least).

Agree that the name and structure need to change but that's part of the larger refactoring @AnthonyHorton and I always want to get to.

src/panoptes/pocs/camera/camera.py
wtgee: The alt and az can be derived from the information here so I wasn't including it. We would just want some simple reversal of `altaz_to_radec`.  

I wouldn't have an issue adding it here but I'm not going to as part of this PR.  Mo' better would be to allow arbitrary keyword headers, as mentioned elsewhere.

src/panoptes/pocs/camera/camera.py
wtgee: Agreed, but outside the scope of this PR unfortunately.

wtgee: The loop could be rewritten as:

```py
            for metadata_key, metadata_value in metadata.items():
                try:
                    field_info = fields[metadata_key]
                    # for metadata_key, field_info in fields.items():
                    fits_key = field_info['keyword']
                    fits_comment = field_info.get('comment', '')
                    # Get the value from either the metadata, the default, or use blank string.
                    fits_value = metadata.get(metadata_key, field_info.get('default', ''))

                    self.logger.trace(f'Setting {fits_key=} = {fits_value=} {fits_comment=}')
                except KeyError:
                    self.logger.trace(f'No mapping found for metadata {metadata_key}'
                                      f', will attempt to add as-is')
                    fits_key = metadata_key
                    fits_comment = ''
                    fits_value = metadata_value

                hdu.header.set(fits_key, fits_value, fits_comment)
```

That allows random things to go into the fits headers although does no check for a valid fits header. Astropy will automatically create it with a `HIERARCH` card, which I don't know much about.

src/panoptes/pocs/camera/camera.py
wtgee: Basically, `None or 42` resolves to `42`.

danjampro: @wtgee what if (for example) `compress_fits=False` is passed to the function, but the config has `observations.compress_fits=True`. In this case, the keyword argument is ignored in favour of the config value?

danjampro (approved):
@wtgee See my response in our discussion about keyword / config parameters. Other than that, LGTM.

src/panoptes/pocs/camera/camera.py
wtgee: Ah, good one. I'm not sure why I have missed that before. Now I need to check some other code...

Thanks!

Issue #1014, opened 2020-10-09
Remove hard-coded mappings in camera class


wtgee:
One way to clean this up a bit (and make it more flexible) would be to offload these mappings to a `header_mappings.yaml` file. Same goes for `_process_exposure`.

_Originally posted by @danjampro in https://github.com/panoptes/POCS/pull/1009#r502094523_

wtgee:
https://github.com/panoptes/POCS/pull/1009/files#r502606617

PR #1015, opened 2020-10-12
Allow arbitrary FITS headers

### LINKED ISSUE #1013: Document allowed FITS headers (or allow any)
Is this true? It seems like only selected headers are written to the file. We might want to make this clear, and acceptable headers should be documented somewhere.

_Originally posted by @danjampro in https://github.com/panoptes/POCS/pull/1009#r502092711_

wtgee:
Allows for adding custom FITS headers from the `headers` param of `take_observation`.  Follow up from #1009 .

## Description
The `take_observation` function accepts a `headers` dict. Previously the keywords in this dict that matched a predefined (and hard-coded, see #1014) pattern would be included in the final FITS image.  

This PR essentially copies everything from `headers` into the header. It will still parse the pre-defined list of headers (e.g. `field_name` -> `FIELD`).  

Keywords in `headers` that don't conform to the FITS standard will be added as `HIERARCH` (see the [`set`(https://docs.astropy.org/en/stable/io/fits/api/headers.html#astropy.io.fits.Header.set)] method for headers; sadly it's not documented that I found, but it's in the source).

## Related Issue
Closes #1013 
From #1009 

## How Has This Been Tested?
Added a new test specifically for the arbitrary headers.

## Types of changes


wtgee:
Oops, based this off `origin` instead of `upstream`.  Closing for now and will reopen.

PR #1021, opened 2020-10-23
Python 3.7 requirement

wtgee:
* Moves the python minimum back to 3.7 following `panoptes-utils v0.2.29`.
* Testing suite returns from `docker` correctly so GHA will fail when tests fail.


AnthonyHorton (approved):
LGTM!

Issue #1025, opened 2020-10-26
SDK cameras no longer working


AnthonyHorton:
**Describe the bug**
Changes to the camera base class have stopped the SDK cameras (ZWO, FLI, SBIG) from working.

**To Reproduce**
1. Attempt an instance of an SDK `Camera` class
2. See errors due to attempts to assign to read-only `is_exposing` property

**Expected behavior**
SDK camera instances can be created and used.

**Additional context**
The root cause of this problem is changes to the camera base class (and DSLR camera class, and camera simulator) to use the `is_exposing` property as an alias for `_is_exposing_event,is_set()`, and, more specifically, using assignment to `is_exposing` to set/clear this event. For the DSLRs this property is read only, and directly tracks the internal state of the camera hardware.

wtgee:
Closed by #1026. Further problems should open a new Issue.

PR #1026, opened 2020-10-26
Fix SDK camera problems resulting from changes to is_exposing/_is_exposing_event

AnthonyHorton:
Minimal set of changes to get SDK cameras (ZWO, FLI, SBIG) working again following changes to the camera base class, DSLR cameras and simulators.

## Descriptiion

* Removes the setter for the `is_exposing` property from the camera base class. This caused problems for the SDK cameras, where`is_exposing` is a read only property that directly queries the camera internal state using the camera exposure status query function of the relevant SDK.
* Replace all instances of assignment to `is_exposing` with `set()`/`clear()` of the `_is_exposing_event`. For DSLR cameras this is equivalent because `is_exposing` is just an alias for `_is_exposing_event.is_set()`.
* Remove attempts to pass a logger to `panoptes.utils.images.fits.write_fits()` by the ZWO and SBIG camera `_readout()` methods. This function no longer accepts a logger, and the logger that was being passed was being interpreted as a `threading.Event` to be set after writing the file, resulting in an error.

## Related Issue

#1025

## How Has This Been Tested?

ZWO camera class tested with a real camera, plus usual unit tests.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


wtgee (approved):
LGTM, thanks for working through that! :camera_flash:

PR #1029, opened 2020-11-07
Autofocus plots refactor

### LINKED ISSUE #846: Hanging camera if autofocus_size is too large
**Describe the bug**
If you set the `autofocus_size` larger than the image it will hang until you manually kill. Then you see an AssertionError.

**Expected behavior**
Handle the error gracefully.

```
In [4]: cam.autofocus()
Out[4]: <threading.Event at 0x7f30c4366fd0>

In [5]: exit
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/storage/miniconda3/envs/panoptes/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/storage/miniconda3/envs/panoptes/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/storage/panoptes/POCS/pocs/focuser/focuser.py", line 353, in _autofocus
    dark=True)
  File "/storage/panoptes/POCS/pocs/camera/camera.py", line 468, in get_thumbnail
    thumbnail = img_utils.crop_data(image, box_width=thumbnail_size)
  File "/storage/panoptes/POCS/pocs/utils/images/__init__.py", line 57, in crop_data
    box_width, data.shape)
AssertionError: Can't clip data, it's smaller than 5000 ((2532, 3352))
```

### LINKED ISSUE #925: Not possible to enable autofocus plots by config file

### LINKED ISSUE #1028: Memory leak from focuser._autofocus with plots
If an autofocus is run with `make_plots=True`, the resulting plot does not release its allocated memory at the end of the sequence. This has been causing out of memory errors for the Huntsman raspberry pi systems.

The code: https://github.com/panoptes/POCS/blob/dc330ac98fcc01f6cd70085f50d985321c42c58e/src/panoptes/pocs/focuser/focuser.py#L562

Suggested change (tested on hardware):
```
import gc

plt.cla()
plt.clf()
plt.close(fig)
gc.collect()
```

wtgee:
* Move the autofocus plotting to a separate file.
* Super dooper explicitly close the figure (closes #1028).
* Logging call cleanups.
* Add default param to config file for making autofocus plots (closes #925).
* Make sure thumbnail is not bigger than image (closes #846 ).
* **Change output filenames for consistency**

## Related Issue
Closes #1028 
Closes #925 
Closes #846 

## How Has This Been Tested?
It hasn't!!  We currently don't save the focus images or data in a way that would allow this to be easily reproduced.

### DIFF
--- a/src/panoptes/pocs/focuser/focuser.py
+++ b/src/panoptes/pocs/focuser/focuser.py
@@ -297,7 +297,7 @@ class AbstractFocuser(PanBase, metaclass=ABCMeta):
             'seconds': seconds,
             'focus_range': focus_range,
             'focus_step': focus_step,
-            'thumbnail_size': thumbnail_size,
+            'cutout_size': thumbnail_size,
             'keep_files': keep_files,
             'take_dark': take_dark,
             'merit_function': merit_function,
@@ -320,7 +320,7 @@ class AbstractFocuser(PanBase, metaclass=ABCMeta):
                    seconds,
                    focus_range,
                    focus_step,
-                   thumbnail_size,
+                   cutout_size,
                    keep_files,
                    take_dark,
                    merit_function,
@@ -349,35 +349,35 @@ class AbstractFocuser(PanBase, metaclass=ABCMeta):
 
         self._autofocus_error = None
 
-        dark_thumbnail = None
+        dark_cutout = None
         if take_dark:
             dark_path = os.path.join(file_path_root, f'dark.{self._camera.file_extension}')
             self.logger.debug(f'Taking dark frame {dark_path} on camera {self._camera}')
             try:
-                dark_thumbnail = self._camera.get_thumbnail(seconds,
-                                                            dark_path,
-                                                            thumbnail_size,
-                                                            keep_file=True,
-                                                            dark=True)
+                dark_cutout = self._camera.get_thumbnail(seconds,
+                                                         dark_path,
+                                                         cutout_size,
+                                                         keep_file=True,
+                                                         dark=True)
                 # Mask 'saturated' with a low threshold to remove hot pixels
-                dark_thumbnail = mask_saturated(dark_thumbnail,
-                                                threshold=0.3,
-                                                bit_depth=self.camera.bit_depth)
+                dark_cutout = mask_saturated(dark_cutout,
+                                             threshold=0.3,
+                                             bit_depth=self.camera.bit_depth)
             except Exception as err:
                 self.logger.error(f"Error taking dark frame: {err!r}")
                 self._autofocus_error = repr(err)
                 focus_event.set()
                 raise err
 
-        # Take an image before focusing, grab a thumbnail from the centre and add it to the plot
+        # Take an image before focusing, grab a cutout from the centre and add it to the plot
         initial_fn = f"{initial_focus}-{focus_type}-initial.{self._camera.file_extension}"
         initial_path = os.path.join(file_path_root, initial_fn)
 
         try:
-            initial_thumbnail = self._camera.get_thumbnail(seconds, initial_path, thumbnail_size, keep_file=True)
-            if dark_thumbnail is not None:
-                initial_thumbnail = initial_thumbnail - dark_thumbnail
-            initial_thumbnail = mask_saturated(initial_thumbnail, bit_depth=self.camera.bit_depth)
+            initial_cutout = self._camera.get_thumbnail(seconds, initial_path, cutout_size, keep_file=True)
+            initial_cutout = mask_saturated(initial_cutout, bit_depth=self.camera.bit_depth)
+            if dark_cutout is not None:
+                initial_cutout = initial_cutout.astype(np.int32) - dark_cutout
         except Exception as err:
             self.logger.error(f"Error taking initial image: {err!r}")
             self._autofocus_error = repr(err)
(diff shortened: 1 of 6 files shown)


wtgee:
I tried to clean up the exceptions some but it wasn't immediately working and I think we'd prefer a switch to `concurrent.futures` eventually, so leaving for now.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -343,49 +340,44 @@ def _autofocus(self,
...
+                dark_thumbnail = self._camera.get_thumbnail(seconds,
+                                                            dark_path,
+                                                            thumbnail_size,
+                                                            keep_file=True,
+                                                            dark=True)
                 # Mask 'saturated' with a low threshold to remove hot pixels
-                dark_thumb = mask_saturated(dark_thumb,
-                                            threshold=0.3,
-                                            bit_depth=self.camera.bit_depth)
+                dark_thumbnail = mask_saturated(dark_thumbnail,
```
danjampro, line 363: Do we actually need/want to be masking parts of the dark image? Would it be more appropriate to apply this once to the dark-subtracted frame?

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -343,49 +340,44 @@ def _autofocus(self,
...
-            self.logger.debug('Taking dark frame {} on camera {}'.format(dark_path, self._camera))
+            dark_path = os.path.join(file_path_root, f'dark.{self._camera.file_extension}')
+            self.logger.debug(f'Taking dark frame {dark_path} on camera {self._camera}')
             try:
-                dark_thumb = self._camera.get_thumbnail(seconds,
-                                                        dark_path,
-                                                        thumbnail_size,
-                                                        keep_file=True,
-                                                        dark=True)
+                dark_thumbnail = self._camera.get_thumbnail(seconds,
```
danjampro, line 357: I kind of prefer the word "subframe" over "thumbnail" - but hey.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -343,49 +340,44 @@ def _autofocus(self,
...
-                                          focus_type,
-                                          "initial",
-                                          self._camera.file_extension)
+        initial_fn = f"{initial_focus}-{focus_type}-initial.{self._camera.file_extension}"
         initial_path = os.path.join(file_path_root, initial_fn)
 
         try:
             initial_thumbnail = self._camera.get_thumbnail(seconds, initial_path, thumbnail_size, keep_file=True)
+            if dark_thumbnail is not None:
+                initial_thumbnail = initial_thumbnail - dark_thumbnail
```
danjampro, line 379: This is where we can have a potential integer underflow.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -401,51 +393,57 @@ def _autofocus(self,
...
+        # Get focus steps.
         focus_positions = np.arange(max(initial_focus - focus_range / 2, self.min_position),
                                     min(initial_focus + focus_range / 2, self.max_position) + 1,
                                     focus_step, dtype=np.int)
         n_positions = len(focus_positions)
 
-        thumbnails = np.zeros((n_positions, thumbnail_size, thumbnail_size),
-                              dtype=initial_thumbnail.dtype)
+        # Set up empty array holders
+        thumbnails = np.zeros((n_positions, thumbnail_size, thumbnail_size), dtype=initial_thumbnail.dtype)
```
danjampro, line 403: Do we need to keep all the thumbnails in memory? Can we just keep the one with the best focus metric?

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -343,49 +340,44 @@ def _autofocus(self,
...
         # Take an image before focusing, grab a thumbnail from the centre and add it to the plot
-        initial_fn = "{}_{}_{}.{}".format(initial_focus,
-                                          focus_type,
-                                          "initial",
-                                          self._camera.file_extension)
+        initial_fn = f"{initial_focus}-{focus_type}-initial.{self._camera.file_extension}"
         initial_path = os.path.join(file_path_root, initial_fn)
 
         try:
             initial_thumbnail = self._camera.get_thumbnail(seconds, initial_path, thumbnail_size, keep_file=True)
```
danjampro, line 377: We can probably clean up this initial thumbnail code by moving it to the main focusing loop below.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -401,51 +393,57 @@ def _autofocus(self,
...
                 self.logger.error(f"Error taking image {i + 1}: {err!r}")
                 self._autofocus_error = repr(err)
                 focus_event.set()
                 raise err
 
             masks[i] = mask_saturated(thumbnail, bit_depth=self.camera.bit_depth).mask
-            if dark_thumb is not None:
-                thumbnail = thumbnail - dark_thumb
+            if dark_thumbnail is not None:
+                thumbnail = thumbnail - dark_thumbnail
```
danjampro, line 427: Potential integer underflow here.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -470,100 +470,68 @@ def _autofocus(self,
...
-        final_fn = f"{final_focus}_{focus_type}_final.{self._camera.file_extension}"
+        # Get final thumbnail.
+        final_fn = f"{final_focus}-{focus_type}-final.{self._camera.file_extension}"
         file_path = os.path.join(file_path_root, final_fn)
         try:
-            final_thumbnail = self._camera.get_thumbnail(
-                seconds, file_path, thumbnail_size, keep_file=True)
+            final_thumbnail = self._camera.get_thumbnail(seconds, file_path, thumbnail_size, keep_file=True)
+            if dark_thumbnail is not None:
+                final_thumbnail = final_thumbnail - dark_thumbnail
```
danjampro, line 503: Potential integer underflow.

danjampro (changes requested):
@wtgee  Looks good. Might be cleaner to move the initial thumbnail code into the main loop. A few comments regarding integer underflow from dark subtraction that we need to address.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -401,51 +393,57 @@ def _autofocus(self,
...
             thumbnails[i] = thumbnail
 
+        self.logger.debug(f'Making master mask with binary dilation for {self._camera}')
         master_mask = masks.any(axis=0)
         master_mask = binary_dilation(master_mask, iterations=mask_dilations)
 
         # Apply the master mask and then get metrics for each frame.
         for i, thumbnail in enumerate(thumbnails):
+            self.logger.debug(f'Applying focus metric to thumbnail {i:02d}')
             thumbnail = np.ma.array(thumbnail, mask=master_mask)
```
AnthonyHorton, line 437: This is a bug. `thumbnail` will in general already have a mask, the bad pixel mask inherited from `dark_thumbnail` during dark subtraction. Instead of just replacing that with `master_mask` (which removes the bad pixel masking) we should be combining `master_mask` with the existing mask.

```suggestion
            thumbnail = np.ma.array(thumbnail, mask=np.ma.mask_or(master_mask, np.ma.getmask(thumbnail)))
```

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -456,9 +454,11 @@ def _autofocus(self,
             # Fit data around the maximum value to determine best focus position.
             # Initialise models
             shift = models.Shift(offset=-focus_positions[imax])
+            # TODO document magic coefficient numbers.
```
AnthonyHorton, line 457: You mean the initial values? Basically they are just very small values with the expected sign, to help the fitting set off in the correct direction.

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -343,49 +340,44 @@ def _autofocus(self,
...
-                                          "initial",
-                                          self._camera.file_extension)
+        initial_fn = f"{initial_focus}-{focus_type}-initial.{self._camera.file_extension}"
         initial_path = os.path.join(file_path_root, initial_fn)
 
         try:
             initial_thumbnail = self._camera.get_thumbnail(seconds, initial_path, thumbnail_size, keep_file=True)
+            if dark_thumbnail is not None:
+                initial_thumbnail = initial_thumbnail - dark_thumbnail
+            initial_thumbnail = mask_saturated(initial_thumbnail, bit_depth=self.camera.bit_depth)
```
AnthonyHorton, line 380: Saturation masking should be done before dark subtraction, not afterwards.

src/panoptes/pocs/focuser/focuser.py
AnthonyHorton: Bound to be underflow.
```suggestion
                final_thumbnail = final_thumbnail.astype(np.int32) - dark_thumbnail
```

src/panoptes/pocs/focuser/focuser.py
```diff
@@ -470,100 +470,68 @@ def _autofocus(self,
...
+        # Get final thumbnail.
+        final_fn = f"{final_focus}-{focus_type}-final.{self._camera.file_extension}"
         file_path = os.path.join(file_path_root, final_fn)
         try:
-            final_thumbnail = self._camera.get_thumbnail(
-                seconds, file_path, thumbnail_size, keep_file=True)
+            final_thumbnail = self._camera.get_thumbnail(seconds, file_path, thumbnail_size, keep_file=True)
+            if dark_thumbnail is not None:
+                final_thumbnail = final_thumbnail - dark_thumbnail
+            final_thumbnail = mask_saturated(final_thumbnail, bit_depth=self.camera.bit_depth)
```
AnthonyHorton, line 504: Saturation masking should be done before dark subtraction, not afterwards.

AnthonyHorton (approved):
Changes look good, but there are some other bugs that we should fix at the same time though (e.g. integer underflow, mask combining).

src/panoptes/pocs/focuser/focuser.py
wtgee: I changed most of the instances within the autofocus code but did not change the camera `get_thumbnail` name.

src/panoptes/pocs/focuser/focuser.py
wtgee: This takes a picture before any movement of the focuser, whereas the first step in the loop is to move to a new position.

You're right that it probably could be cleaned up a bit but I'm going to leave for now.

src/panoptes/pocs/focuser/focuser.py
wtgee: Wow, how did I now know about `mask_or`..Thanks!

PR #1043, opened 2020-12-11
State machine safety checks

danjampro:
## Description
Modifies state machine behaviour to check for state safety between state transitions *only* if state is not always safe.

## How Has This Been Tested?
Unit tests and tested on Huntsman itself.

## Types of changes


wtgee (commented):
I meant to submit this a day or so ago, my apologies.  It looks like a few things have changed so I'm submitting now and will take a look again.

src/panoptes/pocs/state/machine.py
```diff
@@ -109,25 +109,41 @@ def run(self, exit_when_done=False, run_once=False, initial_next_state='ready'):
...
+
+                # Check the horizon here because next state may have changed in loop
+                required_horizon = self._horizon_lookup.get(self.next_state, 'observe')
+                if self.is_dark(horizon=required_horizon):
+                    break
+                self.logger.info(f"Waiting for required_horizon={required_horizon!r} for "
+                                 f"self.next_state={self.next_state!r}")
+
+                # Sleep before checking again
+                check_delay = self.get_config('wait_delay', default=120)
```
wtgee, line 136: This does a lookup in the config server so unless we want to dynamically change it this should be moved outside the while loop ideally. Not critical.

src/panoptes/pocs/state/machine.py
```diff
@@ -287,12 +305,10 @@ def after_state(self, event_data):
     @classmethod
     def load_state_table(cls, state_table_name='simple_state_table'):
         """ Loads the state table
-
```
wtgee, line 290: These are the weird line options that we should figure out for our editors because they seem to be inconsistent.

src/panoptes/pocs/tests/test_pocs.py
```diff
@@ -179,31 +202,6 @@ def test_is_weather_safe_no_simulator(pocs):
...
-    pocs.get_ready()
-    assert pocs.state == 'ready'
-    pocs.schedule()
-    assert pocs.state == 'scheduling'
-
-    # My time goes fast...
-    os.environ['POCSTIME'] = '2020-01-01 18:00:00'
-    pocs.set_config('simulator', hardware.get_all_names(without=['night']))
-
-    assert pocs.is_safe() is False
```
wtgee, line 197: Why is this test removed? Could we change this call using the new `ignore` param and make test work?

src/panoptes/pocs/tests/test_pocs.py
```diff
@@ -40,6 +40,16 @@ def mount():
     return create_mount_simulator()
 
 
+@pytest.fixture(scope='function')
```
wtgee, line 43: Thanks for the cleanup,with the fixtures, much nicer!

wtgee (approved):
Looks good, thanks for adding the changes to make it  more generic, I think that will work well.

A few questions and suggestions. I'll approve, @danjampro let me know when you think ready to merge.

src/panoptes/pocs/tests/test_pocs.py
danjampro: I think this test is pretty much broken, but I've replaced it with another `test_unsafe_park` which tests the same functionality.

PR #1044, opened 2020-12-13
Update Bisque mount status

danjampro:
The Bisque mount status properties (`is_parked`, `is_tacking`, `is_slewing` etc) do not get updated unless `mount._update_status()` is explicitly called. This led to some errors when calling `slewing_to_target`, which was not blocking properly. The fix is to override the properties in `bisque.Mount` to automatically update the status in the properties.

## How Has This Been Tested?
Unit tests and on Huntsman.

## Types of changes

### DIFF
--- a/src/panoptes/pocs/mount/bisque.py
+++ b/src/panoptes/pocs/mount/bisque.py
@@ -212,7 +212,7 @@ class Mount(AbstractMount):
                     'ra': mount_coords[0], 'dec': mount_coords[1]}, timeout=timeout)
                 success = response['success']
                 if success:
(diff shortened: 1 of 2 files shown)


AnthonyHorton (approved):
I'm not familiar with the mount code but these changes make sense to me as a way to ensure we don't use stale status data. I assume `_update_status()` is fairly quick so that calling it frequently is unlikely to cause any problems?

wtgee (approved):
LGTM although as @AnthonyHorton mentions this depends a bit on `_update_status` being a quick call, which is why we would normally opt for a look up of a status dictionary.

PR #1046, opened 2020-12-14
exptime fix

### LINKED ISSUE #1045: Exposure time not used for exposure timeout
`take_exposure` in `camera.take_exposure` does not account for the exposure time when calculating the timeout. This means exposures longer than`readout_time` + `_timeout` will always raise a `TimeoutError`.

https://github.com/panoptes/POCS/blob/736460a3933e57d8fed94dfa70b1ba6b40913be2/src/panoptes/pocs/camera/camera.py#L781

danjampro:
Fixes the exposure time issue described in #1045 by adding the exposure time to the exposure timeout.

Closes #1045

## How Has This Been Tested?
Unit tests and on Huntsman.

## Types of changes

### DIFF
--- a/src/panoptes/pocs/camera/camera.py
+++ b/src/panoptes/pocs/camera/camera.py
@@ -424,6 +424,7 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
                       filename=None,
                       dark=False,
                       blocking=False,
+                      timeout=None,
                       *args,
                       **kwargs):
         """Take an exposure for given number of seconds and saves to provided filename.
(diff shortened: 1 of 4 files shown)


wtgee (approved):
Thanks @danjampro  for doing this!

This solves the immediate problem but highlights some other issues, which I am putting here just to make a note, but they should be done in future PRs.

* The `Camera._timeout` doesn't really seem to have a good meaning. Likewise, the `readout_time` and any other timers should probably be public. Part of the camera class refactor.
* The `Canon` DSLR class we use for PANOPTES has this issue with the `self._timeout` from the camera but then is mis-using the `timer` anyway.  Might be possible to clean up this base method so it's easier for other classes to inherit.

PR #1048, opened 2020-12-15
Reduce memory usage in autofocus routine

danjampro:
The autofocus code is currently causing OOM errors on 1Gb systems. This can be avoided with a few small changes to the autofocus code.

## How Has This Been Tested?
Unit tests, `memory_profiler`.

## Types of changes


wtgee (approved):
LGTM, although 90% of it is just line-length reformatting. We really need to figure out why we're not all consistent.

danjampro:
> LGTM, although 90% of it is just line-length reformatting. We really need to figure out why we're not all consistent.

Thanks for looking! I think there is still some work to do here as we are still getting some issues with Huntsman.
Is the `panoptes` line length limit not 100 chars?

PR #1050, opened 2020-12-18
Park on error

danjampro:
Make sure POCS parks if an error is encountered in one of the states.

## How Has This Been Tested?
Unit tests.

## Types of changes

### DIFF
--- a/src/panoptes/pocs/state/machine.py
+++ b/src/panoptes/pocs/state/machine.py
@@ -143,10 +143,15 @@ class PanStateMachine(Machine):
                 state_changed = self.goto_next_state()
             except Exception as e:
                 self.logger.critical(f"Problem going from self.state={self.state!r} to "
-                                     f" self.next_state={self.next_state!r}"
-                                     f", exiting loop [{e!r}]")
+                                     f" self.next_state={self.next_state!r}: {e}. Parking")
                 # TODO should we automatically park here?
-                self.stop_states()
+                try:
+                    self.next_state = "parking"
+                    self.goto_next_state()
+                except Exception as e:
+                    self.logger.critical(f"Problem while parking after state machine error: {e}.")
+                finally:
+                    self.stop_states()
                 break
 
             # AFTER TRANSITION TO STATE (NOW INSIDE STATE)
--- a/src/panoptes/pocs/state/machine.py
+++ b/src/panoptes/pocs/state/machine.py
@@ -143,7 +143,7 @@ class PanStateMachine(Machine):
                 state_changed = self.goto_next_state()
             except Exception as e:
                 self.logger.critical(f"Problem going from self.state={self.state!r} to "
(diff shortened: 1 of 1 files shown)


src/panoptes/pocs/state/machine.py
```diff
@@ -143,10 +143,14 @@ def run(self, exit_when_done=False, run_once=False, initial_next_state='ready'):
                 state_changed = self.goto_next_state()
             except Exception as e:
                 self.logger.critical(f"Problem going from self.state={self.state!r} to "
-                                     f" self.next_state={self.next_state!r}"
-                                     f", exiting loop [{e!r}]")
-                # TODO should we automatically park here?
-                self.stop_states()
+                                     f" self.next_state={self.next_state!r}: {e}. Parking.")
```
wtgee, line 146:
```suggestion
                                     f" self.next_state={self.next_state!r}: {e!r}. Parking.")
```

src/panoptes/pocs/state/machine.py
```diff
@@ -143,10 +143,14 @@ def run(self, exit_when_done=False, run_once=False, initial_next_state='ready'):
...
+                try:
+                    self.next_state = "parking"
+                    self.goto_next_state()
+                except Exception as e:
+                    self.logger.critical(f"Problem while parking after state machine error: {e}.")
```
wtgee, line 151:
```suggestion
                    self.logger.critical(f"Problem while parking after state machine error: {e!r}.")
```

wtgee (approved):
I really can't remember why I had this as a TODO instead of just doing it. I want to say there was some complicating factor but since it's undocumented and I can't remember, we'll skip it.

Thanks for getting this done!

src/panoptes/pocs/state/machine.py
```diff
@@ -143,10 +143,14 @@ def run(self, exit_when_done=False, run_once=False, initial_next_state='ready'):
...
+                finally:
+                    self.stop_states()
```
AnthonyHorton, line 153: So this would stop the state machine after the transition to "parking" has completed (i.e. the state's on_enter() has run), preventing it continuing to "parked", "housekeeping", etc.?  Given that an error has occurred this basically does what we want (tells the mount and dome to park, and won't do anything else until restarted) but it would be nice if could enter "parked" before stopping.

AnthonyHorton (approved):
Approved, but supporting @wtgee 's suggested changes, and asking a question about when we stop the state machine.

src/panoptes/pocs/state/machine.py
danjampro: Good point.

wtgee:
@danjampro @AnthonyHorton I think we talked about this not actually being needed in the end since it should provide for a fail-safe backup.  

I'm going to close for now, reopen if necessary.

PR #1053, opened 2020-12-19
Increased timeout for waiting for response from focuser. Fixed some indentations. 

JAAlvarado-Montes:
Added timeout as a parameter to the focuser base class and increased it to 5 instead of 2.

## Description
This PR adds timeout as a parameter to the focuser base class instead of having it harcoded. It changes the timeout from 2 to 5 seconds (as default). This also fixes some incorrect indentations

## How Has This Been Tested?
Local tests

## Types of changes

## Checklist:

### DIFF
--- a/src/panoptes/pocs/focuser/focuser.py
+++ b/src/panoptes/pocs/focuser/focuser.py
@@ -25,6 +25,7 @@ class AbstractFocuser(PanBase, metaclass=ABCMeta):
         model (str, optional): model of the focuser
         port (str, optional): port the focuser is connected to, e.g. a device node
         camera (pocs.camera.Camera, optional): camera that this focuser is associated with.
+        timeout (int, optional): time to wait for response from focuser.
         initial_position (int, optional): if given the focuser will move to this position
             following initialisation.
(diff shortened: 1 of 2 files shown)


src/panoptes/pocs/focuser/focuser.py
```diff
@@ -51,6 +52,7 @@ def __init__(self,
                  model='simulator',
                  port=None,
                  camera=None,
+                 timeout=5,
```
lspitler, line 55: Maybe keep the default as it was before?
```suggestion
                 timeout=2.0,
```

lspitler (commented):
looks good @JAAlvarado-Montes. one suggestion. I can't approve merge, so wont mergw

src/panoptes/pocs/focuser/focuser.py
JAAlvarado-Montes: We discussed with @AnthonyHorton that this can help a little with the birger issues we're having, as we wait a bit longer for the response coming from the birgers and thus avoid the `echo` error we've been finding.

wtgee (approved):
LGTM

PR #1058, opened 2021-01-07
Set FW light position

danjampro:
Fixes the (annoying) problem that a light position is not set unless the FW firsts moves to a dark position. This causes errors to be raised unnecessarily. The solution is to set last light position to the new position in `move_to`, provided it is not the dark position.

## How Has This Been Tested?
Unit tests.

## Types of changes


wtgee (approved):
LGTM.

AnthonyHorton (approved):
That's a neat way to fix the problem. LGTM.

PR #1059, opened 2021-01-07
Retry birger commands

JAAlvarado-Montes:
Retry the command until it matches the birger response.

## Description
The command is sent many time until both the command and the response from the birger are the same.

## Types of changes

## Checklist:

### DIFF
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -250,14 +250,14 @@ class Focuser(AbstractSerialFocuser):
             self.logger.critical("Attempt to send command to {} when not connected!".format(self))
             return
 
-        # Clear the input buffer in case there's anything left over in there.
-        self._serial_port.reset_input_buffer()
-
         # Depending on which command was sent there may or may not be any further
         # response.
         response = []
 
         for i in range(self._max_command_retries):
+            # Clear the input buffer in case there's anything left over in there.
+            self._serial_port.reset_input_buffer()
+
             # Send the command
             self._serial_io.write(command + '\r')
 
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -250,8 +250,7 @@ class Focuser(AbstractSerialFocuser):
             self.logger.critical("Attempt to send command to {} when not connected!".format(self))
             return
 
-        # Depending on which command was sent there may or may not be any further
-        # response.
+        # Depending on which command was sent there may or may not be any further response.
         response = []
 
         # Success variable to verify that the command sent is read by the focuser.
@@ -278,8 +277,8 @@ class Focuser(AbstractSerialFocuser):
                 continue
 
             if response_length == 0:
-                # Not expecting any further response. Should check the buffer anyway in case an error
-                # message has been sent.
+                # Not expecting any further response. Should check the buffer anyway in case an
+                # error message has been sent.
                 if self._serial_port.in_waiting:
                     response.append(self._serial_io.readline())
 
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -283,6 +283,12 @@ class Focuser(AbstractSerialFocuser):
                     # Don't know what to expect. Call readlines() to get whatever is there.
                     response.append(self._serial_io.readlines())
 
+                try:
+                    assert (echo, ok) == (command, 'OK')
+                except AssertionError as err:
+                    self.logger.error(err)
+                    warn(err)
+
                 self.logger.trace(f"Got correct response after {i + 1} attempts")
                 break
 
@@ -291,12 +297,6 @@ class Focuser(AbstractSerialFocuser):
 
             if ok != 'OK':
                 self.logger.warning(f"ok != 'OK': {ok} != 'OK'. Retrying command.")
-
-        try:
-            assert (echo, ok) == (command, 'OK')
-        except AssertionError as err:
-            self.logger.error(err)
-            warn(err)
 
         # Check for an error message in response
         if response:
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -267,7 +267,7 @@ class Focuser(AbstractSerialFocuser):
             echo = self._serial_io.readline().rstrip()
 
             if echo != command:
-                self.logger.warning(f"echo != command: {echo!r} != {command!r}. Retrying command.")
+                self.logger.warning(f'echo != command: {echo!r} != {command!r}. Retrying command.')
                 continue
 
             # Adaptor should then send 'OK', even if there was an error.
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -268,7 +268,7 @@ class Focuser(AbstractSerialFocuser):
             ok = self._serial_io.readline().rstrip()
 
             if echo == command and ok == 'OK':
-                # Get any avaialble response left in buffer.
+                # Get any available response left in buffer.
                 response.append(self._serial_io.readlines())
 
                 self.logger.debug(f"Got correct response after {i + 1} attempts")
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -268,35 +268,17 @@ class Focuser(AbstractSerialFocuser):
             ok = self._serial_io.readline().rstrip()
 
             if echo == command and ok == 'OK':
-                if response_length == 0:
-                    # Not expecting any further response. Should check the buffer anyway in case an error
-                    # message has been sent.
-                    if self._serial_port.in_waiting:
-                        response.append(self._serial_io.readline())
-
-                elif response_length > 0:
-                    # Expecting some number of lines of response. Attempt to read that many lines.
-                    for i in range(response_length):
-                        response.append(self._serial_io.readline())
-
-                else:
-                    # Don't know what to expect. Call readlines() to get whatever is there.
-                    response.append(self._serial_io.readlines())
+                # Get any avaialble response left in buffer.
+                response.append(self._serial_io.readlines())
 
                 self.logger.debug(f"Got correct response after {i + 1} attempts")
                 break
 
             if echo != command:
-                self.logger.warning(f"echo != command: {echo} != {command}. Retrying command.")
+                self.logger.warning(f"echo != command: {echo!r} != {command!r}. Retrying command.")
 
             if ok != 'OK':
-                self.logger.warning(f"ok != 'OK': {ok} != 'OK'. Retrying command.")
-
-        try:
-            assert (echo, ok) == (command, 'OK')
-        except AssertionError as err:
-            self.logger.error(err)
-            warn(err)
+                self.logger.warning(f"ok != 'OK': {ok!r} != 'OK'. Retrying command.")
 
         # Check for an error message in response
         if response:
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -252,6 +252,10 @@ class Focuser(AbstractSerialFocuser):
         # Clear the input buffer in case there's anything left over in there.
         self._serial_port.reset_input_buffer()
 
+        # Depending on which command was sent there may or may not be any further
+        # response.
+        response = []
+
         for i in range(self.max_command_attempts):
             # Send the command
             self._serial_io.write(command + '\r')
@@ -263,6 +267,21 @@ class Focuser(AbstractSerialFocuser):
             ok = self._serial_io.readline().rstrip()
 
             if echo == command and ok == 'OK':
+                if response_length == 0:
+                    # Not expecting any further response. Should check the buffer anyway in case an error
+                    # message has been sent.
+                    if self._serial_port.in_waiting:
+                        response.append(self._serial_io.readline())
+
+                elif response_length > 0:
+                    # Expecting some number of lines of response. Attempt to read that many lines.
+                    for i in range(response_length):
+                        response.append(self._serial_io.readline())
+
+                else:
+                    # Don't know what to expect. Call readlines() to get whatever is there.
+                    response.append(self._serial_io.readlines())
+
                 self.logger.debug(f"Got correct response after {i + 1} attempts")
                 break
 
@@ -277,25 +296,6 @@ class Focuser(AbstractSerialFocuser):
         except AssertionError as err:
             self.logger.error(err)
             warn(err)
-
-        # Depending on which command was sent there may or may not be any further
-        # response.
-        response = []
-
-        if response_length == 0:
-            # Not expecting any further response. Should check the buffer anyway in case an error
-            # message has been sent.
-            if self._serial_port.in_waiting:
-                response.append(self._serial_io.readline())
-
-        elif response_length > 0:
-            # Expecting some number of lines of response. Attempt to read that many lines.
-            for i in range(response_length):
-                response.append(self._serial_io.readline())
-
-        else:
-            # Don't know what to expect. Call readlines() to get whatever is there.
-            response.append(self._serial_io.readlines())
 
         # Check for an error message in response
         if response:
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -252,32 +252,28 @@ class Focuser(AbstractSerialFocuser):
         # Clear the input buffer in case there's anything left over in there.
         self._serial_port.reset_input_buffer()
 
-        # Send command for first time
-        self._serial_io.write(command + '\r')
+        for i in range(self.max_command_attempts):
+            # Send the command
+            self._serial_io.write(command + '\r')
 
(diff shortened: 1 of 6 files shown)


src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,13 +249,15 @@ def _send_command(self, command, response_length=None):
         # Clear the input buffer in case there's anything left over in there.
         self._serial_port.reset_input_buffer()
 
-        # Send command
-        self._serial_io.write(command + '\r')
```
AnthonyHorton, line 253: You shouldn't take this out, otherwise you're trying to read the response before you've even sent the command once. This will cause a delay equal to the serial port timeout for no reason, every single time we send a command.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,13 +249,15 @@ def _send_command(self, command, response_length=None):
...
-
         # In verbose mode adaptor will first echo the command
-        echo = self._serial_io.readline().rstrip()
-        assert echo == command, self.logger.warning("echo != command: {} != {}".format(
-            echo, command))
+        echo = [self._serial_io.readline().rstrip()]
+        while echo[0] != command:
+            # Send command
+            self._serial_io.write(command + '\r')
+            echo[0] = self._serial_io.readline().rstrip()
```
AnthonyHorton, line 257: If the Birger has a problem worse than intermittent serial comms error then this will be an infinite loop! You should _always_ have a finite number of retries, with a small default value, and a corresponding config parameter to change it.

If the expected response is not received within the set number of retries then the code should break out of the loop, log an error, and throw an exception.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,13 +249,15 @@ def _send_command(self, command, response_length=None):
...
 
-        # Send command
-        self._serial_io.write(command + '\r')
-
         # In verbose mode adaptor will first echo the command
-        echo = self._serial_io.readline().rstrip()
-        assert echo == command, self.logger.warning("echo != command: {} != {}".format(
-            echo, command))
+        echo = [self._serial_io.readline().rstrip()]
+        while echo[0] != command:
```
AnthonyHorton, line 254: Should log a warning each time the expect response is not received and a retry is needed. We want a record of how often this is happening.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,13 +249,15 @@ def _send_command(self, command, response_length=None):
...
+            # Send command
+            self._serial_io.write(command + '\r')
+            echo[0] = self._serial_io.readline().rstrip()
+
+        assert echo[0] == command, self.logger.warning("echo != command: {} != {}".format(
+            echo[0], command))
```
AnthonyHorton, line 260: They way you currently have the loop working this is totally redundant code, as there is no way get out of the loop until `echo[0] == command` is `True`.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,13 +249,15 @@ def _send_command(self, command, response_length=None):
...
 
         # Adaptor should then send 'OK', even if there was an error.
         ok = self._serial_io.readline().rstrip()
```
AnthonyHorton, line 263: Note that there are several more steps to verifying responses from the Birger. Checking that the Birger echos the command that was sent to it is just the first step.

We also check here that this is followed by 'OK', check for error codes, then retrieve any additional lines of response that are expected for the given command type.

The serial comms problems are just as likely to effect these subsequent checks as the initial echo check, so they should all be moved inside the retry loop.

AnthonyHorton (changes requested):
It's the right idea, but I have several objections to the way you've implemented it.

src/panoptes/pocs/focuser/birger.py
JAAlvarado-Montes: Got it! Didn't realize that, I thought that reading without sending the command we would have gotten an `''`, so it'd get into the while loop and send the first command from there

src/panoptes/pocs/focuser/birger.py
JAAlvarado-Montes: But what if the response from the birger is correct when the command is sent for the first time as expected? If everything is inside the retry loop (which only runs if `echo != command`), that'd mean there would not be subsequent checks if the loop is not triggered.

src/panoptes/pocs/focuser/birger.py
JAAlvarado-Montes: What I'm going to do is to move everything inside the loop, even the first time that the command is sent. That way I wouldn't be reading before sending the command for the first time.

JAAlvarado-Montes:
Hi @AnthonyHorton,

Please review again this PR when you've got a chance. I have addressed all of your comments, though there might still be some things to improve. I would really appreciate your input. Thank you!

Cheers,
Jaime A.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -63,10 +63,13 @@ def __init__(self,
                  model='Canon EF-232',
                  initial_position=None,
                  dev_node_pattern='/dev/tty.USA49*.?',
+                 max_command_attempts=5,
                  *args, **kwargs):
         super().__init__(name=name, model=model, *args, **kwargs)
         self.logger.debug('Initialising Birger focuser')
 
+        self.max_command_attempts = max_command_attempts
```
danjampro, line 71: @JAAlvarado-Montes I thinks this should be a private variable, `self._max_command_attempts`

danjampro, line 71: I think this line needs to come before calling `super().__init__`. Right now I'm getting this error on creating a focuser:
```
D 01-12 03:59:24.850 (03:59:24 UTC) | panoptes.pocs.camera.camera _create_subcomponent:1039        | Creating the base_class_name='AbstractFocuser' object from dict
D 01-12 03:59:24.857 (03:59:24 UTC) | panoptes.pocs.focuser.focuser __init__:99                    | Focuser created: Birger Focuser on /dev/ttyUSB0
D 01-12 03:59:24.873 (03:59:24 UTC) | panoptes.pocs.focuser.serial _connect:107                    | Established serial connection to Birger Focuser on /dev/ttyUSB0.
E 01-12 03:59:24.879 (03:59:24 UTC) | huntsman.pocs.utils.pyro.cli service:87                      | Pyro huntsmanpi000 shutdown unexpectedly AttributeError("'Focuser' object has no attribute 'max_command_attempts'")
```

danjampro (approved):
LGTM and tested on Huntsman

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,50 @@ def _send_command(self, command, response_length=None):
...
+                        response.append(self._serial_io.readline())
+
+                elif response_length > 0:
+                    # Expecting some number of lines of response. Attempt to read that many lines.
+                    for i in range(response_length):
+                        response.append(self._serial_io.readline())
+
+                else:
+                    # Don't know what to expect. Call readlines() to get whatever is there.
+                    response.append(self._serial_io.readlines())
```
wtgee, line 284: The conditionals here don't do anything.  In all situations you are just reading the buffer and storing it. There shouldn't be anything coming back after the `response_length` and if so maybe that shouldn't be added to the `response` anyway.

```suggestion
            if echo == command and ok == 'OK':
                # Get any available response
                response.append(self._serial_io.readlines())
```

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,50 @@ def _send_command(self, command, response_length=None):
...
+
+                self.logger.debug(f"Got correct response after {i + 1} attempts")
+                break
+
+            if echo != command:
+                self.logger.warning(f"echo != command: {echo} != {command}. Retrying command.")
```
wtgee, line 290:
```suggestion
                self.logger.warning(f"echo != command: {echo!r} != {command!r}. Retrying command.")
```

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,50 @@ def _send_command(self, command, response_length=None):
...
+
+            if ok != 'OK':
+                self.logger.warning(f"ok != 'OK': {ok} != 'OK'. Retrying command.")
```
wtgee, line 293: Do we know in what situations the `echo` would be correct but the `ok` wouldn't?

```suggestion
                self.logger.warning(f"ok != 'OK': {ok!r} != 'OK'. Retrying command.")
```

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,50 @@ def _send_command(self, command, response_length=None):
...
+
+        try:
+            assert (echo, ok) == (command, 'OK')
+        except AssertionError as err:
+            self.logger.error(err)
+            warn(err)
```
wtgee, line 299: Don't you do all of this above?

```suggestion
```

wtgee (changes requested):
Seems like this should be built into the `AbstractSerialFocuser` rather than just the Birger `Focuser`.  Any reason it's not?

JAAlvarado-Montes:
Hi @wtgee , this is done only for birgers because only these controllers send a response back that needs to be verified with the command sent. The astromechs don't do this.

I've addressed all of your comments above 👍 

Cheers,
Jaime A.

wtgee:
> Hi @wtgee , this is done only for birgers because only these controllers send a response back that needs to be verified with the command sent. The astromechs don't do this.

Well, there are two things going on in this PR, namely 1) retrying commands that fail and 2) verifying the response.

The retrying of the commands (the title of the PR) should be generic and available to all serial-based focusers.  The verifying of the response could potentially be generic as well.

JAAlvarado-Montes:
However, we don't need to retry commands for other focusers, that's only needed for the birgers. Each focuser has its own way to read the commands, that's why `_send_command` is different for each focuser controller.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,32 @@ def _send_command(self, command, response_length=None):
         # Clear the input buffer in case there's anything left over in there.
         self._serial_port.reset_input_buffer()
```
AnthonyHorton, line 254: This should be done before each time you send the command, i.e. you should move this line inside the loop just before the `self._serial_io.write(command + '\r')` line.

If you don't clear the buffer each time you might end up reading part of the response from a previous failed attempt to send the command, causing the response validation to fail unnecessarily.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,32 @@ def _send_command(self, command, response_length=None):
...
 
-        else:
-            # Don't know what to expect. Call readlines() to get whatever is there.
-            response.append(self._serial_io.readlines())
+            # Adaptor should then send 'OK', even if there was an error.
+            ok = self._serial_io.readline().rstrip()
+
+            if echo == command and ok == 'OK':
+                # Get any avaialble response left in buffer.
+                response.append(self._serial_io.readlines())
```
AnthonyHorton: You've removed some of the functionality of the original code here. Almost all of the Birger commands have an expected response length, i.e. how many lines of response there should be after the echo and the 'OK'. 

`self._serial_io.readlines()` is supposed to be the fall back option only used for the 1 or 2 commands that return an unknown number of lines of response because it will block for the duration of the serial timeout waiting for more lines of response. When the number of lines is known it is far quicker to read only the expected number of lines.

We always try to read at least 1 line though, because it's possible there will be an error message immediately after the 'OK'.

```suggestion
                if response_length == 0:
                    # Not expecting any further response. Should check the buffer anyway in case an error
￼                   # message has been sent.
                    response.append(self._serial_io.readline())
                elif response_length > 0:
                    # Expecting some number of lines of response. Attempt to read that many lines.
                    for i in range(response_length):
￼                        response.append(self._serial_io.readline())
                else:
                    # Don't know what to expect. Call readlines() to get whatever is there.
                    response.append(self._serial_io.readlines())
```

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,32 @@ def _send_command(self, command, response_length=None):
...
+
+                self.logger.debug(f"Got correct response after {i + 1} attempts")
```
AnthonyHorton, line 274: It's debatable but I feel that this message is unnecessary, or at least should be downgraded from DEBUG to TRACE level. If retries are needed then they will be logged and this message won't add much info, and if it works first time like it's supposed to we probably don't need a additional debug message either.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,32 @@ def _send_command(self, command, response_length=None):
...
+                break
+
+            if echo != command:
+                self.logger.warning(f"echo != command: {echo!r} != {command!r}. Retrying command.")
+
+            if ok != 'OK':
+                self.logger.warning(f"ok != 'OK': {ok!r} != 'OK'. Retrying command.")
```
AnthonyHorton, line 281: You are checking the echo and the 'OK' inside the loop, but you've left the check for an error message outside the loop. I know that our bad serial connections are (so far) causing things to fail at the echo check but it is possible that other intermittent Birger problems could pass the echo and 'OK' check but cause occasional error messages to occur (e.g. ERR21) and the retry code could work around those too if you move the check inside the loop.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -249,36 +253,32 @@ def _send_command(self, command, response_length=None):
...
```
AnthonyHorton, line 282: Important thing missing here: there is no check that the command eventually succeeds! If the checks fail it will keep retrying until the maximum number of retries have been reached, but after that the code just carries on regardless.

You need something like a `success` variable that get set to `False` at the start, and gets set to `True` if all the checks pass. Once you exit the loop you need to check the `success` variable and if it is not `True` you need to raise a `PanError` with a suitable message.

AnthonyHorton (changes requested):
Several things I'd still like to see changed here, but the show stopper is the absence of a check that the command eventually succeeds. At the moment the code will just carry on regardless once the maximum number of retries has been reached, even if all attempts fail.

AnthonyHorton:
> Regarding 1) we don't need to retry commands for other focusers, that's only needed for the birgers. Each focuser has its own way to read the commands that are sent to them, that's why `_send_command` is different for each focuser controller. Moreover, it's only the birgers which send a response back and the `OK` verification (which are the two things we use to see if a command needs to be retried). It'd be completely unnecessary and a mistake having all of this for other focuser (e.g. the astromechs).

You can still make this generic. There is no particular reason why we should only retry commands for the Birgers. Any serial connection can drop characters if there is a bad connection, all the focusers we support use serial connections, and all the focusers send a checkable response to at least some of their commands, so all of them could potentially benefit from the retry functionality. The way to implement this would be to have a generic `send_command()` method in the serial focuser base class which includes the retry functionality, and which uses calls to a focuser specific `_validate_response()` method which reads the response (if any) & returns, say, a tuple of `(success, response)`.

> Regarding 2), the verifying response was already there and it's always been only for the birgers focuser, I only moved it into the retry loop. We don't this for the other focusers.

It was written like that because at that point the Birgers were the only focuser. Now there are 3 types we have a better idea of what functionality they have in common (or should have in common), and that common functionality is better implemented in the base class.

src/panoptes/pocs/focuser/birger.py
```diff
@@ -246,39 +250,62 @@ def _send_command(self, command, response_length=None):
...
+
+        try:
+            assert(success)
+        except AssertionError:
+            raise error.PanError(f'The command {command!r} was not properly read by {self}')
```
AnthonyHorton, line 308: With all the edits this has ended up a bit messy. Suggestion to clean it up:

```
        response = []
        success = False

        for i in range(self._max_command_retries):
            self._serial_port.reset_input_buffer()
            self._serial_io.write(command + '\r')

            echo = self._serial_io.readline().rstrip()
            if echo != command:
                self.logger.warning(f"echo != command: {echo!r} != {command!r}. Retrying command.")
                continue

            ok = self._serial_io.readline().rstrip()
            if ok != 'OK':
                self.logger.warning(f"ok != 'OK': {ok!r} != 'OK'. Retrying command.")
                continue

            if response_length == 0:
                # Not expecting any further response. Should check the buffer anyway in case an error
                # message has been sent.
                if self._serial_port.in_waiting:
                    response.append(self._serial_io.readline())
            elif response_length > 0:
                # Expecting some number of lines of response. Attempt to read that many lines.
                for i in range(response_length):
                     response.append(self._serial_io.readline())
            else:
                # Don't know what to expect. Call readlines() to get whatever is there.
                response.append(self._serial_io.readlines())

            success = True
            break

        if not success:
            raise error.PanError(f'Failed command {command!r} on {self}')

```

AnthonyHorton (changes requested):
With all the back and forth edits this has ended up a bit messy. I have suggested a cleaner rewrite of the main part of the code.

In addition this is still only checking for error codes after the retry loop, instead of inside the loop, which would be better.

wtgee (approved):
Thanks @JAAlvarado-Montes and @AnthonyHorton. Approving here to clear my change request, will merge on passing tests.

AnthonyHorton (approved):
LGTM.

PR #1062, opened 2021-01-11
Fix autofocus make_plots

danjampro:
Default value for `make_plots` is `False`, which means plots are not made even if `autofocus_make_plots=True`.


JAAlvarado-Montes (approved):
LGTM!

wtgee (approved):
👍🏼

Issue #1061, opened 2021-01-11
pocs.observatory.mount.slew_to_coordinates not working when coords are in SkyCoord format (as it's required)


JAAlvarado-Montes:
**Describe the bug**
pocs.observatory.mount.slew_to_coordinates requires coordinates to be in SkyCoord format. However, even when using the coordinates in this format the function rises an `AssertionError` error saying that the coordinates are not a tuple. 

**To Reproduce**
coords = SkyCoord(300, -60, unit="deg")
pocs.observatory.mount.slew_to_coordinates(coords)

Tests are not passing in https://github.com/AstroHuntsman/huntsman-pocs/pull/287

JAAlvarado-Montes:
Closed by #1063

PR #1063, opened 2021-01-12
Fix mount.slew_to_coordinates

danjampro:
Fix a bug where `mount.slew_to_coordinates` requires the wrong type for coordinates.


JAAlvarado-Montes:
@danjampro you could mention that this PR solves issue #1061 ?

wtgee (approved):
LGTM.

Just a note that `slew_to_coordinates` as used by PANOPTES is a convenience function. The "normal" process would be `set_target_coordinates` -> `slew_to_target`.

PR #1065, opened 2021-01-14
Add command lock to bisque mount

danjampro:
TheSkyX cannot handle simultaneous commands, which can cause `TheSkyXTimeout` errors which can lead to other errors. The solution is to use a lock to ensure commands are sent one after the other.

## Related Issue
https://github.com/AstroHuntsman/huntsman-pocs/issues/303

## How Has This Been Tested?
Unit tests, will test on Huntsman.

## Types of changes

### DIFF
--- a/src/panoptes/pocs/mount/bisque.py
+++ b/src/panoptes/pocs/mount/bisque.py
@@ -1,6 +1,7 @@
 import json
 import os
 import time
+from threading import Lock
 
 from astropy import units as u
 from astropy.coordinates import SkyCoord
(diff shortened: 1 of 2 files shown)


src/panoptes/pocs/mount/bisque.py
```diff
@@ -122,6 +125,12 @@ def initialize(self, unpark=False, *args, **kwargs):
 
         return self.is_initialized
 
+    def query(self, *args, **kwargs):
+        """ Override the query method to use the command lock. This is required because TheSkyX
+        cannot handle simulataneous commands."""
```
wtgee, line 130: Can we break this up?

wtgee: I'm trying to be good about these docstrings.

wtgee (approved):
So what happens when it's called and locked? It blocks?

PR #1070, opened 2021-01-22
Get serial number for astromechs and use dev_node_pattern

JAAlvarado-Montes:
Add code to get the serial number of astromechanics controllers. Also, add the `dev_node_pattern` parameter.

## Description
Get the serial number of astromechanics controllers and add the `dev_node_pattern` parameter so it can be used to search for both astromechanics and birger controllers and initialize them.

## How Has This Been Tested?
The changes were tested locally using real hardware.

## Types of changes

## Checklist:


src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -8,6 +15,13 @@ class Focuser(AbstractSerialFocuser):
     Args:
         name (str, optional): default 'Astromechanics Focuser'
         model (str, optional): default 'Canon EF/EF-S'
+        initial_position (int, optional): if given the focuser will drive to this encoder position
+            following initialisation.
+        dev_node_pattern (str, optional): Unix shell pattern to use to identify device nodes that
+            may have an Astromechanics adaptor attached. Default is '/dev/tty.usbserial-AG0*.?',
+            which is intended to match all the nodes created by FT232R USB UART serial
+            adaptors, as used at the time of writing by Huntsman.
```
wtgee, line 23: This should default to `None` in POCS and then be set via the config.

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -20,8 +34,15 @@ class Focuser(AbstractSerialFocuser):
     def __init__(self,
                  name='Astromechanics Focuser Controller',
                  model='astromechanics',
+                 initial_position=0,
+                 dev_node_pattern='/dev/tty.usbserial-AG0*',
+                 serial_number_pattern=serial_number_pattern,
```
wtgee, line 39: The dev node pattern should be `None` in POCS to keep it generic and the values can come from the config.

But, if you are not doing anything with these parameters here and they are already in the super class, you don't really need to specify them, they will just go through with `**kwargs`

```suggestion
                 dev_node_pattern=None,
                 serial_number_pattern=serial_number_pattern,
```

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -58,8 +79,18 @@ def connect(self, port):
 
         self._connect(port, baudrate=38400)
```
wtgee, line 80: You are not changing this, but ideally this baudrate should also come from the config.

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -58,8 +79,18 @@ def connect(self, port):
...
+
+        if res and res.endswith("#"):
+            try:
+                dev = usb.core.find(idVendor=0x0403, idProduct=0x6001)
+                self._serial_number = usb.util.get_string(dev, dev.iSerialNumber)
+                # Return string that makes it clear there is no serial number
+                return self._serial_number
+
+            except Exception:
+                raise error.PanError(f"{self.name} not found in {self.port}")
```
wtgee, line 93: I still think the exception should be raised when the `if` fails, not when the serial number check fails.  And can we check it for a valid position response rather than the generic `#`?

src/panoptes/pocs/focuser/birger.py
```diff
@@ -64,57 +63,18 @@ def __init__(self,
                  initial_position=None,
                  dev_node_pattern='/dev/tty.USA49*.?',
                  max_command_retries=5,
+                 serial_number_pattern=serial_number_pattern,
```
wtgee, line 66: Similar to the astromechanics, the `dev_node_pattern` should default to `None` and these parameters probably don't need to be specified.

src/panoptes/pocs/focuser/serial.py
```diff
@@ -15,13 +16,55 @@ class AbstractSerialFocuser(AbstractFocuser):
     # known focuser devices & acts as a check against adaptors assigned to incorrect ports.
     _assigned_nodes = []
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, dev_node_pattern="", initial_position=None,
+                 serial_number_pattern="", *args, **kwargs):
```
wtgee, line 20: Default to `None` for unknown things.

```suggestion
    def __init__(self, dev_node_pattern=None, initial_position=None,
                 serial_number_pattern=None, *args, **kwargs):
```

wtgee: For the `serial_number_pattern`, I would create your own `re` object if given a string.

src/panoptes/pocs/focuser/astromechanics.py
wtgee, line 93:
```suggestion
                raise error.BadSerialConnection(f"{self.name} not found in {self.port}")
```

src/panoptes/pocs/focuser/serial.py
```diff
@@ -15,13 +16,55 @@ class AbstractSerialFocuser(AbstractFocuser):
...
+            # Search in cached device node scanning results for serial number
+            try:
+                device_node = AbstractSerialFocuser._adaptor_nodes[self.port]
+            except KeyError:
+                message = 'Could not find {} ({})!'.format(self.name, self.port)
+                self.logger.error(message)
+                warn(message)
+                return
+            self.logger.debug('Found {} ({}) on {}'.format(self.name, self.port, device_node))
+            self.port = device_node
```
wtgee, line 66: This should all be moved out of the class instance and become a class method that is called before object initialization.

See discussion in PR review.

wtgee (changes requested):
I realize this is just a copy/paste job from what was there, but I think if this is not a huge priority time wise then we should take the time to clean up the constructor.

Ideally the `dev_node_pattern` does not happen in an object instance so I'm proposing to make this a class method that would be called prior to object initialization. 

I don't see any advantage to this happening dynamically upon creation, but let me know if there is a valid reason still.  @AnthonyHorton @JAAlvarado-Montes

JAAlvarado-Montes:
Hi @wtgee @AnthonyHorton @danjampro ,

I have addressed all of @wtgee's comments. I'll submit a different PR to `huntsman-config` to add  `baudrate` (birger and astromech); `id_vendor` and `id_product` (only for astromech). The classes are now much more organized and I even find the code more readable.

Everything was tested on Huntsman's lens Tuc with different Astromechanics focusers. Please let me know any suggestion or comment.

PS. Tests keep failing because of the new `pyusb` module that astromechanics.py is using. However, I've added it as a new requirement for POCS as @wtgee indicated, so I'm not sure what is happening.

Cheers,
Jaime A.

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -8,25 +12,25 @@ class Focuser(AbstractSerialFocuser):
...
     """
 
-    def __init__(self,
-                 name='Astromechanics Focuser Controller',
-                 model='astromechanics',
-                 *args, **kwargs):
+    def __init__(self, name='Astromechanics Focuser', model='Canon EF-232', id_vendor=None,
+                 id_product=None, *args, **kwargs):
+
+        self._id_vendor = id_vendor
```
danjampro, line 25: Minor detail, but shouldn't these be `_vendor_id` and `_product_id`?

src/panoptes/pocs/focuser/birger.py
```diff
@@ -47,77 +41,23 @@ class Focuser(AbstractSerialFocuser):
...
-        initial_position (int, optional): if given the focuser will drive to this encoder position
-            following initialisation.
-        dev_node_pattern (str, optional): Unix shell pattern to use to identify device nodes that
-            may have a Birger adaptor attached. Default is '/dev/tty.USA49*.?', which is intended
-            to match all the nodes created by Tripplite Keyway USA-49 USB-serial adaptors, as
-            used at the time of writing by Huntsman.
-
-    Additional positonal and keyword arguments are passed to the base class, AbstractFocuser. See
-    that class' documentation for a complete list.
+        max_command_retries (int, optional): Maximum number of times to retry a command. Default 5
```
danjampro, line 44: Do we need a `max_retries` for these focusers?

danjampro (commented):
Hi @JAAlvarado-Montes looks good. I'm not an expert in the focuser code so I'll leave it to the others to give the final review. The test coverage is showing a bit of a drop, can anything be done about this?

src/panoptes/pocs/focuser/birger.py
JAAlvarado-Montes: Yes we do, it is used as the maximum number of times to retry a command when it isn't properly sent

lspitler:
@JAAlvarado-Montes what is the status of this? Do you need a review from @wtgee ?

JAAlvarado-Montes:
@lspitler, as far as I'm concerned all the changes from this PR are ready and working with the Astromechs. I followed all of @wtgee's and @danjampro's comments. Also, tests are passing with no issues.

lspitler:
@wtgee @AnthonyHorton we're about to deploy the AstroMechs in ~7 days. @JAAlvarado-Montes is this PR required for deploying them?

wtgee (commented):
@JAAlvarado-Montes I did this review a long time ago but apparently never hit submit, sorry about that!  I think most of the comments still apply but let me know if something doesn't make sense.

JAAlvarado-Montes:
Replaced by #1125

PR #1071, opened 2021-01-27
Add ZWO camera bandwidthoverload arg

danjampro:
- Huntsman ZWO cameras are suffering from USB timeout issues
- Possible solution is to modify the BANDWIDTHOVERLOAD control value
- This PR adds an argument to the ZWO camera class to allow this to be easily set from config

### DIFF
--- a/src/panoptes/pocs/camera/zwo.py
+++ b/src/panoptes/pocs/camera/zwo.py
@@ -21,6 +21,7 @@ class Camera(AbstractSDKCamera):
                  name='ZWO ASI Camera',
                  gain=None,
                  image_type=None,
+                 bandwidthoverload=None,
                  *args, **kwargs):
         """
         ZWO ASI Camera class
@@ -56,6 +57,9 @@ class Camera(AbstractSDKCamera):
 
         if gain:
             self.gain = gain
+
+        if bandwidthoverload is not None:
+            self.bandwidthoverload = bandwidthoverload
 
         if image_type:
             self.image_type = image_type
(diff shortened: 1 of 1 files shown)


src/panoptes/pocs/camera/zwo.py
```diff
@@ -143,6 +147,15 @@ def is_exposing(self):
...
+        return self._control_getter('BANDWIDTHOVERLOAD')[0]
+
+    @bandwidthoverload.setter
+    def bandwidthoverload(self, value):
+        value = get_quantity_value(value, u.percent) * u.percent
+        self._control_setter('BANDWIDTHOVERLOAD', value)
+
     # Methods
 
     def connect(self):
```
Physarah, line 161: Does `BANDWIDTHOVERLOAD` get retrieved here as well?

Physarah (approved):
Looks good!

Physarah:
Sorry Dan I don't have write access ;P

wtgee:
@Physarah Is a property needed? Why not the solution presented in your description in https://github.com/AstroHuntsman/huntsman-pocs/issues/330?

danjampro:
> @Physarah Is a property needed? Why not the solution presented in your description in [AstroHuntsman/huntsman-pocs#330](https://github.com/AstroHuntsman/huntsman-pocs/issues/330)?

@wtgee The problem is that these errors are random. We'll have to run many exposures *during normal pocs operation* to see if this makes a difference. This method will allow us to easily configure e.g. half the cameras on 100% and the other half on a lower value.

danjampro:
@wtgee we discussed this and agreed a property is appropriate. Can we merge?

wtgee (approved):
Ummm.....better late than never?

PR #1069, opened 2021-01-22
Changes to observation

danjampro:
- Exposure times must be greater than *or equal to* 0 to allow for biases.
- Added a `set_is_finished` property to `observation` to allow abstract definitions, which also simplifies the `analyzing` state.
- Optional `dark` attribute is added to observations, to signal whether to take dark exposures or not.
- Added bias and dark observations, move observations classes into the `scheduler.observation` directory.

## How Has This Been Tested?
Unit tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #1072, opened 2021-01-27
Silence missing filter name warning for dark observations

danjampro:
- Currently a logger call at `info` level is made when an observation is taken without a filter specified (`filter_name=None`).
- Dark observations do not know which filter_name corresponds to the dark position for each camera, so `filter_name=None` is always (and necessarily) specified.
- New behaviour is to log a warning if the observation is not dark and the filter is not None


wtgee:
@danjampro I still think it might be a good time to bring the `BiasObservation` and `DarkObservation` into POCS itself, which would clean this up a bit as we wouldn't have to pass around the `dark` flag (which I just don't like for some reason).

Any reason we couldn't move most of this PR here? https://github.com/AstroHuntsman/huntsman-pocs/pull/332

danjampro:
Closing in favour of #1069

PR #1076, opened 2021-02-05
Add optional focus position offsets per filter

danjampro:
## Description
Changing filters can significantly affect the optimal focus position, meaning fine autofocusing does not always work correctly. This motivates (optional) focus positions to be applied per filter, which are applied during calls to `filterwheel.move_to`.

## How Has This Been Tested?
Unit tests.

## Types of changes

### DIFF
--- a/src/panoptes/pocs/camera/camera.py
+++ b/src/panoptes/pocs/camera/camera.py
@@ -358,6 +358,11 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
         """ Return True if the camera has a focuser, False if not. """
         return self.focuser is not None
 
+    @property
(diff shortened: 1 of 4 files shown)


wtgee (approved):
LGTM other than the minor comments.  I would almost think we would return to the focus position and then apply the offset rather than move directly from offset to offset, but I suppose it doesn't really matter.

wtgee:
You could up the -3% coverage on the `filterwheel.py` pretty easily with one or two more small tests.

danjampro:
> You could up the -3% coverage on the `filterwheel.py` pretty easily with one or two more small tests.

Think I managed to up it in that one test after the final changes!

PR #1084, opened 2021-02-19
Fix FW moving error

danjampro:
I broke POCS with my last PR. I am sorry. The fix is to check if the FW is moving before applying focus offsets, rather than after.

## How Has This Been Tested?
Unit tests.

## Types of changes


wtgee (approved):
Thanks for checking into this.  Kind of odd there is such a hit on coverage in `test_logger`, I haven't seen that. The focuser is down 2% too which is also odd, but not a show stopper.

Issue #1087, opened 2021-03-04
Add config option for Moon avoidance


danjampro:
It would be useful for Huntsman to have a configurable Moon avoidance parameter:

https://github.com/panoptes/POCS/blob/cf17bc9ab888fd37bfc84d7f8f480d18e0053d94/src/panoptes/pocs/scheduler/__init__.py#L61

Issue #402, opened 2018-01-21
Specify constraints in config file


wtgee:
The constraints to use are currently hard-coded into the `Observatory` class but need to come from a config file. This would like be part of the move to dependency injection on part of the scheduler.

jermainegug:
How are they hard-coded into the `Observaory` class? (Not questioning you, just not sure how to spot this myself)

wtgee:
When an `Observatory` object is created it calls out `_create_scheduler`, which has the constraints hard-code [here](https://github.com/panoptes/POCS/blob/develop/pocs/observatory.py#L786). The idea would be to have a list of the constraint names under the `scheduler` key in the config file and then to use the `load_module` (see  [`_create_mount`](https://github.com/panoptes/POCS/blob/develop/pocs/observatory.py#L619) for an example of using this.)

wtgee:
Not that you need the names and potentially any options in the config so might be a little tricky.

jermainegug:
```
constrains = load_module(
                    'pocs.scheduler.{}'.format(constraint))
```
Do I want to do something like this?

jermainegug:
Quite lost with this at the moment, not sure how to do it. Can't understand how `_create_module` works.

wtgee:
Ok, I wouldn't worry about this for now. We want to pull all those private create methods out soon anyway and it is probably best done there.

PR #1107, opened 2021-05-28
Fix hardcoded scheduler

### LINKED ISSUE #1086: Hardcoded reference to panoptes.pocs in create_scheduler_from_config
https://github.com/panoptes/POCS/blob/cf17bc9ab888fd37bfc84d7f8f480d18e0053d94/src/panoptes/pocs/scheduler/__init__.py#L47

wtgee:
Removing the hard-coded namespace for loading of schedulers. Fully qualified namespace is required in config file now, i.e.

```yaml
scheduler:
  type: dispatch
```
becomes
```yaml
scheduler:
  type: panoptes.pocs.scheduler.dispatch
```

## Description
* Moves the hard-coded namespace out of the code and into the `get_config`.  This does not add a check for the namespace in the code but instead assumes the scheduler loader will raise an Exception.
* Changes the default `directories.base` to `.` instead of `/panoptes-pocs`.
  * Specify relative path to fields file.
* Moves all of the scheduler tests into a subfolder for better organization although that is not required by the PR.
* Removes `${PANDIR}` from the `docker-compose.yaml` file.
* **Unrelated to PR** The `test_sdk_camera_not_found` is creating a camera class and then raising an exception with an invalid config, however the camera object is still in memory and not getting cleaned up by the garbage collector. This is causing a race condition on the `test_sdk_already_in_use` test, causing it to hang indefinitely in an odd way.  Simply moving the offending test below seems to fix the problem although it should probably be fixed.

## Related Issue
Closes #1086 

## How Has This Been Tested?
Added a simple test for failure on just name (previous behavior) and passing on full namespace.


danjampro (approved):
LGTM. The only thing I can suggest is making `create_scheduler_from_config` more generic. It would be nice to fully configure the scheduler from the config. This can be accomplished by parsing config items under `scheduler` to its init function.

wtgee:
> LGTM. The only thing I can suggest is making `create_scheduler_from_config` more generic. It would be nice to fully configure the scheduler from the config. This can be accomplished by parsing config items under `scheduler` to its init function.

This is being done now for everything currently in the default `pocs.yaml`.  Huntsman has added some constraint information to the scheduler config, which I think is what you are referring to.  Fully agree that it should be done but I'll leave that as separate PR (#402 #1087) to get to soon.

@danjampro let me know if you were thinking of something different.

Issue #1124, opened 2021-06-22
Use panoptes-utils `SerialDevice` in `AbstractSerialFocuser`


danjampro:
Tests with Birger and Astromechanics focusers on Huntsman have shown that a speed up of ~5x can be achieved by doing this.

See AstroHuntsman/huntsman-pocs#488

wtgee:
I added something to #1125 that might help. Essentially any time the serial class tries to read (either via a direct `read` or some kind of `readline`) there is a risk of it waiting for a response from the device unless you give it a `hint`, which is the first (and only) parameter to those functions (and is the number of bytes to read, even with `readline(s)`.  The preferred way to overcome that is to give a hint equal to the `in_waiting` property, which will default to `0` if nothing is there to be read, otherwise the actual number of bytes.

Issue #1106, opened 2021-05-27
Inconsistency in `focuser.move_by`


danjampro:
The abstract `Focuser.move_by` returns the final focuser position:

https://github.com/panoptes/POCS/blob/e79ac499c924b0a06eb2944a347f8d9251f6db51/src/panoptes/pocs/focuser/focuser.py#L176

Whereas the Birger version returns the change in focuser position:

https://github.com/panoptes/POCS/blob/e79ac499c924b0a06eb2944a347f8d9251f6db51/src/panoptes/pocs/focuser/birger.py#L228

PR #1125, opened 2021-06-22
Astromech find serial port

wtgee:
* Use the `vendor_id` and `product_id` of the astromech to find the matching serial device path. This is mostly taken from #1070 but minimizes the changes to just the astromech and leaves the Birger and base focuser class alone. **This is expecting the vendor and product ids to be in hex (e.g. `0x12345`), is this correct?**
* Moves the `#` command terminator to `_send_command` as a `post_cmd` default param. I've added `pre_cmd` for completeness and to be compatible with a future change to the base class.
* Also adds a `hint` to `readline` to try and speed up the reading from serial.
* Whitespace and f-string replacements included.

## Related Issue
Replaces #1070 
#1124 might not be necessary with this.

@JAAlvarado-Montes does this accomplish what #1070 was supposed to?

### DIFF
--- a/src/panoptes/pocs/focuser/astromechanics.py
+++ b/src/panoptes/pocs/focuser/astromechanics.py
@@ -1,39 +1,47 @@
+from panoptes.utils import error
+from panoptes.utils.serial.device import find_serial_port
 from panoptes.pocs.focuser.serial import AbstractSerialFocuser
 
 
 class Focuser(AbstractSerialFocuser):
     """
-    Focuser class for control of a Canon DSLR lens via an Astromechanics Engineering Canon EF/EF-S adapter.
+    Focuser class for control of a Canon DSLR lens via an Astromechanics
+    Engineering Canon EF/EF-S adapter.
 
     Args:
         name (str, optional): default 'Astromechanics Focuser'
         model (str, optional): default 'Canon EF/EF-S'
+        vendor_id (str, optional): idVendor of device, can be retrieved with lsusb -v from terminal.
+        product_id (str, optional): idProduct of device, can be retrieved with lsusb -v from terminal.
 
-    Additional positonal and keyword arguments are passed to the base class, AbstractSerialFocuser. See
-    that class' documentation for a complete list.
+    Additional positonal and keyword arguments are passed to the base class,
+    AbstractSerialFocuser. See that base class documentation for a complete list.
 
     Min/max commands do not exist for the astromechanics controller, as well as
     other commands to get serial numbers and library/hardware versions. However,
     as they are marked with the decorator @abstractmethod, we have to override them.
     """
 
-    def __init__(self,
-                 name='Astromechanics Focuser Controller',
-                 model='astromechanics',
-                 *args, **kwargs):
-        super().__init__(name=name, model=model, *args, **kwargs)
-        self.logger.debug('Initialising Astromechanics Lens Controller')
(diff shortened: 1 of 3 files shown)


src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -107,62 +113,63 @@ def move_by(self, increment):
         self.logger.debug(f"Moved by {increment} encoder units. Current position is {new_pos}")
         return new_pos
```
danjampro, line 114: In order for things to be consistent, this should probably be `self.position`. Also, with the Birger class, `move_by` commands actually return the encoder increment rather than the actual position. Although that does seem inconsistent with the abstract focuser class. That might lead to some serious errors if there are a mix of astromechs and birgers as we have on Huntsman.

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -1,39 +1,47 @@
...
+    def __init__(self, name='Astromechanics Focuser', model='Canon EF-232', vendor_id=None,
+                 product_id=None, *args, **kwargs):
+        # Check if have device, raise error.NotFound if unable to find.
+        self._serial_number = find_serial_port(vendor_id, product_id)
+        self._vendor_id = vendor_id
+        self._product_id = product_id
+
         super().__init__(name=name, model=model, *args, **kwargs)
-        self.logger.debug('Initialising Astromechanics Lens Controller')
+        self.logger.debug(f'Initializing {name}')
```
danjampro, line 34: I noticed that the move to initial position is currently done in the Birger class rather than the abstract serial class or abstract focuser class. Maybe it should be implemented here or in the base class for consistency?

danjampro (commented):
LGTM. I marked one potential major problem in the `move_by` command, that is associated with the `Birger` class and issued in #1106

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -107,62 +113,63 @@ def move_by(self, increment):
...
-
-    def _move_zero(self):
-        self.logger.debug('Setting focus encoder zero point')
         self._is_moving = True
         try:
-            # Set focuser to 0 position
-            self._send_command('M0#')
-
-            self.logger.debug('Moved to encoder position 0')
+            self._send_command('A00')
```
JAAlvarado-Montes, line 159:
```suggestion
            self._send_command('A00#')
```

src/panoptes/pocs/focuser/astromechanics.py
```diff
@@ -107,62 +113,63 @@ def move_by(self, increment):
...
         finally:
             self._is_moving = False
+
+    def _move_zero(self):
+        self.logger.debug('Setting focus encoder zero point')
+        if self.position != 0:
+            self._is_moving = True
+            try:
+                # Set focuser to 0 position
+                self._send_command('M0')
```
JAAlvarado-Montes, line 170:
```suggestion
                self._send_command('M0#')
```

JAAlvarado-Montes (changes requested):
LGTM.

src/panoptes/pocs/focuser/astromechanics.py
wtgee: I'm going to do a separate small PR.

wtgee: https://github.com/panoptes/POCS/pull/1126

PR #1130, opened 2021-06-24
Allow arbitrary fits headers (again x 3)

wtgee:
Add HIERARCH card without astropy warning for long metadata keys

## Related Issue
Replaces #1016, which was based of `origin` instead of `upstream`.


wtgee:
Closed by #1150. Re-open if issue persists.

Issue #1123, opened 2021-06-22
Hanging tests


wtgee:
We are occasionally getting hanging tests, although it is not always reproducible. ~~It seems to happen either in the initial camera cooling tests or in `test_observatory` after the `test_status`. ~~

This is happening primarily in two places:

* The `test_observatory.py::test_default_config` test, which seems to have problems with reusing the `observatory` fixture.
* The `test_camera.py::test_set_target_temperature[dslr]`, which occasionally hangs before any output is given.

Both of these happen on github actions but do not seem to ever occur locally.

PR #1132, opened 2021-06-27
DSLR camera simulator

### LINKED ISSUE #1022: Simulated DSLR camera clears exposure event before FITS file is written (bug)
This bug causes a minor delay between ending the exposure and writing the FITS file, which leads to errors.

Code:
https://github.com/panoptes/POCS/blob/804e48d3e66fb6d413aef6ac2f675349bcd0f819/src/panoptes/pocs/camera/simulator/dslr.py#L58

wtgee:
The DSLR simulator was clearing the exposure event even though that should be handled by the base class.

## Description
* Don't clear exposing event in simulator as it's handled in the base class.

## Related Issue
Fixes #1022 
Hopefully helps with #1123


wtgee:
Replaced by #1149

PR #1136, opened 2021-07-12
Fix Birger Focuser initialize

danjampro:
Silly mistake left over from my last PR that causes `camera.take_exposure` to fail.

### DIFF
--- a/src/panoptes/pocs/focuser/birger.py
+++ b/src/panoptes/pocs/focuser/birger.py
@@ -153,9 +153,6 @@ class Focuser(AbstractSerialFocuser):
 
         super().connect(port=self.port, baudrate=baudrate, **kwargs)
 
-        # Set the serial number
-        self._serial_number = self._get_serial_number()
-
     def move_to(self, position):
         """
         Moves focuser to a new position.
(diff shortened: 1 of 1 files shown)


src/panoptes/pocs/focuser/birger.py
```diff
@@ -284,34 +310,6 @@ def _parse_move_response(self, response):
...
-        self._zero_encoder()
-        self._min_position = 0
-
-        # Calibrate the focus with the 'Learn Absolute Focus Range' command
-        self._learn_focus_range()
-
-        # Finally move the focus to the far stop (close to where we'll want it) and record position
-        self._max_position = self._move_inf()
-
-        self.logger.info('{} initialised'.format(self))
```
wtgee, line 313: Was this intentionally removed?

wtgee (approved):
LGTM

src/panoptes/pocs/focuser/birger.py
danjampro: The logging call? Yep - there is an equivalent call in the serial base class.

PR #1112, opened 2021-06-16
POCS Service Updates

wtgee:
This PR adds a number of changes to simplify the POCS service. Other updates are included, with the overall goal of making operations smoother.

## Description
* Added pocs service.
* ~~Added basic `pocs` cli command.~~ Moved to #1121 and #1122
* ~~Config server cli.~~ Moved to #1122
* ~~Updates for the CEM40 mounts.~~ Moved to #1119
  * ~~Updated park.~~
  * ~~Added `set_home`.~~
* ~~Scripts for taking pics changed to be consistent with each other (bias and regular). This was work that came from the lunar eclipse event.~~ Moved to #1114
* ~~Power control~~ Moved to #1120 
  * ~~Add config for the Trucker Board pocs-power service.~~
  * ~~Add service description to the docker compose file.~~ Moved to #1121 
  * ~~Better abstraction.~~
  * ~~Web api service.~~ Moved to #1121 
* ~~Add generic docs that got deleted~~
* ~~Improvements to the code in terms of using enums and dataclasses where appropriate.~~ Moved to #1115 
* ~~gphoto class correctly overrides abstract camera base with all needed methods.~~ Moved to #1118 
* ~~A bunch of f-string replacements.~~
* ~~Logger set back to a default of `DEBUG` and `INFO`.~~ Moved to #1116
* ~~Adds a `scripts/gphoto2.sh` that will either run a local `gphoto2` or pass it via ssh to the server at `$REMOTE_GPHOTO2_SERVER` based on the presence of that envvar.~~ Moved to #1114

Issue #786, opened 2019-02-19
Refactor Camera and Driver classes


AnthonyHorton:
SBIG, FLI & ZWO camera classes all do very similar things but in slightly different ways. Makes testing and maintenance hard, & will just get harder if more camera types do their own thing in the future.

A few things in particular:

- [x] When taking an exposure all three do some setup, then start the exposure, then spawn a thread to poll the camera and do readout when the exposure is complete. SBIG does this all in the low level `Driver` class while FLI & ZWO do this in the `Camera` class. None of them are using their own `is_exposing` property (recently added), instead they poll the driver method directly.  This all could be made much more consistent across camera types, and much of it moved into a base class.
- [x] Likewise all of them use a `Driver` method to scan for connected cameras, search for one with a matching UID, then claim the corresponding 'handle' or 'camera_ID'.  Again SBIG does this in the driver class while FLI and ZWO do this in the Camera class.  The code is very similar in all three cases, and most of it could be made generic and moved to a base class.
- [x] More fundamentally all of them use a single instance of a low level driver class per camera type, but create that instance differently. FLI and ZWO do this simply, making the `Driver` a class variable of the `Camera` class and creating it in the `Camera.__init__()` if no other `Camera` instance has already done so. Conversely the older SBIG code does this by overriding the `Camera` class `__new__()` method. I now have no idea why I did it this way...
- [ ] More nitpicky: I want to standardise on using Python `enum` as the Python version of the enums in the SDK C headers. I didn't know about this module when I wrote the SBIG and FLI `Driver` classes they mostly use pairs of dictionaries instead. `enum.IntEnum` and `enum.Flag` are a much closer match to the functionality of C enums.
- [x] Generally speaking there is need for more consistency in the `Driver` and `Camera` class split. I want the `Driver` classes to be thin wrappers around the corresponding camera SDK API, with a more or less one-to-one mapping of class methods to SDK functions. Additional functionality shouldn't extend much beyond Python->C & C->Python conversion of inputs and outputs, validating inputs, parsing errors, and logging. The `Camera` class, conversely, should present a standardised POCS interface via its public methods and properties, and include all the logic to convert that into calls of the low level `Driver` methods.  This is basically already the case with ZWO, is mostly true of FLI, but isn't really the case with SBIG.

Trying to work out how to attack this in manageable chunks, in a way that leaves it working at each step. Would be a mammoth PR to do it all at once.

wtgee:
@AnthonyHorton can we close this?

[...]

Issue #747, opened 2018-12-07
Allow for cameras to operate independently


wtgee:
I'm going to try some exposures of the comet 46P/Wirtanen tonight which is at a bright 5.3 magnitude. 
Would be nice to support operation where one camera exposes at 10 seconds while the other takes longer exposures (it happens to be in one of the TESS fields right now so the other camera could continue "normal" observations).

AnthonyHorton:
Feels like the ideal model is like a worker pool concurrency model. I can imagine the cameras finishing an exposure and sending a message to the observatory "I finished that one, now what?", and the observatory checking its queue of exposures and responding with either "OK, now do this one" or "OK, just wait for now".  That would be an asynchronous, messaging based system, of course.

wtgee:
It would be nice to start moving to this model in some sense. It of course connects a number of ideas we have been having (links below). Part of the problem is I think we are waiting to do a big push on this when instead we should be incrementally adding. Harder to the the latter but might be the only way we can proceed. The GSoC student would help a lot in this respect but we won't see much activity on that until about June (if that project were even selected).

* https://forum.projectpanoptes.org/t/pocs-indi-support-thread/191
* https://projectpanoptes.org/gsoc-2019/
* Inter-operability with e.g. SkyNet

AnthonyHorton:
Agreed, need to make progress in this direction. Issue is identifying manageable sized steps that take us towards the goal. I could certainly do with some pointers!

I feel like a master Issue with a check list of incremental steps would be helpful.

wtgee:
> I feel like a master Issue with a check list of incremental steps would be helpful.

We have started this (somewhere! I'll find) but I'll dig back up and then kick-start the conversation again.

I think the first step would be to split the logging mechanism out, which should be easy to do. There are plugins for pyzmq to make it easy to just broadcast the messages and then we would just need to have some kind of central subscriber mechanism that is collecting them and writing to file. This would help with a lot of components but would obviously also have immediate impact even for the pyro cameras.

AnthonyHorton:
> I think the first step would be to split the logging mechanism out, which should be easy to do. There are plugins for pyzmq to make it easy to just broadcast the messages and then we would just need to have some kind of central subscriber mechanism that is collecting them and writing to file. This would help with a lot of components but would obviously also have immediate impact even for the pyro cameras.

[...]

PR #1137, opened 2021-07-23
Remote Camera Service

wtgee:
A simple camera service that wraps the `gphoto2` command. Currently just supports the simple taking of a picture via a POST.

## Description
The service is a [fastapi](https://fastapi.tiangolo.com/) service that is capable of simple picture taking.  The service accepts a POST request defined by the following pydantic class:

```py
class Exposure(BaseModel):
    exptime: float
    filename: Optional[str] = None
    base_dir: Path = '.'
    iso: int = 100
```

If the `filename` is not provided it will default to a timestamp. `exptime` is required and the given value will first be checked if it corresponds to a built-in shutter speed, in which case that shutter speed will be used directly, otherwise a manual bulb setting will be used. `exptime` is a `float` so to use, for example the `1/4000` shutter speed index the value of `0.00025` should be given.

Since fastapi is used there is built in documentation available.  The endpoints will attempt to mimic the [Alpaca API](https://ascom-standards.org/api/) although it deviates from this implementation and may be changed in the future.  Currently the only supported endpoint is `/camera/{camera_number}/startexposure` which accepts the above.

The service will currently block until exposure is complete although there are plans to change this going forward.

## Examples:
These examples assume the service is running on the localhost on port 8080.  A Dockerfile and docker-compose is provided.

From python:
```py
import requests

camera_host = 'http://localhost'
camera_port = 8080
camera_num = 0

camera_url = f'{camera_host}:{camera_port}/camera/{camera_num}/startexposure'

exptime = 4
save_dir = '/images'
response = requests.post(camera_url, json=dict(exptime=exptime, base_dir=save_dir, iso=400))
```

From command line (via `httpie`):
```sh
http :8080/camera/0/startexposure exptime=4 base_dir=/images
```

## Related Issue
#1112 
#786 
#747

/gcbrun

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -159,9 +159,7 @@ def pytest_runtest_logfinish(nodeid, location):
         location – a triple of (filename, linenum, testname)
     """
     with suppress(Exception):
-        logger.log('testing', '')
-        logger.log('testing', f'       END TEST {nodeid}')
-        logger.log('testing', '##########' * 8)
+        logger.log('testing', f'{" END TEST " + nodeid:*^80}')
 
 
 @pytest.fixture(scope='session')
--- a/conftest.py
+++ b/conftest.py
@@ -147,9 +147,7 @@ def pytest_runtest_logstart(nodeid, location):
         location – a triple of (filename, linenum, testname)
     """
     with suppress(Exception):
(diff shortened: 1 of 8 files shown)


wtgee:
Closing in favor of #1144.

PR #1140, opened 2021-08-03
Use `self._driver` in `SDKCamera` instances

danjampro:
SDK camera subclasses are currently very hard / impossible to override cleanly because of problems with the driver. This is because `Camera` (the class name) is directly referenced in instance methods, which should be using `self` instead.

## Types of changes


wtgee (approved):
LGTM!

PR #1147, opened 2021-08-10
Apply initial focus offset when initialising camera

danjampro:
Apply initial focus offset when initialising camera so that the focuser `initial_position` definition is defined for filters with `focus_offset=0`. This ensures the first coarse focus is done with the correct focus offset regardless of the current filter on camera initialisation.

## Related Issue
AstroHuntsman/huntsman-pocs#524

## Types of changes

### DIFF
--- a/src/panoptes/pocs/camera/camera.py
+++ b/src/panoptes/pocs/camera/camera.py
@@ -131,6 +131,15 @@ class AbstractCamera(PanBase, metaclass=ABCMeta):
                 # Keep a list of active subcomponents
                 self.subcomponents[attr_name] = subcomponent
 
+        # Apply the initial focus offset
+        # This is required so that the focuser initial position corresponds to focus_offset=0
+        if self.has_focuser and self.has_filterwheel:
+            current_filter = self.filterwheel.current_filter
+            focus_offset = self.filterwheel.focus_offsets.get(current_filter, 0)
+            self.logger.debug(f"Initial focus offset for {current_filter} filter: {focus_offset}")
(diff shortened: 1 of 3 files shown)


src/panoptes/pocs/camera/camera.py
```diff
@@ -131,6 +131,15 @@ def __init__(self,
                 # Keep a list of active subcomponents
                 self.subcomponents[attr_name] = subcomponent
 
+        # Apply the initial focus offset
+        # This is required so that the focuser initial position corresponds to focus_offset=0
+        if self.has_focuser and self.has_filterwheel:
+            current_filter = self.filterwheel.current_filter
+            focus_offset = self.filterwheel.focus_offsets.get(current_filter, 0)
+            self.logger.debug(f"Initial focus offset for {current_filter} filter: {focus_offset}")
+            if focus_offset:
```
wtgee, line 140: Won't this always have a value now and so the if check isn't necessary?

wtgee: Or do you want to skip the call if the value is zero?

wtgee (approved):
LGTM

src/panoptes/pocs/camera/camera.py
danjampro: @wtgee Yeah, that was just to skip the call if the value was 0, to maximise efficiency. Can change it if you want?

wtgee: Ok with me. I'll merge.

PR #1158, opened 2021-12-29
Dockerfile updates

wtgee:
Cleanup for namespaces and updates to Dockerfile.  Some of this is taken from #1150.

* Remove unused and/or incorrect `__init__.py` files.
* Remove `mocket` and corresponding TheSkyX tests until `mocket` can be replaced.
* Dockerfile is based of `micromamba` image.
* Default `CMD` runs startup script in ipython.
* Added missing dependencies.

/gcbrun


wtgee:
Closing. We've decided (after all these years) to strip Docker from the main repo so that it builds and is testable on it's own. Other (docker) services may then include this module.

PR #1167, opened 2022-03-16
Task Manager

wtgee:
Adding a celery task manager that will become the basis for much of the future POCS operation. Maybe. ;)

For now the simple command line utilities will start and stop the messaging (rabbitmq) and results backend (redis) using the python docker sdk.

```sh
Usage: pocs tasks [OPTIONS] COMMAND [ARGS]...

  Interact with the TaskManager.

Options:
  --help  Show this message and exit.

Commands:
  start-backends  Start the celery backends specified by the config.
  stop-backends   Stop the running celery backends specified by the config.
```

## Description
The [`celery`](https://docs.celeryproject.org/) distributed task manager will allow for a number of tasks to be run in a distributed manner. For the POCS core this can include things that need knowledge of the time (e.g. startup at sunset rather than issue a sleep timer at sundown), but can also start to serve as the basis for distributed hardware control (e.g. the forth-coming task-based gphto2 camera class).


wtgee:
This is never going to get merged as is and should probably be done differently these days anyway. Closing.

PR #1249, opened 2024-04-23
Firestore database usage

wtgee:
This changes the behavior so that if the config item `panoptes_network.use_firestore` is set, then firestore will be used as the main database for long term storage and a local json file will *not* be written. The "current" document (e.g. `~/json_store/panoptes/current_weather.json`) will still be written.

This also changes the metadata firestore key to be `/metadata/{collection}/records/`, with an auto generated id.  Each record has a `unit` reference, a `received_time`, and `collection: {collection`}`.

### DIFF
--- a/.github/workflows/pythontest.yaml
+++ b/.github/workflows/pythontest.yaml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


wtgee:
This is going to be pulled from the normal POCS operations and instead act on the files itself in a separate process.

PR #1286, opened 2025-02-18
Fix simulator config when using auto run.

### LINKED ISSUE #1285: Simulators specifed on the command line persist in the config file.
When running a simulator via the command line, e.g. `pocs run -s weather auto`, the simulators used are written to the config file but are not cleared upon shutdown so they are incorrectly used the next time pocs is run.  

`pocs run auto` should be clear about what simulators are being used and perhaps given the option to enable/disable.

wtgee:
The `simulator` config item should be set even when it is empty, which will clear any previous runs. Also be more explicit in the logging and printing.

Fixes #1285

### DIFF
--- a/.github/workflows/pythontest.yaml
+++ b/.github/workflows/pythontest.yaml
@@ -43,7 +43,7 @@ jobs:
       - name: Install
         run: pip install ".[google,focuser,sensors,testing]"
       - name: Test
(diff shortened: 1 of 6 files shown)


wtgee:
Some of the tests rely on this saving behavior and will need to be updated.

PR #1297, opened 2025-03-31
Take pics futures

wtgee:
This was code being used on `007` for use with the `pocs camera take-pics`, which cleans up read timeouts.


wtgee:
Tested on `026` and it didn't seem to make a difference. The real culprit seemed to be the `cameras.default.readout_time` setting was too low. Closing this ticket for now.

PR #1307, opened 2025-04-08
Refined alignment procedure

wtgee:
Adds a new alignment procedure that simply takes three different images (home, 5 seconds to east, 5 seconds to west), identified the same star in each of the three images, and computes the circle of rotation for the mount and gives offsets.

image


wtgee:
Messed up history.

PR #1326, opened 2025-06-04
Reset the current observation when getting a new observation.

wtgee:
There is a small bug that occurs sometimes when getting a new observation that causes the old obseravtion information to be used. Resetting the `current_observation` at the start of `get_observation` should clean things up.


wtgee:
This isn't the right approach. Closing.

